PR #474, opened 2014-06-16
Endf in C++

crbates:
from #454


scopatz:
Thoughts on this @crbates?

crbates:
I'm torn between elegant c++ objects and something that can have a sane cython/xdress friendly interface and where to compromise between the two. I have to think more about how I want to design the interface. I'm sorta leaning towards the masochistic version: an elegant c++ parser with a seperate c++ wrapper that is cython friendly.

scopatz:
On that note, it is probably possible to have a hand-written cython wrapper which is much nicer than anything you'd get from a tool.  I don't think that this would be too ugly.

src/endf_mt.cpp
```diff
@@ -0,0 +1,600 @@
...
+  mf3_ob.energy = tab.x;
+  mf3_ob.sigma = tab.y;
+
+  std::string line;
+  getline(infile, line);
+  return mf3_ob;
+}
+
+
+pyne::endf::mf2 pyne::endf::read_mf2(std::ifstream &infile, int lrp) {
```
crbates: @scopatz let me know what you think of this data structure (it's not done yet but this is the general idea).

scopatz: The mf2, I assume?

src/endf_mt.h
```diff
@@ -0,0 +1,222 @@
...
+    
+    typedef struct mf2 : mt_base {
+      int nis;
+
+      // mega-maps map(pair(quantity, unique_index), data)
+      std::map<std::pair<std::string, std::string>, int> data_i;
+      std::map<std::pair<std::string, std::string>, double> data_d;
+      std::map<std::pair<std::string, std::string>, std::vector<int> > data_vi;
+      std::map<std::pair<std::string, std::string>, std::vector<double> > data_vd;
+    } mf2;
```
scopatz: Yes, this seems reasonable. It might be nicer to do this with tuples in C++11. But 2-tuples are essentially pairs and so I don't think that it is a big deal.

crbates: tuples would be nice because then the unique index could be a tuple with unused indices defaulting to -1 instead of a "_" concatenated string which was my current plan . It would make ordering operations much simpler.

scopatz: I like that a lot more too.  I think that this is very compelling reason for moving to C++11.  @makeclean wants to too, I think.

makeclean:
@crbates could you rebase against develop and push again?

gonuke:
@crbates - has this been superseded by the emergence of njoy/ENDFtk ?

If so - perhaps we can close this PR?  Or you could withdraw it?

makeclean:
I think we can close those?

PR #640, opened 2014-12-23
forward decl for decayer in case of amalgamation

scopatz:


scopatz:
How were you compiling the test? Sorry to get back to this so late.

crbates:
I made a basic amalgamation test shell script + makefile + helloworld.cpp. I will put it up on github tonight if I can find it.

scopatz:
Thanks!

crbates:
@scopatz  here's the test program I used https://github.com/crbates/pyne-cpp

scopatz:
Hi @crbates! Thanks this clears things up. The reason this fails is because it is not downloading and compiling the decay functionality. That brings up an important question about what to do... In cyclus we have dummy files:

* https://github.com/cyclus/cyclus/blob/develop/src/_pyne_decay.h
* https://github.com/cyclus/cyclus/blob/develop/src/_pyne_decay.cc

Perhaps we should have more fully fledged dummies that throw an error if they are called.  I think a major problem is that we don't want to amalgamate decay, since it needs to be compiled with `-O0`. Ideas?

crbates:
can't we just use the existing dummy decay files in PyNE during the amalgamation step?

crbates:
Also I agree there should be some error or warning if the `nuc_id` isn't in the switch case list.

scopatz:
> can't we just use the existing dummy decay files in PyNE during the amalgamation step?

And this is why you are the smartest developer alive, @crbates :).  I must have put these dummies in there, but I totally forgot that they existed. Why I remember the cyclus ones is beyond me. I must be going mad.

> Also I agree there should be some error or warning if the nuc_id isn't in the switch case list.

That should probably be in the decaygen. Mind if that is a separate PR?

scopatz:
Ok, so say that we use the dummies. Then someone does try to download and (statically) link in the real files, won't the symbols collide?

scopatz:
Maybe we could have pound-defined variable encapsulating the dummies.  It doesn't really change the header at all.

crbates:
I think this is the right idea. I mean the use case is needing to amalgamate instead of linking to `libpyne.*` and using the headers. 

>> Also I agree there should be some error or warning if the nuc_id isn't in the switch case list.

> That should probably be in the decaygen. Mind if that is a separate PR?

I think that makes sense.

scopatz:
Ok! I think that this is ready to go. I have put in a PR to https://github.com/crbates/pyne-cpp/pull/1 which now works for me.  Also do you think that this test repo should live in the pyne org. Seems super useful.

scopatz:
Also, super huge thanks.

crbates:
we seem to be in the queue nether.

crbates:
I'm going to PR the python 3 one separately. thanks @scopatz!

PR #646, opened 2015-01-15
Adding partisn module with write_partisn_input

kkiesling:
This is a new module for partisn tools. Included in the module is a tool (write_partisn) to create a partisn input file. Also added in this PR is a function to the materials module to return atom density (added userguide entry) and a function in the dagmc module to return a dictionary of material and cell assignments. Note: I played around a bit with the location of the "get_cell_material_assignments" so let me know if you think it would work better in a different location within the file. All tests are included.


pyne/partisn.py
```diff
@@ -0,0 +1,816 @@
...
+    for z in zones_mats.keys():
+        zones_novoid[z] = {'mat':[], 'vol_frac':[]}
+        for i, mat in enumerate(zones_mats[z]['mat']):
+            if mat not in skip_list:
+                name = strip_mat_name(mat)
+                zones_novoid[z]['mat'].append(name)
+                zones_novoid[z]['vol_frac'].append(zones_mats[z]['vol_frac'][i])
+    
+    # Put zones into format for PARTISN input
+    if 'x' in bounds.keys():
```
scopatz: Better to do `if 'x' in bounds`, etc.

kkiesling: I guess I never realized that worked. I will have to go through and change that in several places.

pyne/dagmc.pyx
```diff
@@ -591,6 +599,80 @@ def get_material_set(**kw):
...
 
+
+def get_cell_material_assignments(hdf5):
+    """Get dictionary of cell to material assignments
+    
+    Parameters:
+        hdf5 : str, path to hdf5 material-laden geometry
+    
+    Returns:
+        mat_assigns : dict, keys are cell numbers and values are material names
```
scopatz: This docstring is not in numpydoc format.

pyne/dagmc.pyx
```diff
@@ -591,6 +599,80 @@ def get_material_set(**kw):
             mat_ids.add(d['material'])
     return mat_ids
 
+
+def get_cell_material_assignments(hdf5):
```
scopatz: I think that this function could just be called `cell_material_assignments()`

pyne/partisn.py
```diff
@@ -0,0 +1,816 @@
...
+    must be set. 
+    
+    Notes:
+        This does not write out all necessary inputs for the solver and cross
+        section library (block 3 and 5). There is no assumed cross section 
+        library type.
+    
+    Parameters:
+    -----------
+        mesh : PyNE mesh object, a premade mesh object that conforms to the
```
scopatz: This should not be indented to be numpydoc compliant.

pyne/partisn.py
```diff
@@ -0,0 +1,816 @@
...
+        mesh : PyNE mesh object, a premade mesh object that conforms to the 
+            geometry. Bounds of the mesh must correspond to the desired PartiSn
+            fine mesh intervals. One fine mesh per coarse mesh will be created. 
+            Can be 1-D, 2-D, or 3-D.
+            Note: Only Cartesian meshes are currently supported.
+        hdf5 : file path, a material-laden dagmc geometry file.
+        ngroup : int, the number of energy groups in the cross section library
+        nmq : int, the number of moments in a P_n expansion of the source
+        
+    Optional Parameters:
```
scopatz: No need for an optional paremeters section, should be under parameters

pyne/partisn.py
```diff
@@ -0,0 +1,816 @@
...
+    for i in range(im):
+        for jk in range(jm*km):
+            zones_formatted[i,jk] = voxel_zone[n]
+            n += 1
+    
+    return zones_formatted, zones_novoid
+    
+
+def _write_input(title, block01, block02, block03, block04, block05, **kwargs):
+    """Write all variables and comments to a file.
```
scopatz: put `name=None` here in the signature.

pyne/partisn.py
```diff
@@ -0,0 +1,816 @@
...
+    """
+    
+    # Create file to write to
+    if 'name' in kwargs:
+        f = open(kwargs['name'], 'w')
+    else:
+        file_name = str(title) + '_partisn.inp'
+        f = open(file_name, 'w')
```
scopatz: Rearrange this to be

```python
file_name = str(title) + '_partisn.inp' if name is None else name
f = open(file_name, 'w')
```

pyne/partisn.py
```diff
@@ -0,0 +1,816 @@
...
+    if "/" in hdf5:
+        title = hdf5.split("/")[len(hdf5.split("/"))-1].split(".")[0]
+    else:
+        title = hdf5.split(".")[0]
+    
+    # call function to write to file
+    if input_file_tf:
+        _write_input(title, block01, block02, block03, block04, block05, name=input_file)
+    else:
+        _write_input(title, block01, block02, block03, block04, block05)
```
scopatz: These four lines could be replaced with the following:

```python
input_file = input_file if input_file_tf else None
_write_input(title, block01, block02, block03, block04, block05, name=input_file)
```

pyne/partisn.py
```diff
@@ -0,0 +1,816 @@
...
+    mats_collapsed = {}
+    for mat_name in mats.keys():
+        if collapse:
+            mats_collapsed[mat_name] = mats[mat_name].collapse_elements(mat_except)
+        else:
+            mats_collapsed[mat_name] = mats[mat_name]
+
+    # convert mass fraction to atom density in units [at/b-cm]
+    mat_lib = {}
+    for mat_name, comp in mats_collapsed.iteritems():
```
scopatz: probably better to use `mats_collapsed.items()` for Python 3 compatibility.

pyne/partisn.py
```diff
@@ -0,0 +1,816 @@
...
+    section library from the list of nuc_names.
+    """
+    
+    xs_names = []
+    for mat, nuc_set in mat_xs_names.iteritems():
+        for name in nuc_set.keys():
+            if name not in xs_names:
+                xs_names.append(name)
+    
+    return xs_names
```
scopatz: This is basically a set operation so I would write the body of this function as:

```python
xs_names = set()
list(map(xs_names.update, mat_xs_name.values()))
return list(xs_names)
```

pyne/partisn.py
```diff
@@ -0,0 +1,816 @@
...
+            count += 1
+            if count == 4:
+                if tot != len(block05['sourcz']):
+                    f.write(";\n        ")
+                else:
+                    f.write(";")
+                count = 0
+            else:
+                f.write("; ")
+        f.write("\n")
```
scopatz: So this is a somewhat unsafe way to create files.  It is much better to create the full string representation in memory ahead of time and then call `write()` once and only once at the end of the function.  In this way you don't risk accidentally writing part of file. Also, it is faster because you are not constantly hitting the hard disk.

scopatz:
Hi @kkiesling - Overall, the content here seems very reasonable and worthwhile.  However, there are a number of stylistic improvements that affect safety and performance that should probably be addressed. I'll let @makeclean review more thoroughly

pyne/dagmc.pyx
kkiesling: That is much better. Is where I placed the function in the file a reasonable location? I wasn't sure because there seemed to be a few "groups" of functions in the file.

scopatz: Where it is seemed fine to me.

kkiesling:
I believe I have addressed all of @scopatz 's comments. @makeclean you should check the module and tests out to make sure they meet your needs. I have not addressed the dagmc.load issue with multiple loads so I know the the build and tests do not pass yet on batlab.

makeclean:
@kkiesling following our chat on Friday, can you update this PR, the produced PartiSn file had a few issues, but apart from that it was pretty good. Once those changes are addressed I will comment again. Thanks!

kkiesling:
@makeclean The changes we talked about have been updated.

kkiesling:
So this passes all the build and tests now except for on Ubuntu10 where its seems that the modules sets and izip do not exist. Does anyone know some good work arounds for this?

Other than that, @scopatz my changes to the dagmc tests are here and should be good to go if you wanted to check them. All I did was remove the class and added a function to call the individual tests in new processes. And @elliottbiondo maybe just double check the changes I made to your r2s module/test.

scopatz:
Hi @kkiesling, just remove the "from sets import Set" line and use the built-in `set` instead of `Set`. see https://docs.python.org/2/library/sets.html#comparison-to-the-built-in-set-types

Also, Ubuntu 10 is where we test python 3.  In Python 3, izip was renamed to zip and izip was removed.  You can probably add a line or two like the following at the top of the file:

```py
import sys
if sys.version_info[0] > 2:
    izip = zip
```

scopatz:
Tests seem good too.

scopatz:
Thanks @kkiesling

kkiesling:
Ping! It looks like it passes on CI!

scopatz:
Sweet @kkiesling! It looks like there might be a merge conflict. Would you mind merge from develop into this branch, resolving any conflicts and then pushing back up? Thanks!

kkiesling:
Whoops! Critical mistake on my part.. fixed now!

scopatz:
Thanks @kkiesling! This is great. From now on though, try to put in smaller PRs.  I am glad we worked through the test issues.  Proves that the dev calls are actually useful :)

PR #648, opened 2015-01-16
Updating write_tape5_irradiation() docs

jdangerx:
While running ORIGEN today I found this interesting behavior: if you write out a TAPE5.INP with `nan` or `inf` flux values, ORIGEN will run but not produce any output in the TAPE6. Here is a warning in the docs so we never have to re-find this bug again.


scopatz:
Hey John,  can you also make this throw a ValueError too please?

jdangerx:
Yeah. I'll ping when I do.

jdangerx:
Ping! I also took the liberty to lint the tests file. Turns out `test_parse_tape6` was defined twice, shadowing the first declaration, so I fixed that too.

scopatz:
Thanks John!

scopatz:
This is a big help.

PR #649, opened 2015-01-21
Update Material docs from attrs to metadata

gonuke:
Discovered that the users guide section on materials still referred to 'attrs'.


crbates:
looks good. thanks @gonuke

PR #652, opened 2015-01-28
added optional setuptools

scopatz:
Should have been done a while ago, given distutils lack of future support.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -300,6 +300,12 @@ def parse_args():
 
 
 def setup():
+    try:
+        from setuptools import setup as _setup
+        have_setuptools = True
+    except ImportError:
+        from distutils.core import setup as _setup
+        have_setuptools = False
     scripts = [os.path.join('scripts', f) for f in os.listdir('scripts')]
     scripts = [s for s in scripts if (os.name == 'nt' and s.endswith('.bat'))
                                      or (os.name != 'nt' and
@@ -346,7 +352,7 @@ def setup():
         "data_files": data_files,
         "scripts": scripts,
         }
-    rtn = core.setup(**setup_kwargs)
+    rtn = _setup(**setup_kwargs)
 
 
 def cmake_cli(cmake_args):
--- a/setup.py
+++ b/setup.py
@@ -351,6 +351,7 @@ def setup():
         "package_data": pack_data,
         "data_files": data_files,
         "scripts": scripts,
+        "zip_safe": False,
         }
     rtn = _setup(**setup_kwargs)
 


jdangerx:
Looks like the CI isn't making it to the test phase. Not entirely sure why.

scopatz:
Relaunching

scopatz:
Actually what is the website?

jdangerx:
Looks like a legit install error. The logs say:
`ImportError: libpyne.so: cannot open shared object file: No such file or directory`
and
`/slots/02/dir_5046/userdir/anaconda/lib/python3.4/site-packages/pyne-0.5.dev0-py3.4.egg/pyne/__init__.py:15: Warning: Error importing PyNE: you should not try to import PyNE from its source directory; please exit the PyNE source tree, and relaunch your python interpreter from there.`
Incidentally, this message is kind of misleading - though importing PyNE from /pyne will cause an import error, other things can do it too. I've been bit by this before. Maybe we should change the wording a bit?

scopatz:
The dashboard is at http://gorgus.pyne.io/dashboard doh

scopatz:
Uggh I think that this is a CI & setuptools related issue. The PR works just fine on my machine. Can you confirm?

crbates:
I have the same problem as CI (and using macports Python not conda). I think it's because the relative path links get messed up since it's now in the egg folder but thats just a guess.

scopatz:
Ohhh I have a fix for this

scopatz:
Ok, the fix is in.  Hopefully it passes CI.

crbates:
`ImportError: libpyne.so: cannot open shared object file: No such file or directory ` :(

scopatz:
Hmm this works just fine on my computer.  Does it work locally for you @jdangerx or @crbates?

crbates:
This doesn't work locally for me. I think it tries to outsmart cmake when re-linking libpyne (which ends up in `egg/lib/libpyne.*`) but it's not 100% clear

scopatz:
Python packaging is so gross. I hate eggs, and not just because I am vegan.

katyhuff:
:+1:

scopatz:
Found this and their arrogance: https://pythonhosted.org/setuptools/setuptools.html#automatic-resource-extraction

After looking at this some more, I don't think that our cmake, make, setup strategy is going to work with modern setuptools.  We could add some sort of ridiculous awful hack that attempts to figure out where setup tools installed things and then move the lib and include dirs out of that directory into the prefix directory after install.  But frankly, I don't trust setuptools to not break things again.

A better strategy might be to move to a cmake, make, setup, make install strategy, where everything in the src dir gets installed with cmake and everything in the pyne dir is installed with setup (even if it was built with cmake).  

Also, do we need to install the fortran *.mod files? I don't think we do right now. Do this go into include or lib? Forgive my newbishness, my fortran knowledge starts at subroutines and ends at `!`

makeclean:
For fortran I would suggest a route like we have for c++. We could chuck all the .o files into an archive .a file shove that into the lib.

As far as I remember the mod files are produced at compile time as a first step, used to make .o files and then are forgotten about.


Sent on a Sprint Samsung Galaxy Note® 3

<div>-------- Original message --------</div><div>From: Anthony Scopatz <notifications@github.com> </div><div>Date:02/02/2015  7:23 PM  (GMT-06:00) </div><div>To: pyne/pyne <pyne@noreply.github.com> </div><div>Cc:  </div><div>Subject: Re: [pyne] added optional setuptools (#652) </div><div>
</div>Found this and their arrogance: https://pythonhosted.org/setuptools/setuptools.html#automatic-resource-extraction

After looking at this some more, I don't think that our cmake, make, setup strategy is going to work with modern setuptools. We could add some sort of ridiculous awful hack that attempts to figure out where setup tools installed things and then move the lib and include dirs out of that directory into the prefix directory after install. But frankly, I don't trust setuptools to not break things again.

A better strategy might be to move to a cmake, make, setup, make install strategy, where everything in the src dir gets installed with cmake and everything in the pyne dir is installed with setup (even if it was built with cmake).

Also, do we need to install the fortran *.mod files? I don't think we do right now. Do this go into include or lib? Forgive my newbishness, my fortran knowledge starts at subroutines and ends at !

scopatz:
> As far as I remember the mod files are produced at compile time as a first step, used to make .o files and then are forgotten about.

good to know.

scopatz:
Moving to v0.6

PR #653, opened 2015-01-29
ensdf fixes

crbates:
This improves ensdf parsing for levels. It also switches to using ensdf decay dataset data instead of ensdf level dataset data for children and branching ratios in decaygen.py eliminating the need for special cases. Also adds a data checksum test.


crbates:
lots of things are broken. I'll get to tests on this later this week.

scopatz:
Nice! Thanks for fixing this up.  Also would you mind adding a couple of tests for `gamma_child_byen()` and `gamma_child_byparent()`?

crbates:
sure. I'm also going to add a test for the special cases in `decaygen.py` I'm claiming I fixed.

scopatz:
molto bene

crbates:
ping @scopatz . no rush

scopatz:
This requires a `nuc_data_make --clean 1`, right?

crbates:
yes. Several of the tables have pretty significant changes.

crbates:
it's probably a good idea to re-run the ORIGEN benchmarks too if that is not too involved.

scopatz:
I want to bring the orgien benchmarks in as pyne tests, relatively soon. I don't really want to block this on that though. Do you? It is a little hard to tell what is dependent on what here though...

scopatz:
I think this looks good and I am ready to merge pending your approval, @crbates .

crbates:
I don't think it's necessary to block on the benchmark. We aren't guaranteeing it works yet so it's fine. It now uses information from the decay dataset data instead of the level dataset data which is more accurate based on my anecdotal survey. At some point someone should look into that (maybe it could be a GSOC type project as it would be pretty straightforward and at least conference worthy).

scopatz:
Thanks @crbates!

scopatz:
I do look forward to updating the benchmark :).

PR #655, opened 2015-02-05
adds docs only for additional helper functions

gidden:
this is a by-hand update and will require additional by-hand updates


gidden:
either this or #656 should be merged but not both

scopatz:
Closing if favor of #656. Thanks @gidden

PR #656, opened 2015-02-05
automodules all of enr

gidden:
this PR will lose the (nice) headings in favor of automatic updates


gidden:
either this or #655 should be merged but not both

gidden:
also, this should likely be for @scopatz to decide since the original was his work

scopatz:
This is less work to maintain so I'll merge this.

scopatz:
Thanks!

PR #657, opened 2015-02-09
make f2py search robust

crbates:


scopatz:
Looks great @crbates! Just waiting on CI.

PR #658, opened 2015-02-09
fix ahot tests on Python 3

crbates:


crbates:
I should have caught this in the f2py PR. oops

scopatz:
Thanks!

PR #664, opened 2015-02-13
convert to state_id's in decay

crbates:
This should fail until a new decay.tar.gz is uploaded.


crbates:
The failing test includes isotopes in the decay result that are known to **not** be U235 daughters (this change removes them).

PR #666, opened 2015-02-14
Move gallery to examples and add links

crbates:
also fix ipython notebook formatting for firefox and chrome. #663 reopened


crbates:
screenshot of updated homepage

screen shot 2015-02-13 at 21 25 44

scopatz:
I can't say that I am surprised that #666 has fixes related to the browser rendering wars ;)

scopatz:
I am having trouble building on Python 3.4 because of numpydoc.  Could we make this an explicit dependency and remove the local copy? I'd be happy to do this in a separate PR.

crbates:
> I am having trouble building on Python 3.4 because of numpydoc. Could we make this an explicit 
> dependency and remove the local copy? I'd be happy to do this in a separate PR.

Sounds good to me. I think it's justifiable in the name of Python 3 compatibility

scopatz:
I approve of this PR, but #670 should probably be merged instead.  @crbates, will you do the honors?

PR #662, opened 2015-02-13
Update tutorial to by yt 3.x compliant

crbates:

PR #667, opened 2015-02-14
Update tutorial to be yt 3.x compliant

crbates:
and other fixes. #662 re-opened.


crbates:
the diffusion notebook now runs without any yt errors but JSanimation no longer seems to work properly with recent ipython versions.

scopatz:
I think that is OK.  Maybe we should drop JSanimation. I was a neat trick, but non-essential.

scopatz:
Also pymovie seems much better

crbates:
@scopatz I propose we merge this after adding a warning at the start of the 8th tutorial that links to the issue because I don't think obscure yt/mesh problems should block 0.5.

scopatz:
I completely agree.

scopatz:
We are going to need to sit down with some of the yt folks to really hammer this out.

crbates:
ping @scopatz

scopatz:
This looks about as good as it is going to get.  Thanks for all the hard work @crbates!

PR #668, opened 2015-02-15
fix some weird errors on mac

crbates:
clang + fortran do not play well together. This switches to the c++ endftod if the fortran one doesn't work in tests.


crbates:
closing in favor of #669

Issue #558, opened 2014-11-03
Tests fails on mac when libgcc_s is linked when compiled with gcc


crbates:
`test_material.test_getitem_str` calls SIGABRT when raising the exception on recent versions of gcc.

crbates:
So this isn't a problem on arch linux with gcc 4.9.2. perhaps it's a macports thing :(

scopatz:
This is also not a problem on Ubuntu 14.10.

crbates:
After much forum searching about this I found a comment on an old macports ticket https://trac.macports.org/ticket/25042 and did some regression testing to determine that this issues is caused by `libgcc_s` causing exceptions to not being caught/passed to linked libraries properly.

scopatz:
Sadly, this is from 5 years ago with no activity in the past 4. Is the right answer to write an LLVM-based Fortran compiler?!

crbates:
I think the right answer is not to use `gcc_s` on mac. I don't know that it's needed for anything. The `ahot` stuff seems to build fine without it and linking it doesn't fix any of the `ahot` runtime problems as far as I can tell.

scopatz:
Closed by #669

PR #669, opened 2015-02-17
fix pyne build with gcc on macports

crbates:
This closes issue #558 by not linking against `libgcc_s` on mac. It also fixes issues with gfortran/c compatibility with the default compiler options on mac (pre-empting #668)


scopatz:
Awesome! Thanks @crbates!

PR #677, opened 2015-02-25
Hatedf5

kkiesling:
Attempt at fixing the hdf5 version issue for 1.8.11. The 1.8.14 file version works for an installed version of 1.8.11.


scopatz:
hmmm It looks like you added a lot of extra very large files here, that probably should be ignored.  Could you please branch from develop again, redo the change, and commit only the change to test_decay.py?  Thanks!

scopatz:
Also close this issue too :)

kkiesling:
You're right. I didn't see those files. They don't appear as tracked files at all on my branch. I'll fix it.

Issue #682, opened 2015-02-26
isomer treatment in pyne.endf


ohnemax:
I ran into trouble reading an ENDF formatted file containing data for an Isomer (e.g. Pm-148M, https://www-nds.iaea.org/public/download-endf/ENDF-B-VII.1/n/n_6153_61-Pm-148M.zip) using the class provided by pyne.endf. The data of the endf file is stored using index 611480000, however it should be 611480001. I tried to write a fix, but that broke other tests. So here is some explanation:

I think the excited state should be triggered by the value of the LISO entry in the first part of the endf file. The manual says "LISO: Isomeric state number. The ground state is indicated by LISO=0.  [...]" 

The data structure of the beginning of 451 from ENDF Manual looks like this:
```
[MAT, 1,451/ ZA, AWR, LRP, LFI, NLIB, NMOD]HEAD
[MAT, 1,451/ ELIS, STA, LIS, LISO, 0, NFOR]CONT
[MAT, 1,451/ AWI, EMAX, LREL, 0, NSUB, NVER]CONT
[MAT, 1,451/ TEMP, 0.0, LDRV, 0, NWD, NXC]CONT
```

The file I used is an ENDF File downloaded from IAEA database (). It has LISO=1, but the current code fails to use the correct nuclide id. I add a PR with some example ENDF snippet + test in a few minutes. Tracing back the issue in https://github.com/pyne/pyne/blob/develop/pyne/endf.pyx it seems to be a problem in the following code:

```
line = fh.readline()
mat_id = int(line[66:70].strip() or -1)
# originally in a float version of ZZAAA.M, ie 94242.1
nuc = cpp_nucname.id(<int> (endftod(line[:11])*10))
# Make a new dict in self.structure to contain the material data.
```

Apparently the code assumes that the isomeric state is store as a decimal in the ZA entry, which is not the case for all the files at ENDF-B-VII.1 I looked at. So I made a fix suitable for this case (second PR).

However, there is a test snippet currently in the test file that has a line, that actually uses the implemented way to distinguish isomers.
```1.003100+3 2.098312+0 1 0 0 1 131 1451 1```

With my fix, this fails, because the test snippet does not have a LISO=1 entry. I'm not sure if this is an inconsistency in the ENDF definition, or just a wrong posed test. I would happy to fix it, either
a) add a check to pyne.endf to make both ways work
b) correct snippet in test_endf
but I am not sure what would be the correct solution.

crbates:
[...]

PR #683, opened 2015-02-26
Isomer test

ohnemax:
This PR only adds a test and is related to issue #682 .


ohnemax:
Added test to fix #684, hence closed

PR #685, opened 2015-02-27
squash bugs that lead to test_decay issues

crbates:
includes:
 * better ensdf parsing of parents
 * unassigned levels default to 9999
 * decays can’t add up to more than 100%


crbates:
This merge is blocked until the default discussion on the list has been settled.

scopatz:
Should this be tagged v0.5?

crbates:
There's no inherent reason I just thought it would be nice to have the decay stuff squared away.

scopatz:
Ok!

crbates:
negative nuc_id's already return `False` for `pyne::isnuclide()` so trying to use rxname parent/child functions will raise an error. I think this is the correct behavior.

scopatz:
Last v0.5 issue!  Do we expect this to fail because it needs a new decaygen?

crbates:
yes

tests/test_decay.py
```diff
@@ -61,11 +61,7 @@ def setup():
     350830000,  # decay to metastables without reason.
     340830000,  # ...
     481170000,  # ...
-    471110001,  # missing branch in origen
-    501130001,  # ...
-    561330001,  # FIXME These have too much mass on our end
-    491180000,  # ...
-    801970001,  # ...
```
scopatz: It seems like this removal is causing the test failures.

src/decaygen.py
```diff
@@ -224,7 +224,7 @@ def b_from_a(cse, a_i):
     bkey = EXP_EXPR.format(a=a_i)
     return cse[bkey]
 
-def chainexpr(chain, cse, b, short=1e-8):
+def chainexpr(chain, cse, b, bt, short=1e-8):
```
scopatz: What is bt out of curiosity?

tests/test_decay.py
scopatz: Oh nm.

crbates: yep, after building `data.cpp` with this pr with a fresh nuc_data.h5 these tests should no need to be blacklisted.

scopatz: Ummhmm - my brain got it, eventually. I feel slow this morning, for sure.

scopatz:
Looks great to me! Do you want to do the decay upload honors, @crbates?

src/decaygen.py
crbates: this keeps track of the total end output and make sure we are not "creating" mass. This seems to be an issue when we have multiple paths to the same stable final nuclide.

crbates:
it's up!

PR #696, opened 2015-03-06
ENDF Library more ENDF-6 Format Compliant

ohnemax:
The ENDF-6 Formats Manual allows to combine data for several materials in single files (than they are often called 'tape'). An example: http://www.nndc.bnl.gov/endf/b6.8/tapes/tape.100
Trying to load the mentioned file lead to errors, commits 445dfb8bd9b244c2b04db71c1eae7eaa268b3943 and 6277679e9864cefd04fd6deb95862f50f5699513 add a test for this scenario.

The issue is caused by an improper implementation of _read_headers and the sample endf tape which has been created for the test cases. The sample endf tape makes use of more than one TEND record (`[ -1, 0, 0/ 0.0, 0.0,  0,  0,  0, 0]`), basically after each material. This seems wrong, compared to other examples. It also has a line of text after each of these records, for which I couldn't find reference in the manual, or examples. I removed both lines. The only text line left is the absolute first line of the file.

To do this, a couple of changes were necessary in the structure of the Library class. Outcome should be the same. I hope I understood the ENDF Manual correctly...

The last large commit contains only style - tried to make the endf.pyx mostly consistent with http://pyne.io/devsguide/style_guide.html#python-style-guide


scopatz:
relaunching CI

pyne/endf.pyx
```diff
@@ -91,6 +93,20 @@ class Library(rx.RxLib):
...
+            if isinstance(self.fh, basestring):
+                fh = open(self.fh, 'r')
+                opened_here = True
+            else:
+                fh = self.fh
+            line = fh.readline()
+            self.chars_til_now = len(line)
+            self.offset = 81 - len(line)
+        else:
+            print('TPID is the first line, has been read already')
```
scopatz: This is probably better as a warning.

pyne/endf.pyx
```diff
@@ -91,6 +93,20 @@ class Library(rx.RxLib):
             fh.close()
         return data
 
+    def _read_tpid(self):
+        if(self.chars_til_now == 0):
```
scopatz: PEP8: space after if

scopatz:
This looks great to me @ohnemax!  Waiting for CI.

scopatz:
It seems that there is a dtype error in python3 http://submit-1.batlab.org/nmi-runs/pyne-ci/2015/03/pyne-ci_submit-1.batlab.org_1425662985_27215/userdir/nmi:x86_64_Ubuntu14/remote_task.err

tests/test_endf.py
```diff
@@ -480,6 +78,28 @@ def test_endftod():
...
+    with open("endftape.100", "rb") as f:
+        obs_hash = md5(f.read()).hexdigest()
+    exp_hash = "b56dd0aee38bd006c58181e473232776"
+    try:
+        assert_equal(obs_hash, exp_hash)
+    except AssertionError:
+        raise AssertionError("endftape.100 hash check failed; please try redownloading the endftape.100 data file.")
+    testlib = Library("endftape.100")
+    exp_nuclides = [10010000, 30060000, 40090000, 50110000, 30070000, 60000000, 50100000]
+    obs_nuclides = testlib.structure.keys()
```
scopatz: this line should probably be 

```python
obs_nuclides = list(map(int, testlib.structure.keys()))
```

ohnemax: @scopatz let me try to setup a python3 virtualenv, but i guess your suggestion should be perfect

scopatz: Ok! You can always change the code and push it up and then rely on the CI servers to test it for you too, if a local python3 is too annoying.

ohnemax: the change seems to work in my python3 env - let's see what the servers say...

scopatz:
Looks great! Thanks @ohnemax!

PR #697, opened 2015-03-06
Tutorial updates (intro and outro)

kkiesling:
I changed the wording in the introduction and outroduction of the tutorial so that it didn't sound like you are actually in a session but can work through it on your own. 

I also tried to play around with (but couldn't figure it out) the PyNE Tree image that doesn't render in the intro and outro. The rendered html just has "PyNE Tree PyNE Tree" and the ipynb has an empty image. There are no changes to reflect that here but I was wondering if anyone else knew how to fix it?

### DIFF
--- a/tutorial/00-intro.ipynb
+++ b/tutorial/00-intro.ipynb
@@ -24,10 +24,9 @@
      "cell_type": "markdown",
      "metadata": {},
      "source": [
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks @kkiesling!

PR #698, opened 2015-03-06
fixed LaTeX equation error

kkiesling:
There was an error in the equation in bins.rst. I simply fixed it so that it would actually render. I am not sure if it is rendering the equation 100% correctly since I do not know what it should be originally. @scopatz it looks like you were the original author so you might want to double check.

### DIFF
--- a/docs/theorymanual/bins.rst
+++ b/docs/theorymanual/bins.rst
@@ -25,7 +25,7 @@ collapsed value :math:`y_g` is:
 
 .. math::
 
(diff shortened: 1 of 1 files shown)


scopatz:
Looks correct to me. Thanks a ton @kkiesling

kkiesling:
I just pulled in the latest changes so when CI is done, go ahead and merge.

scopatz:
In the name of all that is good and right, w00t.

PR #699, opened 2015-03-17
Dagmc tests moved assert statements

kkiesling:
I realized as I was making a presentation about nosetests with multiprocessing that if you have assert statements in the function you are calling (not the function test_X), that you can receive false positives, meaning the test will return a pass even if it shouldn't. I changed all the dagmc tests that were set up like that originally so now all assert statements are made in the outermost function. Luckily, no false positives were occurring before.

I also added one file to the .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ rs.cred
 src/decay.h
 src/decay.cpp
 tests/fort.8
+tests/endftape.100
--- a/pyne/dagmc.pyx
+++ b/pyne/dagmc.pyx
@@ -322,11 +322,11 @@ def load(filename):
 
 def get_surface_list():
     """return a list of valid surface IDs"""
-    return surf_id_to_handle.keys()
+    return list(surf_id_to_handle.keys())
 
 def get_volume_list():
     """return a list of valid volume IDs"""
-    return vol_id_to_handle.keys()
+    return list(vol_id_to_handle.keys())
 
 
 def volume_is_graveyard(vol_id):
--- a/tests/test_dagmc.py
+++ b/tests/test_dagmc.py
@@ -45,16 +45,11 @@ def metadata():
     from pyne import dagmc
     dagmc.load(path)
     
-    rets = [dagmc.volume_is_graveyard(x) for x in range(1, 5)]
-    assert_equal(rets, [True, False, False, False])
-
-    rets = [dagmc.volume_is_implicit_complement(x) for x in range(1, 5)]
-    assert_equal(rets, [False, False, False, True])
-
+    rets1 = [dagmc.volume_is_graveyard(x) for x in range(1, 5)]
+    rets2 = [dagmc.volume_is_implicit_complement(x) for x in range(1, 5)]
     md = dagmc.volume_metadata(2)
-    assert_equal(md['material'], 5)
-    assert_equal(md['rho'], 0.5)
-    assert_true(all(x in md for x in ['material', 'rho', 'imp']))
+    
+    return [rets1, rets2, md]
 
 def versions():
     from pyne import dagmc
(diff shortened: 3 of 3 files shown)


scopatz:
Hi @kkiesling - I am seeing these two errors:

```console
ERROR: test_dagmc.test_list_functions
  File "/home/scopatz/miniconda3/lib/python3.4/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/scopatz/pyne/tests/test_dagmc.py", line 307, in test_list_functions
    r = results.get()
  File "/home/scopatz/miniconda3/lib/python3.4/multiprocessing/pool.py", line 599, in get
    raise self._value
multiprocessing.pool.MaybeEncodingError: Error sending result: '[dict_keys([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18]), dict_keys([1, 2, 3, 4])]'. Reason: 'PicklingError("Can't pickle <class 'dict_keys'>: attribute lookup dict_keys on builtins failed",)'

ERROR: test_endf.test_loadtape
  File "/home/scopatz/pyne/tests/test_endf.py", line 83, in test_loadtape
    assert(os.path.isfile('endftape.100'))
AssertionError

During handling of the above exception, another exception occurred:

  File "/home/scopatz/miniconda3/lib/python3.4/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/scopatz/pyne/tests/test_endf.py", line 90, in test_loadtape
    "endftape.100")
  File "/home/scopatz/miniconda3/lib/python3.4/urllib/request.py", line 186, in urlretrieve
    with contextlib.closing(urlopen(url, data)) as fp:
  File "/home/scopatz/miniconda3/lib/python3.4/urllib/request.py", line 161, in urlopen
    return opener.open(url, data, timeout)
  File "/home/scopatz/miniconda3/lib/python3.4/urllib/request.py", line 461, in open
    req = meth(req)
  File "/home/scopatz/.local/lib/python3.4/site-packages/yt-3.0.2-py3.4-linux-x86_64.egg/yt/utilities/poster/streaminghttp.py", line 150, in http_request
    if req.has_data():
AttributeError: 'Request' object has no attribute 'has_data'

Ran 3654 tests in 15.173s

FAILED (SKIP=25, errors=2)
```

Also note that CI is probably not to be trusted right now.

kkiesling:
Hi @scopatz, I am not getting these errors on my system. I know where the test_list_functions might be getting caught up so I think I can fix that. 

That second one looks like it can't find the endftape.100 file that I added to the gitignore file. I added that because every time I ran nosetests after installing it re-added that file and was appearing as a change I should commit. 

How are you getting these errors? It says it passed the build and test here.

kkiesling:
Actually I take back what I said about being able to fix that test_list functions error.. This error message says it can't unpack a dictionary but that function doesn't even return a dictionary so I have no idea where that is coming from.

scopatz:
@kkiesling for the second one, you might need to call `list(d.keys())`, where d is a dictionary, instead of `d.keys()`. In Python 3, the keys() method works more like the iterkeys() method in Python 2.

scopatz:
for the first one, it seems like I am not even downloading the `endftape.100` file.  This looks like a yt issue.

scopatz:
I'll look into that one.

kkiesling:
That I understand. I am just confused because that test involves no dictionaries at all and I don't do anything that calls d.keys() so I can't figure out where that error is coming from. Nor can I reproduce it on my machine.

scopatz:
dagmc itself has the dictionary and is returning the keys: https://github.com/pyne/pyne/blob/develop/pyne/dagmc.pyx#L327

This should be a quick fix.  Something similar needs to happen for surfs, too, I think.

MatthewTurk:
Fixed in this pull request: https://bitbucket.org/yt_analysis/yt/pull-request/1515

scopatz:
Thanks @MatthewTurk!

kkiesling:
@scopatz I changed the d.keys(). See if it works for you now.

scopatz:
Like a charm!

scopatz:
Thanks @kkiesling!

PR #706, opened 2015-04-03
fixes render fail

scopatz:
as seen here: http://pyne.io/pyapi/xs/data_source.html#pyne.xs.data_source.SimpleDataSource.discretize

Didn't have time to re-implement latex.


crbates:
is this the expected behavior of Travis? Looks good to me.

scopatz:
We have bugs on develop that are causing issues on Travis, so it is known.

scopatz:
But this really shouldn't affect those!

crbates:
thanks @scopatz!

PR #707, opened 2015-04-06
Changed the tag naming convention for MeshTally creation to be more cons...

Pshriwise:
...istent with the error suffix used in the mesh operator. Also updated said error suffix to be more accurate about what is contained in that tag value.

### DIFF
--- a/pyne/mcnp.py
+++ b/pyne/mcnp.py
@@ -2053,9 +2053,9 @@ class MeshTally(StatMesh):
 
         if tag_names is None:
             self.tag_names = ("{0}_result".format(self.particle),
-                              "{0}_rel_error".format(self.particle),
+                              "{0}_result_rel_error".format(self.particle),
                               "{0}_result_total".format(self.particle),
-                              "{0}_rel_error_total".format(self.particle))
+                              "{0}_result_total_rel_error".format(self.particle))
         else:
             self.tag_names = tag_names
 
--- a/pyne/mesh.py
+++ b/pyne/mesh.py
@@ -1298,7 +1298,9 @@ class StatMesh(Mesh):
         # Exclude error tags because result and error tags are treated
         # simultaneously so there is not need to include both in the tag
         # list to iterate through.
-        tags = set(tag for tag in tags if not tag.endswith('_error'))
+        error_suffix = "_rel_error"
+
+        tags = set(tag for tag in tags if not tag.endswith(error_suffix))
 
         if in_place:
             mesh_1 = self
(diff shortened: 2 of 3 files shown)


scopatz:
This looks reasonable to me, but someone more familiar with the mesh stuff should probably also review (@elliottbiondo @makeclean or @kkiesling)

Pshriwise:
Taking a look at the failing tests in CI now...

scopatz:
Oh, those aren't your fault.  One is @kkiesling's and the other is this weird bins issue that happens randomly that we should try to fix.  I can't reproduce that on my machine, though @elliottbiondo has in the past.

Pshriwise:
Ahhh ok I remember that from my PR to the sandbox.

elliottbiondo:
We aren't getting any meaningful info out of CI because of the test_bins.py issue and Travis seems to give up upon receiving the first failure.

I would not be surprised if this PR breaks something in R2S though, so either CI or old-school CI (manually clone PR branch, build, run tests) needs to be done.

scopatz:
> We aren't getting any meaningful info out of CI because of the test_bins.py issue

It seems to fail randomly on CI too.  https://travis-ci.org/pyne/pyne/builds/56724604 did not fail with the bins issue, it failed later because of fluka.  Also note that this seems to be a Python 2 only issue.

> Travis seems to give up upon receiving the first failure.

Yeah, I designed it that way because otherwise Travis would report success, and I didn't want to think about the bash logic for running all of the test but it one failed, report an error.  Also, there are many nose runs because loading in many moab files was causing memory errors on Travis, which is pretty memory constrained.

Pshriwise:
Yeah I'm actually getting some broken tests locally now. The tests are still expecting the old error suffix it looks like. Is there a particular procedure for changing something like that in PyNE.

elliottbiondo:
>The tests are still expecting the old error suffix it looks like. Is there a particular procedure for changing something like that in PyNE.

How many tests are we talkin'?

@scopatz I took a look at test_bins.py  and bins.py. As I said in the Hangout today, I recall times where test_bins has failed due to array values being reversed. Likewise, it looks like a lot of reversing is going on in both test_bins.py and bins.py. I am how any single operation would yield non-deterministic behavior. In any case, the current test_bins.py does not seem to be a simple reversal:

(mismatch 100.0%)
 x: array([ 0.375])
 y: array([  4.248662e+27])

I think this is something that is going to require your expertise (I believe you wrote bins.py?)?

Pshriwise:
I'm seeing 4 tests failing. All related to mesh operators.

test_add_statmesh
test_divide_statmesh
test_multply_statmesh
test_subtract_statmesh

All with this error.
TagNotFoundError: iMesh_getTagHandle: problem getting handle for tag named 'flux_error_rel_error'  (MOAB Error Code: MB_TAG_NOT_FOUND)

Because the error suffix has changed.

elliottbiondo:
Those unit tests live in ```tests/test_mesh.py```. You should update the tag names to the correct convention there.

Pshriwise:
Kk will do. Any other documentation to go with changing a test file?

scopatz:
OK, this bins thing is almost certainly a floating point arithmetic issue.

elliottbiondo:
> Any other documentation to go with changing a test file?

No I do not think any is necessary for this case.

elliottbiondo:
(PR closed erroneously by me -- reopened)

scopatz:
I HAVE FOUND THE BUG!

Pshriwise:
Haha nice work @scopatz. Should I wait on the [bin PR](https://github.com/pyne/pyne/pull/713) to go in and add those changes?

Everything is passing now on my machine but someone should probably pull this branch and double-check, yes?

scopatz:
Probably, though I think we can relaunch the job once the bins one is in.

scopatz:
Unlike polyphemus, batlab does a merge when before running the tests.

scopatz:
Relaunching.

scopatz:
Thanks @Pshriwise!  Looks like everything is passing.  I would still feel more comfortable if someone more familiar with  mesh merged this...

elliottbiondo:
This is good. Merged.

elliottbiondo:
Thanks @Pshriwise!

PR #709, opened 2015-04-06
fixed fluka tests

kkiesling:
Fixed the assert_equal and import errors in test_fluka.py. Also added the file it writes out to the the .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ src/decay.h
 src/decay.cpp
 tests/fort.8
 tests/endftape.100
+tests/test_fluka_data.h5m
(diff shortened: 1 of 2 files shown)


scopatz:
Looks reasonable to me :) Waiting for CI.

scopatz:
This needed a relaunch.

kkiesling:
It looks like this is failing with that MOAB ERROR thing now in test_dagmc.py. It doesn't look like there are any fails or errors with the tests though, but the script still fails on test_dagmc.py..

kkiesling:
If this is getting hung up on the fact that the entire test_dagmc.py file is on separate processes, this may or may not be fixed with pr #712 ...

scopatz:
Now that #712 is in, you probably should update this branch @kkiesling. Thanks!

kkiesling:
Looks like it's just those same two tests that have been causing trouble now from test_bins

scopatz:
Looks great!  Thanks!

PR #717, opened 2015-04-21
ENDL module and unit test

arekfu:
I am starting to implement a module to read nuclear data in ENDL format. This pull request contains an embryonic module (based on endf.pyx) and the associated unit test.


pyne/endl.py
```diff
@@ -0,0 +1,239 @@
...
+
+from warnings import warn
+from pyne.utils import QAWarning
+
+import pyne.rxdata as rx
+import pyne.utils as utils
+from pyne import nucname
+
+import numpy as np
+import re
```
scopatz: PEP8: Import ordering.  First there are Python standard library errors (re), then third party libraries (numpy), then current package (pyne)

tests/test_endl.py
```diff
@@ -0,0 +1,58 @@
...
+        urllib.urlretrieve("https://www-nds.iaea.org/epdl97/data/endl/eedl/za082000",
+                    "epdl97_eedl_Pb")
+    from hashlib import md5
+    with open("epdl97_eedl_Pb", "rb") as f:
+        obs_hash = md5(f.read()).hexdigest()
+    exp_hash = "502105669e0c0ad917301d219c649aaf"
+    try:
+        assert_equal(obs_hash, exp_hash)
+    except AssertionError:
+        raise AssertionError("epdl97_eedl_Pb hash check failed; please try redownloading the epdl97_eedl_Pb data file.")
```
scopatz: PEP8: 80 character limit.

pyne/endl.py
```diff
@@ -0,0 +1,239 @@
...
+https://www.oecd-nea.org/dbdata/data/manual-endf/nds_eval_eedl.pdf
+
+For more information, contact Davide Mancusi <davide.mancusi@cea.fr>.
+"""
+from __future__ import print_function, division, unicode_literals
+
+from warnings import warn
+from pyne.utils import QAWarning
+
+import pyne.rxdata as rx
```
scopatz: I mistook this as a `rxname` because [our style guide says that rx stands for "reaction"](http://pyne.io/devsguide/style_guide.html#variable-name-conventions).  Would you mind just doing `from pyne import rxname`? Thanks!

pyne/endl.py
```diff
@@ -0,0 +1,239 @@
...
+#            fh = open(self.fh, 'r')
+#            opened_here = True
+#        else:
+#            fh = self.fh
+#        fh.readline()
+#        data = fromendf_tok(fh.read())
+#        fh.seek(0)
+#        if opened_here:
+#            fh.close()
+#        return data
```
scopatz: Remove commented out code.

scopatz: Please.

pyne/endl.py
```diff
@@ -0,0 +1,239 @@
...
+#                            }})
+
+            # skip to the end of the table
+            line = fh.readline()
+            while line and not END_OF_TABLE_RE.match(line):
+                line = fh.readline()
+
+
+
+    def _linlin(self, e_int, xs, low, high):
```
scopatz, line 144: These functions don't use self at all and could / should be moved off of the class and then have unit tests.

pyne/endl.py
```diff
@@ -0,0 +1,239 @@
...
+
+            # insert the nucleus in the self.structure dictionary
+#            if nuc not in self.structure:
+#                self.structure.update(
+#                        { nuc: {
+#                            'particle_in': [],
+#                            'particle_out': [],
+#                            'data': {},
+#                            'start': start,
+#                            }})
```
scopatz: Remove commented out lines.

tests/test_endl.py
```diff
@@ -0,0 +1,58 @@
+import os
+import io
+import warnings
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
+from math import e
```
scopatz: No need for this import.  Just use `np.e`

tests/test_endl.py
```diff
@@ -0,0 +1,58 @@
+import os
+import io
+import warnings
+try:
+    from StringIO import StringIO
+except ImportError:
```
scopatz: I am assuming that this is because of Python 2/3 compatibility.  In this case you should use something like the following.

``` python
import sys
if sys.version_info[0] > 2:
   from io import StringIO
else:
   from StringIO import StringIO
```

pyne/endl.py
```diff
@@ -0,0 +1,239 @@
...
+        else:
+            fh = self.fh
+
+        while True:
+            # get header lines
+            line1 = fh.readline()
+            line2 = fh.readline()
+
+            # EOF?
+            if not line2:
```
scopatz: It is better to test `if  len(line2) != 0` than `if not ''`

pyne/endl.py
```diff
@@ -0,0 +1,239 @@
+#!/usr/bin/env python
```
scopatz: This module has no main function and should remove this line.

pyne/endl.py
```diff
@@ -0,0 +1,239 @@
...
+#        fh.readline()
+#        data = fromendf_tok(fh.read())
+#        fh.seek(0)
+#        if opened_here:
+#            fh.close()
+#        return data
+
+    def _read_headers(self):
+        opened_here = False
+        if isinstance(self.fh, basestring):
```
scopatz, line 79: basestring is not Python 3 compatible. In other modules in pyne, this variable is defined as str in the event of Python 3.

pyne/endl.py
```diff
@@ -0,0 +1,239 @@
...
+            fh = open(self.fh, 'r')
+            opened_here = True
```
scopatz, line 81: Where is the file handle closed if we opened it?

pyne/endl.py
```diff
@@ -0,0 +1,239 @@
...
+#                        { nuc: {
+#                            'particle_in': [],
+#                            'particle_out': [],
+#                            'data': {},
+#                            'start': start,
+#                            }})
+
+            # skip to the end of the table
+            line = fh.readline()
+            while line and not END_OF_TABLE_RE.match(line):
```
scopatz: `while len(line) > 0 ...`

pyne/endl.py
scopatz, line 144: It also looks like these might be copied from ENDF or somewhere.  Is that true?  If it is, then we should make these functions first class and put them somewhere else, perhaps in bins.py or their own module?

pyne/endl.py
```diff
@@ -0,0 +1,239 @@
...
+        x2 = e_int[1:]
+        y1 = xs[:-1]
+        y2 = xs[1:]
+        A = - np.log(y2/y1)/np.log(x1/x2)
+        B = - (np.log(y1)*np.log(x2) - np.log(y2)*np.log(x1))/np.log(x1/x2)
+        return np.nansum(np.e**B / (A+1) * (x2**(A+1) - x1**(A+1))/de_int)
+
+
+
+class NotFound(Exception):
```
scopatz: This class isn't used anywhere and probably should be removed.  Also, I am wary of new exceptions :).

tests/test_endl.py
```diff
@@ -0,0 +1,58 @@
...
+
+from pyne.utils import QAWarning
+warnings.simplefilter("ignore", QAWarning)
+
+from pyne.endl import Library
+
+import nose
+from nose.tools import assert_equal
+
+def ignore_future_warnings(func):
```
scopatz: What is the purpose of this function?  It isn't used.

tests/test_endl.py
```diff
@@ -0,0 +1,58 @@
...
+    return new_func
+
+def test_loadfile():
+    try:
+        assert(os.path.isfile('epdl97_eedl_Pb'))
+    except AssertionError:
+        try:
+            import urllib.request as urllib
+        except ImportError:
+            import urllib
```
scopatz: This import should happen at the top of the module and should use the `sys.version_info` check.  See other test modules for examples.

tests/test_endl.py
```diff
@@ -0,0 +1,58 @@
...
+        urllib.urlretrieve("https://www-nds.iaea.org/epdl97/data/endl/eedl/za082000",
+                    "epdl97_eedl_Pb")
```
scopatz: Indentation line up.

tests/test_endl.py
```diff
@@ -0,0 +1,58 @@
...
+            return func(*args, **kwargs)
+    new_func.__name__ = func.__name__
+    new_func.__doc__ = func.__doc__
+    new_func.__dict__.update(func.__dict__)
+    return new_func
+
+def test_loadfile():
+    try:
+        assert(os.path.isfile('epdl97_eedl_Pb'))
+    except AssertionError:
```
scopatz: This should be an if-statement rather than an assertion.

tests/test_endl.py
```diff
@@ -0,0 +1,58 @@
...
+            import urllib
+        urllib.urlretrieve("https://www-nds.iaea.org/epdl97/data/endl/eedl/za082000",
+                    "epdl97_eedl_Pb")
+    from hashlib import md5
+    with open("epdl97_eedl_Pb", "rb") as f:
+        obs_hash = md5(f.read()).hexdigest()
+    exp_hash = "502105669e0c0ad917301d219c649aaf"
+    try:
+        assert_equal(obs_hash, exp_hash)
+    except AssertionError:
```
scopatz: No need to catch this.  I think you can pass in the message to `assert_equal()`

tests/test_endl.py
```diff
@@ -0,0 +1,58 @@
...
+        raise AssertionError("epdl97_eedl_Pb hash check failed; please try redownloading the epdl97_eedl_Pb data file.")
+    testlib = Library("epdl97_eedl_Pb")
+#    exp_nuclides = [10010000, 30060000, 40090000, 50110000, 30070000, 60000000, 50100000]
+#    obs_nuclides = list(map(int, testlib.structure.keys()))
+#    assert_array_equal(exp_nuclides, obs_nuclides)
```
scopatz: Are these commented out for a reason?  Some quantitative tests would be nice.

scopatz:
Hi @arekfu!  Thanks for opening this.  I think ENDL support would be super valuable.  Also, thanks for keeping the initial code small.  I have done a pretty thorough review.  There are a couple of style issues which should be easy to fix.

scopatz:
Also, I didn't get to add this before, but I think there are some architectural issues that are important but not difficult to address.  It is good that we can have a discussion early.  I hope that this conversation helps!

pyne/endl.py
arekfu, line 144: Yes, they are copied from ENDF. We should probably move them.

pyne/endl.py
arekfu: I will soon uncomment and rework it.

pyne/endl.py
arekfu: You mean `from pyne import rxdata`, I guess? This line is also straight from ENDF, though.

arekfu:
Hi @scopatz, thank you for your review. Many of your remarks actually apply to endf.pyx and test_endf.py, since I copy-pasted and modified most of the code.

pyne/endl.py
arekfu: Already uncommented in future commit.

pyne/endl.py
scopatz: OK.  It should probably be fixed there too.  (Though that isn't something you need to do.)

scopatz:
> Hi @scopatz, thank you for your review. Many of your remarks actually apply to endf.pyx and test_endf.py, since I copy-pasted and modified most of the code.

I am sure :).  I'll make a different issue to do a style review of that module as well.

pyne/endl.py
arekfu: Why? Just curious.

scopatz: It is a more precise and explicit test.  What you care about here is that the list is zero, not that the the boolean happens to evaluate to false.  It is mostly for readability.

tests/test_endl.py
arekfu: They are commented out because, as things stand, `endl.Library` does not fill its internal structure dictionary (yet). The commented code is there to remind me to add the test as soon as the functionality is implemented.

tests/test_endl.py
arekfu: I'm a bit surprised that the `pep8` checker did not pick this up!

pyne/endl.py
gonuke, line 144: This is one of a few different interpolation functions.  Do any of the already included modules - I'm thinking numpy here - have interpolation routines already?  If not, these should perhaps live in some utility module somewhere.

scopatz:
Also, @arekfu - let us know when you'd like us to take a look again!

pyne/endl.py
scopatz, line 144: > This is one of a few different interpolation functions. Do any of the
> already included modules - I'm thinking numpy here - have interpolation
> routines already? If not, these should perhaps live in some utility module
> somewhere.

NumPy does have interpolations functions but I think that the value here is
that they set the bounds correctly to match what is in the spec.  Some of
the implementation could possibly be cleaned up with that in mind.

arekfu:
Hi @scopatz, the commits above introduce parsing of the ENDL table headers. I have also beefed up the unit test. Can you please take a look at this? Sorry if this is taking some time, I am a bit busy right now.

scopatz:
Hi @arekfu - no worries on the delay :).  It looks like this is not Python 3 compliant.  I'll take a look.

pyne/endl.py
```diff
@@ -0,0 +1,264 @@
...
+
+    DataTuple = namedtuple('DataTuple', ['yo', 'limits', 'x1', 'data'])
+
+    @staticmethod
+    def pin_rdesc_rprop_dict_entry():
+        """Static method to generate entries for the pin_rdesc_rprop dict."""
+
+        return {
+                'data_tuples': list()
+                }
```
scopatz: Please put this one line and use a list literal.  ie:

``` python
return {'data_tuples': []}
```

Also, I am not sure something so simple needs its own function, but I am not opposed to it.

pyne/endl.py
```diff
@@ -0,0 +1,264 @@
...
+
+    @staticmethod
+    def structure_dict_entry():
+        """Static method to generate entries for the structure dict."""
+
+        return {
```
scopatz, line 54: Please remove the newline.

pyne/endl.py
```diff
@@ -0,0 +1,264 @@
...
+#        fh.readline()
+#        data = fromendf_tok(fh.read())
+#        fh.seek(0)
+#        if opened_here:
+#            fh.close()
+#        return data
+
+    def _read_headers(self):
+        opened_here = False
+        if isinstance(self.fh, str):
```
scopatz: The str here probably needs to be python 3 compatible.  Look in other modules for code that looks like 

``` python
if sys.version_info[0] > 2:
    basestring = str
```

and then test against basestring here.

pyne/endl.py
```diff
@@ -0,0 +1,264 @@
...
+
+warn(__name__ + ' is not yet QA compliant.', QAWarning)
+
+END_OF_TABLE_RE = re.compile(' {71}1')
+
+
+class Library(rxdata.RxLib):
+    """A class for a file which contains multiple ENDL tables."""
+
+    DataTuple = namedtuple('DataTuple', ['yo', 'limits', 'x1', 'data'])
```
scopatz: Probably best to move this to global scope rather than living on the library class.  Also nice use of namedtuple :)

tests/test_endl.py
```diff
@@ -0,0 +1,66 @@
...
+if sys.version_info[0] > 2:
+    import urllib.request as urllib
+else:
+    import urllib
+
+
+def test_loadfile():
+    if not os.path.isfile('epdl97_eedl_Pb'):
+        urllib.urlretrieve(
+                "https://www-nds.iaea.org/epdl97/data/endl/eedl/za082000",
```
scopatz: This isn't a big deal but our style guide says to use single quotes for string literals (rather than double quotes) when it makes sense to do so. http://pyne.io/devsguide/style_guide.html#python-style-guide

scopatz:
Thanks @arekfu. This is getting very close.  I think you just need to fix the issue that is causing the Python 3 failure.  For reference you can see the failing test here: https://travis-ci.org/pyne/pyne/jobs/62099575#L2614

pyne/endl.py
arekfu: I need a factory function for `defaultdict`. It can be a `lambda`, though.

scopatz: Sounds good to me.

arekfu:
Yay! Python 3 works now.

arekfu:
Another batch of updates:
- the Python 3 tests run fine;
- the module is able to do some serious parsing;
- new `fromendl_tok` function in the utils module.

Also, I gave up the idea of caching parsed data in the `Library` object because I realised that this is much better done at a higher level, by the to-be-written `ENDLDataSource` class.

The `Library` class seems more or less complete to me. However, I am not sure about the way I should handle double-differential data (such as angle-energy differential cross sections for photon emission in bremsstrahlung). For the moment I just return a big, 3-column `ndarray`. Perhaps it would make more sense to clearly expose the double-differential structure, but I'm not sure how to do this. How is this handled in the ENDF pipeline?

scopatz:
> Perhaps it would make more sense to clearly expose the double-differential structure, but I'm not sure how to do this. How is this handled in the ENDF pipeline?

I don't know... But it is probably a 2D numpy array.

pyne/endl.py
```diff
@@ -0,0 +1,341 @@
...
+warn(__name__ + ' is not yet QA compliant.', QAWarning)
+
+if sys.version_info[0] > 2:
+    basestring = str
+
+END_OF_TABLE_RE = re.compile(' {71}1')
+
+DataTuple = namedtuple('DataTuple', ['yo', 'limits', 'x1'])
+
+nfields_rprop = {
```
scopatz: PEP8 global constants should be in all upercase

pyne/endl.py
```diff
@@ -0,0 +1,341 @@
...
+        -------
+        array, float64
+            float64 NumPy array containing the reaction data. The shape of the
+            array depends on the ENDL reaction property.
+        """
+        opened_here = False
+
+        try:
+            pdp_dict = self.structure[nuc]['pin_rdesc_rprop']
+        except KeyError as e:
```
scopatz, line 307: It is generally better to use `if key in ...` when you can rather than relying on exception handling.

scopatz, line 307: Especially since you are just reraising the error.  I would rewrite this as:

``` python
if nuc in self.structure and 'pin_rdesc_rprop' in self.structure[nuc]:
    pdp_dict = self.structure[nuc]['pin_rdesc_rprop']
else:
    msg = ....
    raise KeyError(msg)
```

pyne/endl.py
```diff
@@ -0,0 +1,341 @@
...
+            raise e
+
+        data_tuples = pdp_dict[p_in, rdesc, rprop]['data_tuples']
+
+        # Select the first data_tuple matching x1 and yo, if they are provided
+        # (i.e. not None)
+        def match_x1_yo_to_data_tuple(x, y, d):
+            return (x is None or d.x1 == x) and (y is None or d.yo == y)
+        try:
+            index, data_tuple = next((i, d) for i, d in enumerate(data_tuples)
```
scopatz, line 320: Can't you just test if the length of `data_tuples` >0

tests/files_test_endl/testfile
```diff
@@ -0,0 +1,2545 @@
+ 82000  9  0  0.00000+ 0 8912015 2  0.00000+ 0  0.00000+ 0  0.00000+ 0  
+ 7  0  0  0.00000+ 0  0.00000+ 0  0.00000+ 0  0.00000+ 0  0.00000+ 0
```
scopatz, line 2: Where does this file come from and are you/we allowed to redistribute it?

scopatz:
This is looking great @arekfu! Just a couple of minor issues to address.  Do you want a us to wait for the merge until after the double differential issue is addressed? Or would you prefer that to be in a separate pull request.

tests/files_test_endl/testfile
arekfu, line 2: I started from an EPDL data file (the one downloaded by the unit test), removed some tables and randomized the data in the remaining ones. I guess it's OK?

pyne/endl.py
arekfu, line 320: No, I'm looking for the first data tuple that matches `x1`, `yo` or both, if they are not `None`. The `enumerate` is useless, however. I used it for the caching but then forgot to remove it.

pyne/endl.py
arekfu, line 307: Guess where this code snippet comes from :)

I mildly disagree. I think that the `try`/`except` style is more [pythonic](https://docs.python.org/3.4/glossary.html#term-eafp). Interestingly, dictionary lookup is exactly the example that the [LBYL](https://docs.python.org/3.4/glossary.html#term-lbyl) entry puts forward.

arekfu:
I'd like to fix these points first. I would also like to check if there is an easy solution for the interpolation of the double-differential data.

pyne/endl.py
scopatz, line 307: So I always interpreted that to mean that 

``` python
pdp_dict = self.structure[nuc]['pin_rdesc_rprop'] 
```

on its on is more Pythonic (which it is), and if an error occurs you catch it somewhere farther up in the stack. Also the EAFP comparison is a little disingenuous since C exception handling syntax. However, if you are going to test for inclusion at the call site (though exceptions or conditionals), I think that there is an argument to be made for using conditionals based on their timings.  Consider the following:

``` python
scopatz@athenaie ~ $ def f(test):
`·.,¸,.·*¯`·.,¸,.·*¯     if test:
`·.,¸,.·*¯`·.,¸,.·*¯         pass
`·.,¸,.·*¯`·.,¸,.·*¯     else:
`·.,¸,.·*¯`·.,¸,.·*¯         pass
`·.,¸,.·*¯`·.,¸,.·*¯ 
scopatz@athenaie ~ $ def g(test):
`·.,¸,.·*¯`·.,¸,.·*¯     try:
`·.,¸,.·*¯`·.,¸,.·*¯         assert test
`·.,¸,.·*¯`·.,¸,.·*¯     except AssertionError:
`·.,¸,.·*¯`·.,¸,.·*¯         pass
`·.,¸,.·*¯`·.,¸,.·*¯ 
scopatz@athenaie ~ $ timeit "f(True)"
The slowest run took 130.28 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 76.8 ns per loop
scopatz@athenaie ~ $ timeit "f(False)"
10000000 loops, best of 3: 75.8 ns per loop
scopatz@athenaie ~ $ timeit "g(True)"
10000000 loops, best of 3: 91.4 ns per loop
scopatz@athenaie ~ $ timeit "g(False)"
The slowest run took 59.51 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 387 ns per loop
```

The False case for exceptions is _much_ slower.

tests/files_test_endl/testfile
scopatz, line 2: Yup that is OK, since it is randomized, thanks!

pyne/endl.py
scopatz, line 320: But next() is only called once here... so it will either get the first element or nothing. Right?

pyne/endl.py
arekfu, line 307: > The False case for exceptions is much slower

That is true in most languages, I guess. But the point here is re-raising the exception with a more informative error message; so I would argue timing does not matter.

pyne/endl.py
arekfu, line 320: That's right. I want the first element matching `x1` (if given) and `yo` (if given).

scopatz, line 320: Ahh ok.

scopatz:
Hey @arekfu - please let us know when this is ready!

arekfu:
OK, @scopatz, let's do this. I guess we can merge if you agree.

scopatz:
ooof I am so sorry about the delay in looking at this @arekfu. I have been on vacation ~_~.

scopatz:
Thanks a ton @arekfu!

scopatz:
And sorry again about the delay!

PR #722, opened 2015-04-22
nucname notebook update

elliottbiondo:
Fixed a few typos and changed the `id()` ambiguity example to a more practical one. Note that I made these changes from PyNE VB, so there should not be any ipython version issues.


elliottbiondo:
I also updated the `Material` notebook. This mostly involved rewording and rearranging.

It would be very nice to have this and #719 merged in before 1:30 PM EST that way I can update pyne.io prior to the tutorial session. @rachelslaybaugh or mzweig (not sure why @ is not working for mzweig) maybe you could take a look. Or @scopatz the notebook progenitor :) Or maybe this counts as a "documentation change" and I can merge it in myself? :)

Issue #721, opened 2015-04-22
ENDF module chokes on Windows-style CR-LF-terminated (\r\n) data files


arekfu:
Example session follows. Here `photonuc.Fe56.bcd` is `\r\n`-terminated (Windows style) and `photonuc_unix.Fe56.bcd` is `\n`-terminated (*nix style).

``` bash
$ ls -lFh photonuc*
-rw-r--r-- 1 ******** **** 949K Apr 21 18:43 photonuc.Fe56.bcd
-rw-r--r-- 1 ******** **** 937K Apr 21 18:46 photonuc_unix.Fe56.bcd
$ ipython
[snipping message]
In [1]: from pyne.endf import Library
[snipping QA warning]
In [2]: l=Library('photonuc_unix.Fe56.bcd')

In [3]: l_win=Library('photonuc.Fe56.bcd')
 Something went wrong converting:
 0.000000+
 Something went wrong converting: 0 0.000000+
 Something went wrong converting:
 0.000000
 Something went wrong converting: 3
 0.00000
 Something went wrong converting: 0+0 0.00000
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-3-6d0be553c33f> in <module>()
----> 1 l_win=Library('photonuc.Fe56.bcd')

pyne/endf.pyx in pyne.endf.Library.__init__ (/path/to/pyne/build/pyne/endf.cxx:4541)()

pyne/endf.pyx in pyne.endf.Library._read_headers (/path/to/pyne/build/pyne/endf.cxx:6585)()

pyne/endf.pyx in pyne.endf.Library._read_mat_flags (/path/to/pyne/build/pyne/endf.cxx:6796)()

pyne/endf.pyx in pyne.endf.Library.get_rx (/path/to/pyne/build/pyne/endf.cxx:22627)()

pyne/endf.pyx in pyne.endf.Library._read_nucmfmt (/path/to/pyne/build/pyne/endf.cxx:23092)()

KeyError: u'MT 451 not found in File 1.'
```

scopatz:
Ahh, so some `open()` calls probably need to use universal newlines (the 'u' flag).  see https://docs.python.org/2/library/functions.html#open  If you could hunt these down and open a pull request, that would be awesome!  Thanks!

arekfu:
[...]

PR #723, opened 2015-04-23
ENDF endlines

arekfu:
Fix issue #721 .


scopatz:
Thanks a ton @arekfu!

PR #730, opened 2015-04-28
Get cell number from bit array

everythingfunctional:
- Figured out how the MCNP cell number is stored in the bit
  array in the Surface Source file

### DIFF
--- a/pyne/mcnp.py
+++ b/pyne/mcnp.py
@@ -457,6 +457,7 @@ class SurfSrc(_BinaryReader):
             track_data.record = track_info.get_double(abs(self.ncrd))
             track_data.nps = track_data.record[0]
             track_data.bitarray = track_data.record[1]
+            track_data.cell = int(abs(track_data.bitarray) / 8 % 1e8)
             track_data.wgt = track_data.record[2]
             track_data.erg = track_data.record[3]
             track_data.tme = track_data.record[4]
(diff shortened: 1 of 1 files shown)


gonuke:
Is there a reference for this, or did you dig around in MCNP source code?  I am only wondering because the "% 1e8" doesn't look like a standard bit array operation.  (Where as the "/ 8" does look like a standard bit array operation.)

everythingfunctional:
I had to dig around in the MCNP source code to figure it out. "% 1e8" is modulo 1e8. They did some very convoluted things to pack that bit array. It worked for a problem I had a couple years ago with MCNP5 1.6, but it would definitely be worthwhile if someone could put together a very simple problem to check it.

gonuke:
Most bit array operations work in terms of powers of 2, so that's what made we wonder about the module 1e8.  A test will definitely be useful/necessary.

scopatz:
This failed for unrelated reasons.  relaunched.

pyne/mcnp.py
```diff
@@ -457,6 +457,7 @@ def read_tracklist(self):
             track_data.record = track_info.get_double(abs(self.ncrd))
             track_data.nps = track_data.record[0]
             track_data.bitarray = track_data.record[1]
+            track_data.cell = int(abs(track_data.bitarray) / 8 % 1e8)
```
scopatz: I think it might be better to do 

``` python
track_data.cell = abs(track_data.bitarray) // 8 % 100000000
```

Note the `//` for python 3

scopatz:
Looks like the build passes @bprichar

scopatz:
Did you want to change the operators?

makeclean:
Is this all good to go?

scopatz:
I would say it is.

makeclean:
Awesome, merging.

PR #731, opened 2015-04-29
Fixes #725

### LINKED ISSUE #725: System Install Fails Test
If pyne is installed as a system administrator (i.e. sudo setup.py install), one of the test cases fails when executed as a user, presumably because it does not have access to some file. This may not be of particular importance, just thought I'd mention it.

everythingfunctional:
- We'll make a local copy of nuc_data in case we don't have
  permissions to write to the original

### DIFF
--- a/tests/test_data_checksums.py
+++ b/tests/test_data_checksums.py
@@ -1,6 +1,7 @@
 """ test data_checksums and hashing functions"""
 import os
 import warnings
+from shutil import copyfile
 
 import nose
 from nose.tools import assert_equal, assert_true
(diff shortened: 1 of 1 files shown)


tests/test_data_checksums.py
```diff
@@ -35,9 +35,15 @@ def test_data_checksums():
 
 def test_internal_hashes():
     from pyne.dbgen import hashtools
-    hashtools.set_internal_hashes(pyne.nuc_data)
-    for item, val in hashtools.check_internal_hashes(pyne.nuc_data):
+    from shutil import copyfile
```
scopatz: Import should go at the top of the file.

tests/test_data_checksums.py
```diff
@@ -35,9 +35,15 @@ def test_data_checksums():
...
+    test_data = './test_nuc_data.h5'
```
scopatz: The `./` here is not needed and platform dependent. Please remove.

tests/test_data_checksums.py
```diff
@@ -35,9 +35,15 @@ def test_data_checksums():
...
+    # Create a copy so we don't try to modify a file we may not have
+    # permissions for
+    copyfile(pyne.nuc_data, test_data)
```
scopatz: Please only copy the file if the user doesn't have write permission.  Right now it copies it all the time.  I think that you can use `os.stat()` to get the permissions.

tests/test_data_checksums.py
```diff
@@ -35,9 +35,15 @@ def test_data_checksums():
...
+    hashtools.set_internal_hashes(test_data)
+    for item, val in hashtools.check_internal_hashes(test_data):
         assert_true(val)
+    os.remove(test_data)
```
scopatz: This is only needed if the file was copied.

scopatz:
Made a few comments, thanks for putting this in!

everythingfunctional:
Made the changes you suggested.

scopatz:
Really weird test failures.  Relaunching.

scopatz:
The build errors are from uploading a new conda package.  I have put in a fix to #734  and pyne/pyne-ci#29.

scopatz:
Relaunching after fix.

scopatz:
Thanks a lot @bprichar! Sorry we ran into the CI issues here.

everythingfunctional:
No problem. Glad I could help.

PR #736, opened 2015-05-28
Improve mcnp mat

makeclean:
This PR improves the existing capabilities of mats_from_inp in the mcnp.py file. Previous to this PR the function would fail on materials like

```
M1  1001.21c   1.0
    26056.21c  0.1
C I LIKE PUTTING COMMENTS IN THE MIDDLE OF MATERIALS
    28058.21c  0.1
```

It would only read part of the material and stop at the first 'C'.

This PR also addresses the shortcoming (in the reader) that MCNP allows the user to have the same nuclide multiple times, like

```
M1  1001.21c   1.0
    26056.21c  0.1
C I LIKE PUTTING COMMENTS IN THE MIDDLE OF MATERIALS
    26056.21c  0.1
C I JUST CANT GET ENOUGH IRON
    28058.21c  0.1
C srsly I love it
    26056.21c  0.1
```

The added test ensures that:
1. The routine continues to read data even if it encounters a 'c' or 'C'
2. That multiple entries of the same nuclide are summed rather than overwritten.

The tests pass on my machine.

### DIFF
--- a/pyne/mcnp.py
+++ b/pyne/mcnp.py
@@ -1389,12 +1389,19 @@ def mat_from_inp_line(filename, mat_line, densities='None'):
     # collect all material card data on one string
     line_index = 1
     line = linecache.getline(filename, mat_line + line_index)
-    while line[0:5] == '     ':
+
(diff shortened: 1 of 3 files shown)


scopatz:
Thanks @makeclean!

PR #740, opened 2015-06-30
Duplicate keys in ace.Library

sellitforcache:
Hi guys!

The new naming schemes in the ENDFB/VII data give them two names (eg 1001.80c in xsdir and 1001.710nc as the filename).  The current ace module will name it as the filename only (1001.710nc) in the library, so entering a key of '1001.80c' doesn't work.  This small change fixes that.  

I did this for WARP, and figured it might be helpful for other people too.


scopatz:
Hi @sellitforcache! Thanks for putting this in.  There seems to be an build error. I am going to try relaunching.  (Also, sorry for the slow reply.... vacation ~_~)

pyne/ace.pyx
```diff
@@ -331,6 +333,8 @@ class Library(object):
                 temp_in_K = round(temp * 1e6 / 8.617342e-5)
                 print("Loading nuclide {0} at {1} K".format(name, temp_in_K))
             self.tables[name] = table
+            if name_old:
```
scopatz: This case will also be skipped if `name_old` is a str and has length 0.  Is this desirable?

sellitforcache:
Hi Anthony,

I'd say having the case skipped on len=0 is desirable.  As for the build error... I'm not sure what it's about.  Who is travis, why doesn't he have HDF5, and why is he trying to import pyne from the build directory?

sellitforcache:
There is an error for loading sab tables this way.  I will fix and request a new pull.  But is there something else I should be doing to make the test build work correctly?

scopatz:
Relaunching.

scopatz:
Hi @sellitforcache - Also is there any way you could add a test, please, for the new naming convention.  Thanks!

scopatz:
Also, super sorry about the delay stemming from CI breakage.

makeclean:
@sellitforcache is there any way you could add some tests for your feature?

PR #741, opened 2015-07-02
Do not merge!

scopatz:
Travis builds seem to be failing for HDF5 related reasons. This is only a test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,11 @@ python:
   - "2.7"
   - "3.4"
 before_install:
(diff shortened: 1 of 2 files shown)


scopatz:
Yup, there is clearly something wrong.

scopatz:
This seems to be an issue with HDF5 v1.8.15, which I can thankfully reproduce locally

scopatz:
Ahh, I think it is because I need to update the MOAB package.

scopatz:
New error :(

scopatz:
Yes!  CI is working again!

PR #748, opened 2015-08-13
Rtflux

elliottbiondo:
This PR implements the capability of reading 1D, 2D, and 3D RTFLUX binary files and creating corresponding PyNE Meshes, for use with PARTISN. The ATFLUX (adjoint) file is also supported, as it is the same as RTFLUX.

For 1D, the binary format is slightly different than 2D/3D (as specified [here](http://t2.lanl.gov/codes/transx-hyper/rtflux.html)). Strangely, this format did not seem to work using the PyNE BinaryReader (though I verified it works in C++). However with the PyNE BinaryReader, using the 3D specification for 1D strangely DOES work. I tested this with up to 7000 mesh elements, so I think everything is fine. I left some comments in the code to mention that I do this.

1D, 2D, and 3D RTFLUXs are all tested.


scopatz:
Hey @elliottbiondo - this looks great. Can you please add a link to the spec that you mentioned somewhere in the docs, though?  Or is it there somewhere already?

elliottbiondo:
http://t2.lanl.gov/nis/codes/transx-hyper/rtflux.html

elliottbiondo:
Oh sorry, I read that too fast. The link is at the top of the file.

elliottbiondo:
Maybe I should add the link to the class so it gets incorporated into the API documentation?

scopatz:
Yes, that is what I was suggesting :). Sorry for being unclear.

elliottbiondo:
Done :)

scopatz:
Thanks a ton @elliottbiondo!

PR #749, opened 2015-08-14
PARTISN SOURCF

elliottbiondo:
This PR adds the capability of generating isotropic sources for PARTISN from a supplied PyNE `Mesh`. This is done via the PARTISN SOURCF card. Outputted cards are wrapped to 80 characters and make use of the repeated value shorthand format for zero entries (e.g. 3R 0 = 0 0 0).

The SOURCF card produced in the nosetest was confirmed work with PARTISN. I checked the resulting RTFLUX output in VISIT and was able to determine the the position/energy iteration order was all correct. An example of one such visual confirmation is shown below.

image

### DIFF
--- a/pyne/partisn.py
+++ b/pyne/partisn.py
@@ -830,3 +830,37 @@ def strip_mat_name(mat_name):
         tmp1 = ''.join(tmp2)
     
     return tmp1
+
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks @elliottbiondo! Looks good.

PR #751, opened 2015-08-17
Pointwise collapse

elliottbiondo:
This PR expands on the existing bins.pointwise_linear_collapse to include lin-lin, lin-log, log-lin, and log-log collapsing. Some additional input validity checks were also added.

### DIFF
--- a/pyne/bins.pyx
+++ b/pyne/bins.pyx
@@ -152,9 +152,16 @@ def pointwise_collapse(np.ndarray[np.float64_t, ndim=1] x_g,
     cdef int n0, n1  # current point index
     cdef double val, ylower, yupper
     cdef np.ndarray[np.float64_t, ndim=1] y_g = np.empty(G, dtype='float64')
+
+    # Ensure input is monotonically increasing/decreasing
+    for z in [x_g, x]:
+        if not (np.all(np.diff(z) > 0.) or np.all(np.diff(z) < 0)):
+            raise ValueError("x and x_g arrays must be monotonically"  
+                             "increasing/decreasing.")
+
     reversed = False
     if x_g[0] > x_g[-1]:
-        # monotonically decreaing, make increasing so logic is simpler
+        # monotonically decreasing, make increasing so logic is simpler
         x_g = x_g[::-1]
         x = x[::-1]
         y = y[::-1]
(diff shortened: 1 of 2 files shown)


pyne/bins.pyx
```diff
@@ -118,19 +118,66 @@ def pointwise_linear_collapse(np.ndarray[np.float64_t, ndim=1] x_g,
...
+    ----------
+    x_g : array-like
+        Group boundaries, length G+1 for G groups, must be monotonic in the 
+        same direction as x.
+    x : array-like
+        Pointwise abscissa to be collapsed, must be monotonic in the same direction
+        as x_g and have the same length as y.
+    y : array-like
+        Pointwise data to be interpolated, must have the same length as x.
+    log : tuple of bools, optional, default = (False, False)
```
scopatz: This is a little strange.  Most APIs have `log -> loglog`, `logx | xlog -> loglin`, and `logy | ylog -> linlog`.  Only one of these three options should be not None.

scopatz: errr, not exactly, sorry.  If logx and logy are set, they override log.

pyne/bins.pyx
```diff
@@ -118,19 +118,66 @@ def pointwise_linear_collapse(np.ndarray[np.float64_t, ndim=1] x_g,
...
         x_g = x_g[::-1]
         x = x[::-1]
         y = y[::-1]
         reversed = True
+
+    # Handle logrithmic interpolations
+    if log[0]:
+        if x_g[0] <= 0.0 or x_g[0] <= 0.0:
+            raise ValueError("x values must be positive for logrithmic interpolation")
+        x_g = np.array([np.log(z) for z in x_g])
```
scopatz: This is really inefficient to have a list comprehension here.  Just use the numpy function:

``` python
x_g = np.log(x_g)
```

Same for all below.

pyne/bins.pyx
```diff
@@ -157,5 +204,10 @@ def pointwise_linear_collapse(np.ndarray[np.float64_t, ndim=1] x_g,
         y_g[g0] = val / (x_g[g1] - x_g[g0])
     if reversed:
         y_g = y_g[::-1]
+
+    # Handle logrithmic interpolation
+    if log[1]:
+        y_g = np.array([np.exp(z) for z in y_g])
```
scopatz: Same here, just use numpy's exp()

scopatz:
Some API and numpy issues, but other than that, looks good.  Thanks for the great tests.

pyne/bins.pyx
elliottbiondo: Should `log`, `logx`, `logy` be bools? If they aren't and they are set to `None`, then what are they set to to turn them on?

scopatz: Yes, they should be bools or None.  If all of them are None, then it should be assumed to be linlin.

elliottbiondo:
@scopatz I think I got everything fixed. I am not sure I got exactly what you were suggesting with "bool or None", so let me know if I am still off the mark.

scopatz:
This looks great, thanks @elliottbiondo

PR #753, opened 2015-08-29
mats_from_inp

elliottbiondo:
This PR fixes a bug that affected the way `mcnp.mats_from_inp` read in materials with multiple densities in order to create `MultiMaterials`.  A test was added to ensure this is fixed.

### DIFF
--- a/pyne/mcnp.py
+++ b/pyne/mcnp.py
@@ -1344,7 +1344,7 @@ def mats_from_inp(inp):
            else:
                same_bool = False
                for j in range(0, len(densities[mat_num])):
-                   if abs(den - densities[mat_num][j])/den < 1E-4:
+                   if abs(den - densities[mat_num][j]/den) < 1E-4:
                        same_bool = True
 
                if same_bool is False:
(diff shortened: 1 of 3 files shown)


pyne/mcnp.py
```diff
@@ -1344,7 +1344,7 @@ def mats_from_inp(inp):
            else:
                same_bool = False
                for j in range(0, len(densities[mat_num])):
-                   if abs(den - densities[mat_num][j])/den < 1E-4:
+                   if abs(den - densities[mat_num][j]/den) < 1E-4:
```
scopatz: The new one doesn't seem right from a unit analysis perspective.

scopatz: Sure that this didn't mean:

``` python
abs((den - densities[mat_num][j])/den)
```

elliottbiondo: That's exactly what I meant. My mistake. Fixed.

scopatz:
Thanks @elliottbiondo

PR #754, opened 2015-09-01
Updating files for MOAB 4.9

elliottbiondo:
This PR updates `source_sampling` to allow it to be used with new versions of MOAB. These are actually the only changes required for the whole PyNE code base. The changes are also fully backwards compatible: all unit tests now pass with both MOAB 4.9.2 and MOAB 4.6.2. Thanks to @makeclean for the help with this.


scopatz:
Thanks @elliottbiondo! Surprised that it was so few changes.

PR #755, opened 2015-09-03
Fixed atom density mistake

elliottbiondo:
In MCNP, if the density of a cell is specified by a number >0, it is interpreted as an atom density in units of 1E24 atoms/cm^3, (MCNP Manual Vol II, page 3.9). The function `mats_from_mcnp` neglected this factor of 1E24. This PR fixes this and updates the unit tests accordingly.


scopatz:
Nice catch! Thanks @elliottbiondo

PR #757, opened 2015-09-03
Fixing energy ordering

elliottbiondo:
In PARTISN, energy-dependent quantities are ordered from high-energy to low-energy. I could not find explicit documentation of this, but I was about to confirm this by experiment. I ran transport problems with source intensity only in the last energy group (i.e. the lowest energy group in PARTISN), and confirmed that fluxes in all other groups were zero. I also did the reverse --- only flux in the first group --- and confirmed that particles were indeed downscattering in the other groups.

In PyNE we have been using the opposite convention --- low-energy to high-energy --- consistent with MCNP. The PR fixes PARTISN functionality to be consistent with the rest of PyNE. For unit tests, I was able to simply reverse the order of input spectra, minimizing what needed to be changed.


scopatz:
Thanks @elliottbiondo!

PR #758, opened 2015-09-05
write_partisn_input() improvements

elliottbiondo:
This PR consists of a collection of tweaks to the  `write_partisn_input` function. Here they are listed roughly in order of decreasing importance:
1. Additional keyword arguments:
   - `cards` - A dictionary that allows for the specification of any arbitrary additional cards (both the keyword and value e.g. isn = 6) that should appear in any of the blocks of the the product input file. Note that any cards included in `cards` will not appear as "suggested" commented-out cards.
   - `dg` - the output of `discretize_geom()`. This is useful for workflows where discretize_geom has already been run once (e.g. using `partisn.isotropic_vol_source`) because this step is expensive.
   -  `fine_per_course` - the number of fine mesh intervals per course mesh interval. Previously this was essentially hardcorded as 2. I made the default value 1 because I think that is the most intuitive. Mesh generation should be a seperate option; a mesh is required to read RTFLUX output anyhow, which is on a fine-mesh basis.
2. The "suggested" cards that are printed (commented out), now correspond to the cards necessary for a minimum working PARTISN input file. I verified this by running PARTISN.
3. The default source is printed only if it not specified in `cards`. Like before it is uniform in space/energy and isotropic in direction. It does not include `sourcx`, `sourcy` or `sourcz` as this is unnecessary for this particular purpose. Note that source generation capabilities are present in `partisn.isotropic_vol_source()` and `partisn.mesh_to_isotropic_source()`.
4. "/rho" is stripped out of material names. "/rho" appears in names via the UWUW workflow. PARTISN crashes if "/" appears in material names.
5. Default HDF5 paths are now "/material" and "/nucid", which is where uwuw_preproc writes them. In fact these pathes were already being used in the tests.
6. For the `zones` card, 20 entries per line were being printed. I found that this caused entries to go past PARTISN's 80 characters per line limit, so I changed it to 10. This should really be fixed using `wrap(s, 80)`, but I figured this PR is large enough already.
7. Three output files generated from the unit tests were added erroneously. I deleted them and modified the unit tests to delete them automatically.
8. For testing, I changed the existing tests to reflect the new defaults. I also created a new test that exercises most of the features that I added.

With these changes I was able generate valid PARTISN input for a neutron transport problem with the Frascati neutron generator geometry. See PARTISN RTFLUX output below. 

@kkiesling has expressed that she would like to review this PR. She has Ph.D. qualifying exams this week, so I'll be patient! This may be of interest to @makeclean as well.

fng_partisn


scopatz:
Cool :) Glad that @kkiesling wants to review

kkiesling:
Ah yes! This is a lot to review. I will look at this after quals!

elliottbiondo:
I added one last commit: I no longer strip the density out of material names, in order to keep them unique. Thanks @makeclean for pointing that out.

pyne/partisn.py
```diff
@@ -822,17 +759,19 @@ def strip_mat_name(mat_name):
...
+    # Replace "/rho" with an underscore
+    mat_name = string.replace(mat_name, "/rho:", "_")
     
-    # Remove 'mat:'
-    tmp1 = mat_name.split(':')[1]
-    
-    # Remove other special characters
-    special_char = [':', ',', ' ', ';', "'", '"']
+    # Remove any other special characters
+    special_char = [':', ',', ' ', ';', "'", '"', '/']
```
makeclean: I think you should do something more robust like, remove all non alphanumeric characters, for example soon we will add the '@' operator, which I expect would break this. Something like http://stackoverflow.com/a/1276804

pyne/partisn.py
```diff
@@ -822,17 +759,19 @@ def strip_mat_name(mat_name):
...
     for char in special_char:
-        tmp2 = tmp1.split(char)
-        tmp1 = ''.join(tmp2)
+        mat_name = string.replace(mat_name, char, "")
     
-    return tmp1
+    return mat_name
```
makeclean: Are there any strict limits on the size of the string for name? I remember reading 24 characters was a limit for the material name, but I couldn't point you to a page in the manual.

elliottbiondo: Good point. By experiment, it looks like it's 8! I guess the FLUKA name really is the best choice here.

elliottbiondo:
@makeclean I just modified things so that unique FLUKA names are used.

pyne/partisn.py
```diff
@@ -101,12 +112,16 @@ def write_partisn_input(mesh, hdf5, ngroup, pn, **kwargs):
...
-    Returns
-    -------
-    PARTISN Input file named by 'input_file' above or the default name.
-        Note: read comments generated in file. Not all variables will be 
-        assigned that are necessary.
+    dg : record array, optional, default = None
+        The output of pyne.dagmc.discretize_geom(). Use this input option if 
+        discretize_geom() has already been run, to avoid duplicating this
+        expensive step.
+    find_per_coarse : int, optional, default = 1
```
kkiesling: @makeclean Didn't we find when we were originally making input that the # of fine mesh intervals per coarse had to be at least two. @elliottbiondo does this work with just one fine mesh per coarse?

Also should be fine_per_coarse not find_per_coarse.

pyne/partisn.py
```diff
@@ -117,101 +132,65 @@ def write_partisn_input(mesh, hdf5, ngroup, pn, **kwargs):
...
     block01['niso'] = len(xs_names)
     block03['names'] = xs_names
 
-    block01['igeom'], bounds, nmq = _get_coord_sys(mesh, pn)
+    block01['igeom'], bounds = _get_coord_sys(mesh)
     block01['ngroup'] = ngroup
     block01['mt'] = len(mat_lib)
     
-    block02['zones'], block04['assign'] = _get_zones(mesh, hdf5, bounds, num_rays, grid)
+    block02['zones'], block04['assign'] = _get_zones(mesh, hdf5, bounds, num_rays, grid, dg, unique_names)
```
kkiesling, line 169: What if dg is not supplied as an input?

kkiesling, line 169: Never mind, I see how it is handled in _get_zones

pyne/partisn.py
elliottbiondo: I found that 1 fine per course works fine, as long as you have at least 7 fine intervals total (or something to that affect). The FNG image I generated above used a fine_per_coarse of 1.

Having anything other than a fine_per_coarse of 1 makes it difficult to read the RTFLUX file, because there is one entry per every fine mesh interval. I figured it couldn't hurt to make it an option anyway.

pyne/partisn.py
kkiesling: I am not sure where 8 came from. Pg 2-47 of the partisn manual states character limits for other things but not the MATLS specifically. Just that there can be at most 500 comps and each comp's name can be at most 24 characters.

pyne/partisn.py
```diff
@@ -315,27 +295,22 @@ def _get_coord_sys(mesh, pn):
...
     """
     
-    # load the geometry
-    from pyne import dagmc
-    dagmc.load(hdf5)
-    
-    # Descretize the geometry and get cell fractions
-    dg = dagmc.discretize_geom(mesh, num_rays=num_rays, grid=grid)
+    # Discretize the geometry and get cell fractions
+    if dg is None:
```
kkiesling, line 311: I am a fan of this addition here. Good thinking.

pyne/partisn.py
```diff
@@ -472,36 +454,19 @@ def _check_fine_mesh_total(block01):
         if key in ['it', 'jt', 'kt']:
             total += block01[key]
     
-    if total >= 7:
-        # no warning necessary
-        return False
-    else:
-        # warn the user
-        return True
+    if total < 7:
```
kkiesling, line 457: Just out of curiosity @elliottbiondo, did you actually run into this problem where if you had less than 7 total it wouldn't work? It seems like a very odd number but @makeclean said his wasn't working for < 7.

pyne/partisn.py
```diff
@@ -586,7 +561,7 @@ def _write_input(title, block01, block02, block03, block04, block05, name=None):
         for num in row:
             partisn += "{} ".format(num)
             count += 1
-            if count == 20:
+            if count == 10:
```
kkiesling, line 564: When I was writing this function I wasn't exactly sure how to limit line length so 20 was a guess that worked at the time. But you seem to know how to do this with wrap(s,80). It might be worth adding the wrap(s,80) right away with this PR too.

kkiesling:
I like what you did with this @elliottbiondo. I don't see any major problems with it.

pyne/partisn.py
elliottbiondo, line 564: The `wrap(s, 80)` is a pretty big job because it really should be done on the whole file. I think that should wait.

pyne/partisn.py
elliottbiondo, line 457: I recall getting error messages to that nature but I did not thoroughly investigate. In practice I don't think it will come up much; any real problem will be plenty more then 7 fine mesh intervals.

elliottbiondo:
Thanks for the review @kkiesling! I fixed the typo in the API doc string. I think this is ready to merge.

kkiesling:
Merged!

PR #759, opened 2015-09-07
dagmc.cell_materials()

elliottbiondo:
This PR adds a function that reads a DAGMC material-laden geometry and returns a mapping between cell numbers and PyNE `Material` objects. Tests were added for this new function and also a similar function, `cell_material_assignments`, which previously did not have a test.


elliottbiondo:
It looks like I just forgot some `skiptests`. Fixed. This should be good to go.

scopatz:
This looks good. Thanks @elliottbiondo!

PR #761, opened 2015-09-09
alara.record_to_geom() improvements

elliottbiondo:
This PR adds a few improvement to `alara.record_to_geom()`:
1. `Materials` that specify void regions are now supported, and this is compatible with what I added in PR #759. 
2. `Material.metadata['name']` is now the name that is printed to the ALARA input file. This makes more sense in the general case, and will also benefit us here at UW with some of our new workflows. 

It looks like the commits from PR #759 are in here too, so that should go in first.


pyne/alara.py
```diff
@@ -266,20 +266,21 @@ def record_to_geom(mesh, cell_fracs, cell_mats, geom_file, matlib_file,
...
-            if cell_mats[row['cell']].metadata['mat_number'] \
-                not in ve_mixture.keys():
-                ve_mixture[cell_mats[row['cell']].metadata['mat_number']] = \
-                    round(row['vol_frac'], sig_figs)
+            cell_mat = cell_mats[row['cell']]
+            name = cell_mat.metadata['name']
+            if _is_void(name):
+                name = "mat_void"
+            if name not in ve_mixture.keys():
+                ve_mixture[name] = round(row['vol_frac'], sig_figs)
```
scopatz: Not that this is blocking, per say, but should `np.round()` be used instead of the built-in `round()`

pyne/alara.py
```diff
@@ -296,19 +297,34 @@ def record_to_geom(mesh, cell_fracs, cell_mats, geom_file, matlib_file,
...
         f.write(matlib)
 
+def _is_void(name):
+    """Private function for determining if a material name specifies void.
+    """
+    if "vacuum" in name.lower() or "void" in name.lower() \
+        or "graveyard" in name.lower():
+        return True
+    else:
+       return False
```
scopatz: not that this is blocking either but this could be rewritten as just 

``` python
return "vacuum" in name.lower() or "void" in name.lower() or "graveyard" in name.lower()
```

Or the faster:

``` python
lname = name.lower()
return "vacuum" in lname or "void" in lname or "graveyard" in lname
```

pyne/dagmc.pyx
```diff
@@ -658,6 +659,48 @@ def cell_material_assignments(hdf5):
                         
     return mat_assigns
 
+def cell_materials(hdf5, **kwargs):
+    """Obtain a material object for each cell in a DAGMC material-laden
+    geometry, tagged in UWUW format, i.e. "mat:<name>/rho:<density>" or
```
scopatz: documentation on this format would be nice :) Do you have a link?

pyne/dagmc.pyx
```diff
@@ -658,6 +659,48 @@ def cell_material_assignments(hdf5):
...
+        The path in the heirarchy to the material data table in the HDF5 file.
+    nucpath, str, optional, default='/nucid'
+        The path in the heirarchy to the nuclide array in the HDF5 file.
+    
+    Returns:
+    --------
+    cell_mats : dict
+        Dictionary that maps cells numbers to PyNE Material objects. 
+    """
+
```
scopatz: excessive whitespace

scopatz:
Hi @elliottbiondo - this looks good, overall.  There are some CI failures.

Also I noticed that double quotes were used in many places where single quotes would do. The SG says to prefer single quotes in these cases: http://pyne.io/devsguide/style_guide.html#python-style-guide

elliottbiondo:
Hi @scopatz,  thanks for the comments. I have fixed everything you requested. Note that the comments you made on dagmc.pyx have been fixed in my other PR: #759 and merged with this branch.

scopatz:
Looks great! Thanks @elliottbiondo

PR #762, opened 2015-09-11
float -> int

elliottbiondo:
This is a minor change to the `source.F90` provided for use with the `source_sampling` module. The second argument to `chkcel()` should be an `int`. Thanks @makeclean for figuring this out.


scopatz:
Does @makeclean want to review and merge?

makeclean:
Sure, looks good to me.

PR #763, opened 2015-09-11
R2s frontend

elliottbiondo:
This PR creates a CLI for the in capabilities in the `r2s` module. I will make a user manual entry for this too, which I will cite from this script.

### DIFF
--- a/scripts/r2s.py
+++ b/scripts/r2s.py
@@ -168,7 +168,7 @@ def main():
     args, other = parser.parse_known_args()
     if args.command == 'setup':
         setup()
-    if args.command == 'step1':
+    elif args.command == 'step1':
         step1()
     elif args.command == 'step2':
         step2()
--- a/scripts/r2s.py
+++ b/scripts/r2s.py
@@ -1,0 +1,125 @@
+import argparse
+import ConfigParser
+
+from pyne.dagmc import cell_materials
+from pyne.r2s import irradiation_setup, photon_sampling_setup
+
+config_filename = 'config.ini'
+alara_params_filename = 'alara_params.txt'
+
+config = \
+"""[step1]
+# Path to MCNP MESHTAL file containing neutron fluxes
+meshtal: meshtal
+# Tally number within the meshtal file containing the fluxes for activation.
+tally_num: 4
+# Path to the DAGMC material-laden geometry.
+geom: geom.h5m
+reverse: True
+# Number of rays to fire down each mesh row in each direction to calculate
+# cell volume fractions
+num_rays: 10
+grid: False
+
+[step2]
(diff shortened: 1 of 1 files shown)


scopatz:
Should we wait to merge this until the user manual entry, then?

scripts/r2s.py
```diff
@@ -0,0 +1,178 @@
+#!/bin/env python
+import argparse
+import ConfigParser
```
scopatz, line 3: I didn't know anyone used configparse anymore...

scripts/r2s.py
```diff
@@ -0,0 +1,178 @@
...
+    subparsers = parser.add_subparsers(help=r2s_help, dest='command')
+
+    setup_parser = subparsers.add_parser('setup', help=setup_help)
+    step1_parser = subparsers.add_parser('step1', help=step1_help)
+    step2_parser = subparsers.add_parser('step2', help=step2_help)
+
+    args, other = parser.parse_known_args()
+    if args.command == 'setup':
+        setup()
+    if args.command == 'step1':
```
scopatz: Shouldn't this be `elif`?

scopatz:
I think it would be nice to have a command line option to specify the output file. like a `-o` or a `-f`.  Also, I think the the functions here should have some minimal docstrings.  Other than that it looks great.

scripts/r2s.py
elliottbiondo: Yep! Fixed.

elliottbiondo:
Sure, I can just do the user manual now.

In regards to the output files, I am not sure changing the names of files would be too useful. `step1` produces: `alara_geom`, `alara_matlib`, `alara_fluxin`, and the latter two all have to match what is specified in `alara_params`. `step_2` produces `source_1.h5m`, `source_2.h5m` ... `source_N.h5m` and these files all must be eventually changed to `source.h5m` for photon transport due to an MCNP limitation. Each usage of r2s.py is meant to be done in a different folder which I will explain in the user manual.

scopatz:
My concern is that hard coding paths like this prevents best practices such as naming the file after uuids or some other experiment run number. If all files have the same name but are supposed to be in uuid'd directories, this doesn't scale super well in my experience because invariably files get moved around and the provenance of the original file is lost. 

If something needs to be named source.h5m eventually, that should be done as a copy or softlink of the original uuid'd file.

I'm happy for the file names that you chose to be the defaults. They make sense. However, forcing people to use these names is a little problematic, very easy to fix as a developer, and very difficult to fix as a user. I hope this makes sense...

elliottbiondo:
Got it. I just added an option to rename the output.

scopatz:
Thanks @elliottbiondo

PR #765, opened 2015-09-14
Added script for expanding vector tags.

elliottbiondo:
This script reads a mesh (.h5m) file and, for each vector tag of length N, creates N scalar tags. This is useful for visualizing data on mesh with programs that do not support vector tags.

### DIFF
--- a/scripts/expand_tags.py
+++ b/scripts/expand_tags.py
@@ -1,4 +1,4 @@
-#!/bin/env python
+#!/usr/bin/env python
 
 import argparse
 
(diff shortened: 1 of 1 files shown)


scripts/expand_tags.py
```diff
@@ -0,0 +1,44 @@
+#!/bin/env python
```
scopatz: `/usr/bin/env`

elliottbiondo: Oops! fixed.

scopatz:
Thanks @elliottbiondo. Waiting on CI.

PR #750, opened 2015-08-16
PARTISN Isotropic volume sources

elliottbiondo:
This PR adds the capability of generating isotropic volumetric sources for use with PARTISN from arbitrary DAGMC geometry cells. An example is shown below, where two DAGMC geometry cells are designated as sources and used for particle transport. I will defer further explaination to the API documentation I have provided.

This PR also fixes a mistake within the Rtflux tests

iso_vol_src

PR #766, opened 2015-09-15
Fix write_partisn_input() iteration order

elliottbiondo:
Partisn uses what we have been calling "zyx" iteration order, where "x" changes fastest. Evidence of this is found in #750. The `write_partisn_input()` function erroneously writes zones in "zxy" (not a typo) order. This was previously not discovered because unit tests did not test the iteration order and I only recently ran a significantly asymmetric problem for these effects to be noticed.

This PR corrects the iteration order by doing a few matrix operations. @kkiesling is planning on refactoring this function and I imagine it will be greatly simplified by changing the `structured_ordering` attribute on `Mesh`. I have also added unit test. Below is a large, asymmetric problem that I was able to run now that the iteration order is correct.
right_order


elliottbiondo:
Image of unit test, with asymmetry in all 3 dimensions.
unit_test

scopatz:
Who do you want to review this?

elliottbiondo:
I think @kkiesling should review.

scopatz:
Great :) I am super busy with NEUPs

elliottbiondo:
Good luck!

scopatz:
Thanks!

kkiesling:
Looks exactly as we discussed earlier for a temporary fix until I can refactor. I'll merge when travis-ci finishes.

kkiesling:
Hmm it says that the test you wrote @elliottbiondo doesn't pass.

elliottbiondo:
Yep, I forgot to add the unit test geometry. Fixed.

PR #767, opened 2015-09-21
/usr/bin/env

elliottbiondo:

Issue #760, opened 2015-09-07
ahot will not build due to line truncation on gcc 5.2.0


crbates:
I get the following error when trying to build pyne on gcc 5.2.0

``` sh
/pyne/src/transport_spatial_methods/3d/ahotn_kernel_module.f90:283:132:

    a22 = 1.0d0 + 2.0d0*ez*(dzx+dzy) + 2.0d0*gamma1*( (ex*ez-36.0d0*alpha1)*dxz/ex + (ey*ez-36.0d0*beta1 )*dyz/ey ) + 3.0d0*gamma0*ez0 
                                                                                                                                    1
Error: Line truncated at (1) [-Werror=line-truncation]
/pyne/src/transport_spatial_methods/3d/ahotn_kernel_module.f90:284:132:

    a33 = 1.0d0 + 2.0d0*ey*(dyx+dyz) + 2.0d0*beta1 *( (ex*ey-36.0d0*alpha1)*dxy/ex + (ey*ez-36.0d0*gamma1)*dzy/ez ) + 3.0d0*beta0 *ey0 
                                                                                                                                    1
Error: Line truncated at (1) [-Werror=line-truncation]
/pyne/src/transport_spatial_methods/3d/ahotn_kernel_module.f90:285:132:

    a44 = 1.0d0 + 2.0d0*ex*(dxz+dxy) + 2.0d0*alpha1*( (ex*ey-36.0d0*beta1 )*dyx/ey + (ex*ez-36.0d0*gamma1)*dzx/ez ) + 3.0d0*alpha0*ex0
```

crbates:
This just requires 3 line continuations to be added to fix.

scopatz:
Weird.  Thanks for reporting @crbates!

rachelslaybaugh:
I think a PR that should be coming soon from @howland includes this fix...

Correction; this has already been resolved through PR #768 / #767

PR #768, opened 2015-09-23
Same PR as #767

elliottbiondo:
This is the same PR as #767, but I messed some stuff up by pushing to the wrong remote (I have fully resolved this issue).


elliottbiondo:
@scopatz I made the changes in #760 but now some AHOT unit tests fail. Unless I am misunderstanding how Fortran continue lines work I think I may have found some deeper issues...

scopatz:
Don't line continuations require a trailing backslash?

scopatz:
Maybe @crbates can weigh in.

crbates:
So I think this is just an integral tests != correctness tests issue. We 'luckily' had a situation where the fortran 90 cutoff of 132 characters left us with all valid variables names and statements. I think the best we can do is to update the tests to match the new outputs and open an issue to check ahot for correctness.

PR #769, opened 2015-09-27
Ahot Fixes

scopatz:
This fixes the AHOT test issues.  The problem was stemming from some really strange (to me) compiler semantics about when to treat an expression as a real versus when to treat is as a rank 0 array. This in turn would cause floating point evaluation issues.  You can reproduce this issue by:

``` diff
-   a22 = ... + 2.0d0*gamma1*( (ex*ez-36.0d0*alpha1)*dxz/ex + (ey*ez-36.0d0*beta1 )*dyz/ey ) + ...
-   a33 = ... + 2.0d0*beta1 *( (ex*ey-36.0d0*alpha1)*dxy/ex + (ey*ez-36.0d0*gamma1)*dzy/ez ) + ...
-   a44 = ... + 2.0d0*alpha1*( (ex*ey-36.0d0*beta1 )*dyx/ey + (ex*ez-36.0d0*gamma1)*dzx/ez ) + ...
+   a22 = ... + 2.0d0*gamma1*( (ex*ez-36.0d0*alpha1)*dxz/ex + (ey*ez-36.0d0*beta1 )*dyz/ey ) + ...
+   a33 = ... + 2.0d0*beta1*( (ex*ey-36.0d0*alpha1)*dxy/ex + (ey*ez-36.0d0*gamma1)*dzy/ez ) + ...
+   a44 = ... + 2.0d0*alpha1*( (ex*ey-36.0d0*beta1 )*dyx/ey + (ex*ez-36.0d0*gamma1)*dzx/ez ) + ...
```

Note that there is a space after the beta1s in the (-) diff, but not after the alpha1s or gamma1s.  If you remove even the first such space, in the (+) diff, you will get slightly different floating point values in the second or third decimal place. The test was set up to test for 4 decimal places.

Removing the spaces is more stable and seems to be what the author intended. The spaces seemed to be there just so the columns would line up. So this is what I have done, added line continuations (which also triggered this issue), and updated the tests.

Now as to why a fortran compiler would treat `x *( y )` differently than `x*( y )` is beyond me. This language is not made for mortals, clearly :).


scopatz:
Also this should close #767 and #768

elliottbiondo:
Great sleuthing! Waiting for CI.

crbates:
well that's weird. I guess it's a better alternative than something being wrong with the code in general assuming this is a localized problem.

elliottbiondo:
Merged. Thanks @scopatz!

PR #770, opened 2015-09-27
Clean up discretize_geom() output

elliottbiondo:
This PR eliminates structured array entries for cells that have zero volume fraction within mesh volume elements. Note that this is not a correctness issue, but rather it makes the output cleaner.


scopatz:
Would it be reasonable to ask for a test for this case?

elliottbiondo:
@scopatz I cannot consistently replicate this issue to point of making a good test. I know it comes up for production level problems (large geometries, complex surfaces, many rays fired). I have tried making some simple (version control sized) problem for testing with no success.

scopatz:
Ok. Better to have it than not, I guess. merging.

scopatz:
Thanks @elliottbiondo!

PR #771, opened 2015-09-27
structured_coords

elliottbiondo:
Previously, if instantiating a structured `Mesh` object from an .h5m file, the `structured_coords` attribute was not populated. This PR fixes this and adds a corresponding test.


scopatz:
Thanks @elliottbiondo

PR #773, opened 2015-10-07
source.F90 improvements

elliottbiondo:
1)  The second argument to `chkcel` has been changed from 2 to 0. When set to 2, MCNP assumes the cell is "complicated" no matter what, whereas 0 indicates that the cell may or may not be "complicated". Previously this source.F90 only worked with DAG-MCNP5 and this minor change should bring about native MCNP support.

2) The particle type (e.g. neutron, photon) was previously hardcorded and now it is user-specified via `idum`. Documentation has been updated.

@makeclean may want to have a look.


scopatz:
Looks good to me, but @makeclean should definitely look.

makeclean:
Super!

PR #780, opened 2015-11-17
hotfix for ENSDF alpha handling

### LINKED ISSUE #779: Issue when reading ensdf decay file
Hi, first time using PyNE so this might be really obvious or I'm re-reporting a known issue. 
I'm using the PyNE virtual box http://pyne.io/install/vb.html#vb (0.5.0-rc1). I want to use the
`ensdf.decay()` function but I get an error when `decay()` tries to override a tuple entry:

>   File "/home/pyne-user/.local/lib/python2.7/site-packages/pyne/ensdf.py", line 1010, in decays
>     alpha[0] = parent
> TypeError: 'tuple' object does not support item assignment

This shouldn't happen because line 1010 can only be executed once since only one nuclide (U234)
decays to TH230 via alpha decay. The file is directly downloaded from nndc. 

My problem is illustrated here e6452b518ae12c7df7e75df2308cc3da66df7764

scopatz:
This should fix #779, which was caused by an inconsistency in using tuples (for alphas) and lists (for everything else). Does this work for you @snschune?

### DIFF
--- a/pyne/ensdf.py
+++ b/pyne/ensdf.py
@@ -677,7 +677,7 @@ def _parse_decay_dataset(lines, decay_s):
                 aparent = parent2
             level = 0.0 if level is None else level
             adaughter = data.id_from_level(_to_id(daughter), level)
-            alphas.append((aparent, adaughter, dat[0], dat[2]))
+            alphas.append([aparent, adaughter, dat[0], dat[2]])
         ec_rec = _ec.match(line)
         if ec_rec is not None:
             dat = _parse_ec_record(ec_rec)


scopatz:
Yeah, I am pretty sure those errors are unrelated.  They need to be fixed though.

scopatz:
This now passes all tests!

rachelslaybaugh:
I get a bunch of issues about hdf5 versions and some test failures when I check this, but that seems to be my installation. None of the issues are associated with this change - so I'm going to merge.

scopatz:
Thanks @rachelslaybaugh - I am interested in what the other test problems are.  Maybe you could open another issue for that?

PR #781, opened 2015-11-18
Small changes in documentation

ohnemax:
Stumbled upon a small error in documentation of collapse_elements function - a necessary parameter was not mentioned. Added the parameter.

### DIFF
--- a/pyne/material.pyx
+++ b/pyne/material.pyx
@@ -686,14 +686,14 @@ cdef class _Material:
         return newmat
 
     def collapse_elements(self, nucset):
-        """collapse_elements(self)
-        Collapses the elements in the material, excluding the nucids in 
-	paramater set. This function returns a copy of the material.
+        """collapse_elements(self, nucset)
+        Collapses the elements in the material, excluding the nuclids in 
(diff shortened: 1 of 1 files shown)


pyne/material.pyx
```diff
@@ -686,14 +686,14 @@ cdef class _Material:
         return newmat
 
     def collapse_elements(self, nucset):
-        """collapse_elements(self)
-        Collapses the elements in the material, excluding the nucids in 
-	paramater set. This function returns a copy of the material.
+        """collapse_elements(self, nucset)
+        Collapses the elements in the material, excluding the nuclids in
```
scopatz: I am pretty sure that this was meant to be read as `nuc-ids` indicating that they should be in id-form and not any other format.

scopatz:
Thanks for putting this in @ohnemax

pyne/material.pyx
ohnemax: true...

scopatz:
Thanks for the change.  Please remember to post after you update so that we get an email :)

PR #782, opened 2015-11-18
removed failing pprints

scopatz:
This fixes some erroneously failing tests on Python 3, and hopefully Python 2 as well.

### DIFF
--- a/tests/test_stlcontainers.py
+++ b/tests/test_stlcontainers.py
@@ -656,8 +656,6 @@ def test_map_str_vector_double():
     uismap = isinstance([1.0, -65.5555, 1.0, -65.5555], Mapping) 
     m['Aha'] = [42.42, 18, 42.42, 18]
     m['Take'] = [1.0, -65.5555, 1.0, -65.5555]
-    import pprint
-    pprint.pprint(m)
     assert_equal(len(m), 2)
     if uismap:
         for key, value in m['Take'].items():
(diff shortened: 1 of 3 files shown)


scopatz:
it works!

tests/test_stlcontainers.py
```diff
@@ -656,8 +656,6 @@ def test_map_str_vector_double():
     uismap = isinstance([1.0, -65.5555, 1.0, -65.5555], Mapping) 
     m['Aha'] = [42.42, 18, 42.42, 18]
     m['Take'] = [1.0, -65.5555, 1.0, -65.5555]
-    import pprint
```
katyhuff, line 659: Do we lose anything useful by taking out these pprints?

scopatz, line 659: Well, they were causing the tests to fail on Python 3 due to newly unhashable instance methods.  So it is an evolution of the language that we have to adapt around.  I also argue that printing in tests is only useful as a debugging mechanism when the test is failing and it is unclear why.  A more explicit test is better than a print, anyway. Also most of the time what is printed here is hidden anyway.  So I don't think we loose much of value and we gain a passing test suite. These tests are pretty thorough anyway.

katyhuff, line 659: (I understood why you deleted them... just wanted to make sure they weren't going to be missed.)

Sounds good to me. Merging!

scopatz:
Thanks!

PR #775, opened 2015-10-26
fix numpy version

scopatz:
As reported on the mailing list.

Issue #783, opened 2015-11-19
After merging #775, Installation fails with numpy 1.8.2


ohnemax:
I just tried to install a freshly pulled instance, however it failed immediately with the arbitrary error `"ValueError: numpy version too low! 1.8.2 (have) < 1.8.0 (min)"`

(1.8.2. is Debian Jessie package)

It is unclear to me, if the failing is intented because 1.8.2 is not supported, or if it is just an issue that came up after #775.

scopatz:
This is probably a post #755 issue.

ohnemax:
I came up with a possible solution, please check.

ohnemax:
Should be done with #784

PR #784, opened 2015-11-19
Fix to numpy version fix #775

ohnemax:
#775 introduced a problem while installing with numpy 1.8.2 (see my issue #783, created before thinking it through).

This should fix it.

However: I don't know which numpy version is intended to be used and if this fix keeps that intention.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -116,7 +116,7 @@ def cleanpypath(path):
 
 
 def assert_np_version():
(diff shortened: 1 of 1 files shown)


scopatz:
This looks good to me.  Waiting for CI

scopatz:
Thanks!

PR #788, opened 2015-11-30
clarify linux source install instructions

kellyrowland:
This adds a small amount of clarification to the instructions for installing PyNE on a Linux machine from source.

Previously, the directions read as if one should enter the "pyne" subdirectory and then execute the setup.py command, but this is not the case. There is also an added note that specifies that the "setup.py" command must be done in the top PyNE directory.


scopatz:
Thanks a ton @kellyrowland!

PR #733, opened 2015-05-02
Github Installation and Basic Use

crivera96:

PR #789, opened 2015-12-02
I think I addressed the issues in Github Installation and Basic Use #733

rachelslaybaugh:
but I didn't know how to modify that PR, so I made a new one. Sorry. If there are better directions let me know...


GithubSetup
```diff
@@ -0,0 +1 @@
+Subproject commit a73c2743725df2e1ec2f2327e330adecdb316e17
```
scopatz: Not sure what this file is for...

rachelslaybaugh: I don't either, it was in the original pr. I'll take it out (I'm guessing it was a random file that got added in)

scopatz:
Thanks!

PR #732, opened 2015-05-01
New_dev_guide

pcue:
The file wasn't displaying properly in the Developer's Guide tab under Contribute

PR #790, opened 2015-12-03
Replacement for New_dev_guide #732

rachelslaybaugh:
I believe this corrects the issues with PR #732


scopatz:
Great document, thanks!

PR #793, opened 2015-12-21
User std::abs

### LINKED ISSUE #792: abs used where std::abs seems more appropriate
`abs` implicitly casts its arguments to integers. This produces two compile warnings when building pyne.

dschwen:
Small compile warning fix.

Closes #792


scopatz:
Thanks a ton for putting this in @dschwen.  It seems that there is an unrelated bug in that is causing the CI to fail. Can you pull in my fix from my  `s32cf` branch, please? Thanks!

dschwen:
Will do in about an hour.

dschwen:
Hmm, yeah, no surprise here, as #794 failed as well :-)

scopatz:
Ok #794 now works, can you remerge please, @dschwen?

dschwen:
@scopatz this seems to make a test diff. Could you have a look if this is a sensible change after all?

scopatz:
These are probably legitimate changes in the underlying data file @dschwen.  Feel free to update the tests.  Thanks!

dschwen:
Finally!

scopatz:
Woohoo! Thanks for bearing with us.

PR #794, opened 2015-12-22
EMERGENCY: fix EAF data issue

scopatz:
The IAEA site is down and our backup mirror code has a bug.  This fixes both those issue.


scopatz:
Woohoo! This now passes.

PR #807, opened 2016-01-14
Added CRAM to alara.py

marissachanlatte:
I added CRAM to alara.py and added a new test, test_cram.py, to the test suite.

### DIFF
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -676,4 +676,3 @@ def cram(N, t, n_0, order):
         msg = 'Rational approximation of degree {0} is not supported.'.format(order)
         raise ValueError(msg)
 
-# !!!!
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -27,7 +27,7 @@ from pyne.mesh import Mesh, MeshError
 from pyne.material import Material, from_atom_frac
 from pyne import nucname
 from pyne.nucname import serpent, alara, znum, anum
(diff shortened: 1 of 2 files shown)


rachelslaybaugh:
Not sure what the build issue is here (first pass had travis ci build error). If I check out the branch I build and tests pass without problems. @scopatz or @elliottbiondo , any ideas?

marissachanlatte:
I've been trying to figure it out too. When I looked at the travis ci output I saw this: TypeError: Cannot cast ufunc add output from dtype('complex128') to dtype('float64') with casting rule 'same_kind'. I think it comes from using a deprecated version of NumPy, but I'm not sure how to fix it.

rachelslaybaugh:
Ah, search on ufunc here: https://docs.scipy.org/doc/numpy-dev/user/quickstart.html. I think the problem in python 2.7 is this. 

In python 3 xrange was renamed range (http://stackoverflow.com/questions/17192158/nameerror-global-name-xrange-is-not-defined-in-python-3)

Best way to deal with these is to google the errors directly. The internet usually knows.

marissachanlatte:
Thanks @rachelslaybaugh. I added those fixes, and there was one more thing I had to change for python 3, but now it looks like everything is passing.

rachelslaybaugh:
Looks good, my upated local build still works as well. merging

scopatz:
congrats!

PR #812, opened 2016-01-29
removed unessessary instrinsic function declarations

makeclean:
This PR removes the unnecessary intrinsic function declarations that were causing compile issues on older version of Gfortran.


scopatz:
Cool beans. Thanks @makeclean!

PR #820, opened 2016-02-19
Fix fortran compiler errors when using a source build of GCC

ljacobson64:
I was originally trying to install PyNE and all its dependencies, including GCC, from source. I was able to get it to work just fine if I used the default version of GCC present on the machine, but when using my own build of GCC, I got a few errors like this:

```
                   sigdig = INT(ALOG10(hf(l)))
                                      1
Error: Type of argument 'x' in call to 'alog10' at (1) should be REAL(8), not REAL(4)

                   sigdig = INT(ALOG10(hf(l)))
                               1
Error: Function 'alog10' at (1) is INTRINSIC but is not compatible with an intrinsic
```

and a few like this:

```
             dsum = dsum + DBLE(NORMFUNC(REAL(x)))
                                        1
Error: Type mismatch in argument ‘x’ at (1); passed REAL(8) to REAL(4)
```

and a few like this:

```
             CALL POLINT(h(j-km),s(j-km),k,0.,Ss,dss)
                                          1
Error: Type mismatch in argument ‘x’ at (1); passed REAL(8) to REAL(4)
```

This PR fixes the errors. I was able to build PyNE successfully with these changes.

### DIFF
--- a/src/ensdf_processing/ALPHAD/alphad.f
+++ b/src/ensdf_processing/ALPHAD/alphad.f
@@ -179,7 +179,6 @@
       INTEGER(KIND=4), INTRINSIC :: INDEX, INT, LEN, MIN0, NINT,        &       
      &                              LEN_TRIM
       INTEGER(KIND=4), EXTERNAL :: INDEXF, IVLSTR, TYPSTR
-      REAL(KIND=4), INTRINSIC :: ALOG10
       REAL(KIND=4), EXTERNAL :: VALSTR
       REAL(KIND=8), INTRINSIC :: DABS, DBLE, DMOD, DSQRT
       REAL(KIND=8), EXTERNAL :: DVALST
@@ -931,7 +930,7 @@
      &                  ' HF for alpha to', enxit(l), ' not saved.'
                   ihf = ihf - 1
                ELSE
-                  sigdig = INT(ALOG10(hf(l)))
+                  sigdig = INT(LOG10(hf(l)))
                   IF(sigdig.GT.0) THEN
                      sigdig = sigdig + 1
                   ELSE IF(sigdig.LE.0) THEN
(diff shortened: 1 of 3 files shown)


scopatz:
Maybe @makeclean or @howland can review this?

howland:
@ljacobson64 seems like these changes don't work with the f95 compiler travis-ci is using:

```
CMakeFiles/gtol.dir/ensdf_processing/GTOL/gtol.f.o: In function `trapzd.1593':
gtol.f:(.text+0x4fc5): undefined reference to `real_'
CMakeFiles/gtol.dir/ensdf_processing/GTOL/gtol.f.o: In function `qromb.1599':
gtol.f:(.text+0x5184): undefined reference to `real_'
CMakeFiles/gtol.dir/ensdf_processing/GTOL/gtol.f.o: In function `fndlmt.1607':
gtol.f:(.text+0x5532): undefined reference to `real_'
gtol.f:(.text+0x55bc): undefined reference to `real_'
collect2: ld returned 1 exit 
```

ljacobson64:
I'm looking into it. It loos like travis-ci is using gfortran 4.6.3, which isn't one of the versions with which I tested my code. I only tested it with 4.7+

update: the travis compile errors have nothing to do with the version of gfortran. I tried it with my own build of 4.6.3 and it worked fine.

ljacobson64:
The travis compile errors have nothing to do with the version of gfortran. I tried it with my own build of 4.6.3 and it worked fine.

ljacobson64:
Green circle!

scopatz:
Woohoo!  Does someone who actually uses the formula translator regularly want to give this a review.  Seems good to me.

makeclean:
I can on Monday, all services back to normal :)

makeclean:
Looks good!

PR #822, opened 2016-02-29
doc cleanup

elliottbiondo:
I had a stray comment that I forgot to fix when I updated this code a while ago. Fixed here.


scopatz:
thanks!

PR #823, opened 2016-03-01
Add a threshold capability for uniform biasing.

gonuke:
When using uniform sampling of source meshes, it is possible that some voxels with negligible source (e.g. 10^40) density will be sampled often.  One reason for a very low source density is a very small volume fraction of non-void material, in which case such a volume be selected undergo a lot of void rejection, and then ultimately be rejected for another voxel.  In some cases this can slow down a sampling process beyond reasonable use.

This adds a threshold that removes voxels that have a density the threshold from being sampled.  The user supplies the threshold in the rdum input card.

### DIFF
--- a/src/source_sampling.cpp
+++ b/src/source_sampling.cpp
@@ -6,16 +6,16 @@
 static pyne::Sampler* sampler = NULL;
 
 // Fortran API
-void pyne::sampling_setup_(int* mode) {
+void pyne::sampling_setup_(int* mode, double* threshold) {
   if (sampler == NULL) {
     std::string filename ("source.h5m");
     std::string src_tag_name ("source_density");
     std::string e_bounds_file ("e_bounds");
     std::vector<double> e_bounds = read_e_bounds(e_bounds_file);
     if (*mode == 0) {
-      sampler = new pyne::Sampler(filename, src_tag_name, e_bounds, false);
+      sampler = new pyne::Sampler(filename, src_tag_name, e_bounds, false, 0);
     } else if (*mode == 1) {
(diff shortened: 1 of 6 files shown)


elliottbiondo:
Needs tests and docs. The user manual should mention the rdum and ideally the theory manual would describe the modified method and maybe even modify or add a new sample calc.

src/source_sampling.cpp
```diff
@@ -221,14 +223,14 @@ std::vector<double> pyne::Sampler::read_bias_pdf(moab::Range ves,
...
-        if (q_in_group > 0) {
+        if (q_in_group > threshold) {
           for (j=0; j<num_e_groups; ++j) {
             bias_pdf[i*num_e_groups + j] =
                 volumes[i]*pdf[i*num_e_groups + j]/q_in_group;
           }
         } else {
           for (j=0; j<num_e_groups; ++j) {
-            bias_pdf[i*num_e_groups + j] = 0.0;
+            bias_pdf[i*num_e_groups + j] = pdf[i*num_e_groups + j];
```
pshriwise: @gonuke shouldn't this value be multiplied by the volume of the voxel as well?

gonuke: This is a choice... I decided not, but it shouldn't really matter anyway.  It would just result in a different biasing, although slightly higher because the volume is ~125.

makeclean: 8000 right? The source voxel size is 20x20x20 isnt it?

makeclean:
@gonuke could we organise tests for this feature?

gonuke:
Withdrawing this as I no longer think it is correct (and the correct implementation may be tricky)

PR #826, opened 2016-03-04
following changes to allow nuclear data to be retrieved from memory, …

makeclean:
…expand elements was broken, now fixed, needed refactor

### DIFF
--- a/src/material.cpp
+++ b/src/material.cpp
@@ -1159,47 +1159,57 @@ double pyne::Material::molecular_mass(double apm) {
   return atsperm / inverseA;
 }
 
-
+// Expands a PyNE material object, such that all elements are expanded into
+// their isotopic form
 pyne::Material pyne::Material::expand_elements() {
-  // Expands the natural elements of a material and returns a new material note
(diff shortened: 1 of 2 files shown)


scopatz:
CI fail :(

makeclean:
Yeah, but not due to me I think. The failure was wrt to a specific gfortran version not being found. Right? Or am I misreading the log?

scopatz:
Yeah, this seems to be a gfortran version error for some reason I do not understand.  It might be time to switch to CircleCI, where we would have more control over these kinds of things.

makeclean:
I think the problem is that one of the dependencies pulled in by Conda (I'm thinking scipy) may be out of date, I'll create an issue tomorrow and start some deeper digging.

gonuke:
@makeclean - this was a very long time ago now.... but do you know if this was fixed with the simpler change in #1117?  If not, we might want to revisit this soon.  We have a little more attention in PyNE so we can probably move this along more quickly.

makeclean:
I think this was fixed in #1117

PR #836, opened 2016-04-07
fix isnan compiler errors for newer clang,gcc

rwcarlsen:
Newer compiler versions (at least clang 3.7.1, gcc 5.3) change their stdlib
shenanigans a bit causing compiler errors like this:

```
    /home/r/src/cyclus/cyclus/src/pyne.cc:7975:8: error: use of undeclared identifier 'isnan'
      if (!isnan(k_conv)) {
```

So isnan calls need to be explicitly std:: prefixed - so the pre-existing
macro for apple+win systems was just scoped to include all systems (linux
too).

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -24,14 +24,12 @@
 #include <vector>
 #include <algorithm>
 
-#if defined __APPLE__ || defined __WIN_GNUC__
 #if (__GNUC__ >= 4)
   #include <cmath>
   #define isnan(x) std::isnan(x)
 #else
   #include <math.h>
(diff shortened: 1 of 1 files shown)


rwcarlsen:
For the future, it might be worth considering removal of these macros and just explicitly `std::` prefixing all uses of isnan and just requiring newer compilers.  Then we won't mess up and shadow replace isnan for everyone who includes `pyne.h`.

scopatz:
I am for getting rid of this macro if we can. Thanks for putting this in @rwcarlsen!

scopatz:
Thanks!

PR #837, opened 2016-04-14
putting in a temporary catch for missing ace files

FlanFlanagan:
PyNE seems to be trying to import cross section files into a data_source that may not exist on the local machine. This is a first cut at fixing this issue.


scopatz:
Looks good to me. Thanks @FlanFlanagan!

PR #841, opened 2016-04-26
Data source fix

FlanFlanagan:
A key error catch for missing cross sections within a cross section cache when building TAPE9.INP files. Probably needs some refinement.

### DIFF
--- a/pyne/origen22.py
+++ b/pyne/origen22.py
@@ -1509,19 +1509,29 @@ def xslibs(nucs=NUCS, xscache=None, nlb=(201, 202, 203), verbose=False):
         t9[nlb[2]][field] = {}
     # fill with data
     for nuc in nucs:
+        if verbose:
+            print('computing {0}'.format(nucname.name(nuc)))
+        key = nucname.zzaaam(nuc)
         try:
-            if verbose:
-                print('computing {0}'.format(nucname.name(nuc)))
-            key = nucname.zzaaam(nuc)
             if nuc in ACTIVATION_PRODUCT_NUCS:
                 _compute_xslib(nuc, key, t9[nlb[0]], xscache)
+        except KeyError:
+            if verbose:
+                print('Key Error with ', key, ' in a computing activation product cross sections')
+            continue 
+        try:
             if nuc in ACTINIDE_AND_DAUGHTER_NUCS:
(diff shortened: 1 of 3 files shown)


pyne/origen22.py
```diff
@@ -1509,18 +1507,22 @@ def xslibs(nucs=NUCS, xscache=None, nlb=(201, 202, 203), verbose=False):
...
-        if verbose:
-            print('computing {0}'.format(nucname.name(nuc)))
-        key = nucname.zzaaam(nuc)
-        if nuc in ACTIVATION_PRODUCT_NUCS:
-            _compute_xslib(nuc, key, t9[nlb[0]], xscache)
-        if nuc in ACTINIDE_AND_DAUGHTER_NUCS:
-            _compute_xslib(nuc, key, t9[nlb[1]], xscache)
-        if nuc in FISSION_PRODUCT_NUCS:
-            _compute_xslib(nuc, key, t9[nlb[2]], xscache)
+        try:
```
scopatz: Try blocks should be a couple of lines at most, and ideally should only be 1 line This is too all-encompassing.  I think you should have three try-except blocks, one in each of the conditionals.

pyne/xs/cache.py
```diff
@@ -126,11 +128,11 @@ def __setitem__(self, key, value):
             value = _valid_group_struct(value)
             cache_value = self._cache['E_g']
             if _same_arr_or_none(value, cache_value):
-                return 
+                return
             self.clear()
             self._cache['phi_g'] = None
             for ds in self.data_sources:
-                ds.dst_group_struct = value
+               ds.dst_group_struct = value
```
scopatz: Why is this a 3-space indent? Shouldn't it be 4-space?

scopatz:
Thanks for putting this in @FlanFlanagan!

scopatz:
Also, could you please add a test to ensure that the new KeyError issue happens?  This should be pretty easy to do since you can create a cache with no data sources and then assert that it raises a KeyError whenever you try to get anything out of it.

FlanFlanagan:
@scopatz ping. I fixed the test, I think.

scopatz:
Thanks @FlanFlanagan!

PR #845, opened 2016-05-09
accept elements as keys in text based materials

### LINKED ISSUE #844: Truncation of mass weights when generating materials from plaintext
Hey folks,

I've only just come across PyNE – what a fantastic thing! Thank you for all your hard work. I have a question about reading in material compositions. When I supply a dictionary, I can create a new Material instance easily enough. However, when I try to read the same data from a plain text file called _concrete_composition_: 

> H 0.56
> O 49.75
> Na 1.71
> Mg 0.26
> Al 4.69
> Si 31.47
> S 0.13
> K 1.92
> Ca 8.28
> Fe 1.24

With the following code: 

```
from pyne.material import Material

concrete_from_file = Material()
concrete_from_file.from_text('concrete_composition')
print(concrete_from_file)
```

It only partially reads in the data, returning this:

> Material:
> mass = 0.0
> density = -1.0
> atoms per molecule = -1.0
> Al =  4.6
> Ca =  8.2
> Fe =  1.
> H =  0.5
> K =  1.9
> Mg =  0.2
> Na =  1.7
> O =  49.7
> S =  0.1
> Si =  31.4

Which has truncated some of the values. How do I stop this from happening? I don't know if this is relevant, but when I run the `expand_elements()` method on the concrete partially read from file, it doesn't work. It does work for the case read in from a dictionary.

scopatz:
fixes #844. There is only one real line change here.


scopatz:
ping!

scopatz:
@FlanFlanagan can you take a look at this please and merge it in if you feel it is up to spec for you.

src/material.cpp
```diff
@@ -843,7 +843,8 @@ void pyne::Material::from_text(std::string filename) {
     } else if (keystr == "APerM") {
       f >> valstr;
       atoms_per_molecule = pyne::to_dbl(valstr);
-    } else if (pyne::nucname::isnuclide(keystr)) {
+    } else if (pyne::nucname::isnuclide(keystr) ||
+               pyne::nucname::iselement(keystr)) {
```
katyhuff, line 847: It seems like is the only line actually changed (other than whitespace)? I'm not sure what this is trying to solve, but it looks fine to me. If we actually want this bool to be true for either isnuclide or iselement, then I think we should pull it. If you can confirm this @scopatz , I'll pull it for you.

katyhuff, line 847: Correction. I've just read the relevant issue again and I agree this should fix it. Perhaps this is a good opportunity for  a quick test replicating #844 so that we don't reintroduce the bug?

scopatz, line 847: Thanks for keeping me honest, @katyhuff :). I have added a test and removed more whitespace.

FlanFlanagan:
Works on my machine and looks good to me. If you're okay with it @katyhuff I'll go ahead and merge.

katyhuff:
@FlanFlanagan  go for it!

PR #849, opened 2016-06-03
first round at self shielding

FlanFlanagan:
@scopatz updates to self shielding pyne side.


pyne/xs/data_source.py
```diff
@@ -362,7 +383,6 @@ def discretize(self, nuc, rx, temp=300.0, src_phi_g=None, dst_phi_g=None):
             Temperature [K] of material, defaults to 300.0.
         src_phi_g : array-like, optional
             IGNORED!!!  Included for API compatability
-        dst_phi_g : array-like, optional
```
scopatz: Why is this line removed?

pyne/xs/data_source.py
```diff
@@ -213,9 +214,29 @@ def discretize(self, nuc, rx, temp=300.0, src_phi_g=None, dst_phi_g=None):
...
+        Parameters
+        ----------
+        mat : array of floats.  
+            A map of the number densities of each of the nuclides of the material for 
+            which self-shielding is being calculated. 
+        data_source: pyne data_source
+            Contains the cross section information for the isotopes in the material 
+            that is experiencing the self shielding. 
+        """
+        weight = {}
```
scopatz: This should be `weights` since it is a collection, and therefore plural

pyne/xs/data_source.py
```diff
@@ -213,9 +214,29 @@ def discretize(self, nuc, rx, temp=300.0, src_phi_g=None, dst_phi_g=None):
...
+        for iso in mat:
```
scopatz: Use nuc instead of iso. See the style guide: http://pyne.io/devsguide/style_guide.html#terminology

pyne/xs/data_source.py
```diff
@@ -213,9 +214,29 @@ def discretize(self, nuc, rx, temp=300.0, src_phi_g=None, dst_phi_g=None):
...
+            weight[iso] = 0.
+            for iso_2 in mat:
```
scopatz: `iso_2` is bad.  `i` and `j` are more traditional loop variables for nuclides, or `n` and `m`

pyne/xs/data_source.py
```diff
@@ -213,9 +214,29 @@ def discretize(self, nuc, rx, temp=300.0, src_phi_g=None, dst_phi_g=None):
...
+                if iso_2 != iso:
+                    weight[iso] += mat[iso_2]*self.reaction(iso_2, 1,temp)
+            weight[iso] = 1/(weight[iso]/mat[iso] + self.reaction(iso, 1,temp))
```
scopatz: Please use `1.0` here, since it a float

pyne/xs/data_source.py
```diff
@@ -213,9 +214,29 @@ def discretize(self, nuc, rx, temp=300.0, src_phi_g=None, dst_phi_g=None):
...
+        data_source: pyne data_source
+            Contains the cross section information for the isotopes in the material 
+            that is experiencing the self shielding. 
+        """
+        weight = {}
+        for iso in mat:
+            weight[iso] = 0.
+            for iso_2 in mat:
+                if iso_2 != iso:
+                    weight[iso] += mat[iso_2]*self.reaction(iso_2, 1,temp)
```
scopatz: I think it would be easier to have another top-level `mass_rx` loop that stores the results of this expr for all nuclides, rather than recomputing it each time.

pyne/xs/models.pyx
```diff
@@ -160,8 +160,31 @@ def phi_g(E_g, E_n, phi_n):
...
+    Returns
+    -------
+    num_den : dictionary
+        dictionary of isotope to number density
+    """
+    num_den_mat = 6.0221409E23 * density / a_mass
+    num_den = {}
+    for iso in mat:
+        num_den[iso] = mat[iso] * num_den_mat[0]
+    return num_den
```
scopatz: This whole function seems like you are reinventing the whell: http://pyne.io/pyapi/material.html#pyne.material.Material.number_density

pyne/xs/models.pyx
```diff
@@ -160,8 +160,31 @@ def phi_g(E_g, E_n, phi_n):
...
+
+def group_collapse(sigma_n, phi_n, phi_g=None, partial_energies=None, E_g=None, E_n=None, wgts=None):
```
scopatz: PEP8, this line looks too long.  Also, please spell out the name `weights` to be more explicit

pyne/xs/models.pyx
```diff
@@ -201,16 +224,25 @@ def group_collapse(sigma_n, phi_n, phi_g=None, partial_energies=None, E_g=None,
         pem = partial_energies
     elif (phi_g is None) and (partial_energies is not None):
         pem = partial_energies
-        phi_g = np.dot(pem, phi_n)
+        if wgts == None:
```
scopatz: use `weights is None`, not `weights == None`.  It is at least 100x faster :).

pyne/xs/models.pyx
scopatz: Also please add weights to the docstring.

pyne/xs/models.pyx
```diff
@@ -201,16 +224,25 @@ def group_collapse(sigma_n, phi_n, phi_g=None, partial_energies=None, E_g=None,
...
+           phi_g = np.dot(pem, phi_n)
+        else:
+           phi_g = np.dot(pem, phi_n * wgts[::-1])
     elif (E_g is not None) and (E_n is not None):
         pem =  partial_energy_matrix(E_g, E_n)
-        phi_g = np.dot(pem, phi_n)
+        if wgts == None:
+           phi_g = np.dot(pem, phi_n)
+        else:
+           phi_g = np.dot(pem, phi_n * wgts[::-1])
```
scopatz: the `weights[::-1]` thing seems weird to me.  Shouldn't we just enforce / assume that the weights have the same tonicity as `phi_n`, not enforce that it has the opposite?

scopatz:
Thanks for putting this in @FlanFlanagan! Please ping when you think that it is ready for another look.

pyne/xs/data_source.py
```diff
@@ -213,9 +214,30 @@ def discretize(self, nuc, rx, temp=300.0, src_phi_g=None, dst_phi_g=None):
...
+            that is experiencing the self shielding. 
+        """
+        weights = {}
+        for i in num_dens:
+            weights[i] = 0.0
+            for j in num_dens:
+                if j != i:
+                    weights[i] += num_dens[j]*self.reaction(j, 'total', temp)
+            weights[i] = 1.0/(weights[i]/num_dens[i] + self.reaction(i, 'total', temp))
+        print(weights)
```
scopatz: Extra print statement

makeclean:
@FlanFlanagan could you rebase against develop and push again?

PR #850, opened 2016-06-08
Pointwise interface for OpenMC Data Source

scopatz:
This adds a public interface for pulling out the point wise cross sections from the openmc data source.  This is is a relatively simple rearrangement of another function, and is thus, already tested.  @FlanFlanagan interested in looking at this?


FlanFlanagan:
Looks good. Merging

PR #852, opened 2016-06-15
Added c++ such that elemental atomic masses are generated

makeclean:
Adds the capability for when using the atomic_data without loading it from file, the elemental atomic masses are calculated at map instanciation.


makeclean:
Could anyone review? @katyhuff @gonuke ?

src/_atomic_data.cpp
```diff
@@ -17,6 +18,26 @@ void pyne::_load_atomic_mass_map_memory() {
   } else { 
     _insert_abund_map();
   }
+
+  // calculate the atomic_masses of the elements
+  std::map<int,double> :: iterator it;
+
+  for ( int i = 0 ; i < 92 ; i++ ) {
```
katyhuff: From a style guide perspective, it seems like there is too much whitespace in this line. (Maybe drop the spaces just inside the parentheses?) https://google.github.io/styleguide/cppguide.html#Horizontal_Whitespace .

Also, this loop would be a little more clear (and actually would avoid declaring `int atomic_number = i + 1`) if the for loop explicitly looped over atomic number. 

So, this is just preference for brevity, but I would suggest something along the lines of:

```
for (int z=1; z<=92; z++){
     double element_atomic_weight = 0.0;
     ...
           if(pyne::nucname::znum(it->first) == z ) {
     ...

```

tests/test_data_nofile.py
```diff
@@ -45,7 +45,14 @@ def test_natural_abund_excited_state():
     nabund = data.natural_abund(excited)
     assert_equal(nabund, data.natural_abund_map.get(excited))
 
-
+def test_elements():
+    # set the datapath to nonsense so we call, the cpp data version
```
katyhuff, line 49: I don't think you meant to leave a comma in this comment... 

```
    # set the datapath to nonsense so we call, the cpp data version
```

Should be:

```
    # set the datapath to nonsense so we call the cpp data version
```

tests/test_data_nofile.py
```diff
@@ -45,7 +45,14 @@ def test_natural_abund_excited_state():
...
+    pyne_conf.NUC_DATA_PATH = b'bobbobhonkeytonk'
+    # initialize natural_abund_map
+    # test a series of elements
+    assert_equal(data.atomic_mass('H'),1.0079407540557774)
+    assert_equal(data.atomic_mass('Li'),6.9400366029079965)
+    assert_equal(data.atomic_mass('U'),238.02891048471406)
```
katyhuff: An additional test here might be to see if you still get the response you expect when you ask for Pu or some other transuranic? Your new loop (I assume intentionally, but I haven't thought about it very hard) doesn't include those elements. . .

katyhuff:
@makeclean I have made some review comments.

src/_atomic_data.cpp
scopatz: Also, why does it stop at 92?  Is this where the database stops?

tests/test_data_nofile.py
scopatz: PEP8 single space after comma

scopatz:
Thanks for putting this in @makeclean! Sorry about the delay!

src/_atomic_data.cpp
makeclean: So the database are for elements only, and therefore stops at 92

tests/test_data_nofile.py
makeclean: Just a note so you aren't surprised, but when asking for the atomic_mass of elements only valid results are returned for naturally occuring elements, for z>92 you get 0.0. This is the same if you use a nuc_data.h5 file.

makeclean:
@katyhuff this ok now?

katyhuff:
merged! thanks for the reminder @makeclean !

PR #854, opened 2016-06-30
added cross section scaler functionality

FlanFlanagan:
@scopatz


pyne/xs/cache.py
```diff
@@ -70,6 +70,7 @@ def __init__(self, group_struct=None,
                 self.data_sources.append(ds)
         self._cache['E_g'] = _valid_group_struct(group_struct)
         self._cache['phi_g'] = None
+        self._xs_scalers = {}
```
scopatz: Users should be able to pass in a value via the constructor. Namely, `scalars=None` and then this line will be:

```
self._scalars = {} if scalars is None else scalars
```

pyne/xs/cache.py
```diff
@@ -112,7 +113,8 @@ def __getitem__(self, key):
                 else:
                     raise KeyError
         # Return the value requested
-        return self._cache[key]
+        scaler = self._xs_scalers.get(kw['nuc'], 1)
```
scopatz: I would push this logic farther up so that the scaled value is stored in the cache, rather than recomputing it each time.  Also, I would make it so the default is None, and then the value is only scaled if not None.  That prevents extra floating point ops that aren't needed.

scopatz:
Thanks for putting this in @FlanFlanagan! Just a couple of small comments.

makeclean:
@FlanFlanagan any further progress to this PR?

FlanFlanagan:
I've updated this but due to an oversight I coupled some self shielding updates into this. Might be best to remove this PR and #849 and create a new PR.

PR #858, opened 2016-09-06
fix small typos in theory pages

kellyrowland:


makeclean:
Looks great, thanks @kellyrowland

PR #860, opened 2016-09-16
Dagmc singleton support

ljacobson64:
This PR removes support for the now defunct DAGMC singleton. This change is needed since we now only support the DAGMC constructor. Due to the large changes in the DAGMC and a number of non-backwards compatible changes in MOAB, only MOAB versions 4.9.2 or higher are supported.


ljacobson64:
@makeclean  Looks like this will require an update to Docker to get the new version of MOAB

ljacobson64:
Closing this as things have changed again in the moab 5.0 release

PR #861, opened 2016-09-26
Allow multiple occurences of a single nucID in a text representation

gonuke:
If the definition for a material is made by combining other materials (e.g. homogenization), it may be easier for a user if some nucIDs appear more than once.  The previous `from_text` overwrote with the last entry for any nucID.  This adds the fractions of each entry together.

This uncovered another bug in which the last entry was always used twice because the EOF didn't trigger at the right time.


makeclean:
Perfect, thanks @gonuke

PR #862, opened 2016-09-30
Conda build fixes

scopatz:
Here is a whole slew of updates to pyne to make it build, install, and test with contemporary dependencies.  @FlanFlanagan, maybe you could take a look.


scopatz:
Ok @FlanFlanagan - it works now

src/_atomic_data.cpp
```diff
@@ -18,25 +17,6 @@ void pyne::_load_atomic_mass_map_memory() {
   } else { 
     _insert_abund_map();
   }
-
-  // calculate the atomic_masses of the elements
-  std::map<int,double> :: iterator it;
-
```
FlanFlanagan, line 24: Why was this section removed?

FlanFlanagan:
Other than the one comment this looks good.

src/_atomic_data.cpp
scopatz: I think because it was replaced with the `pyne::_insert_atomic_mass_map()` function below, which gets called above in this function.  There is no longer any reason to read in this data from disk.

scopatz: Good question, though!

scopatz:
Yeah, I think that is legit, though.

src/_atomic_data.cpp
makeclean: @scopatz I think that needs to stay, that part calculates the atomic masses of the elements, the atomic_mass_map is just the list of the atomic masses of each nuclide

scopatz: Oh yeah, you are right. I didn't actually remove this. It was picked up from elsewhere. Not sure who removed it.

makeclean:
@scopatz @FlanFlanagan please back out the last PR, there are some things that need to be put back in :)

scopatz:
Well, the whole thing doesn't need to be reverted, just that piece.

makeclean:
Sure :)

scopatz:
Fix in #863.  Sorry!

PR #863, opened 2016-09-30
put back in element calc

scopatz:
@makeclean @FlanFlanagan - how does this look?

I think what caused this was leaving the keyboard mid-bisect. All apologies.


FlanFlanagan:
This looks better. I'll wait til CI passes.

scopatz:
Thanks!

PR #864, opened 2016-10-01
replaced FileNotFoundError with OSError for Python 2 compatability

scopatz:
@FlanFlanagan, would you mind taking a look at this please? It is needed for conda-forge


scopatz:
Looks like it passes!

FlanFlanagan:
Won't this allow all OSErrors? Seems pretty blanketing even though you're only trying to read the ENDF file.

scopatz:
Yes, but this is what Python 2 actually raises, so it is this or the tests can fail, sadly

PR #872, opened 2016-11-04
Added fortran writing/reading capability

ngranda:
This PR: 
Adds a fortran writer and a fortran reader along with the necessary scripts to use this capability. These files live under the fortranformat folder. 
The license for use has been added under the pyne license.text file


gonuke:
I think it would be good to add a really simple test - basically to show that it can be called from this installation.  Perhaps just use the writer to write a few things to a binary file and then use the reader to read them back and check that it does the round trip successfully.  Nothing too complicated.

gonuke:
@ngranda This is failing because it can't find your module in CI.  Might check with others who know something about PyNE CI

ngranda:
I believe I fixed it. The module should be added to the build now.

gonuke:
Thanks @ngranda - looks good

PR #875, opened 2016-11-29
Fix bug in test script

### LINKED ISSUE #874: Tests have been silently failing since March
So our CI has actually been (potentially) failing since March, due to dd5c2333d74188a269be9334cdb8ea6a0af21430, the offending line

     if [ $x=="ensdf_processing.py" ]; then

should be 

     if [ $x == "ensdf_processing.py" ]; then

Note the additional spaces. What has actually been happening, is that we only actually ran ensdf_processing.py over and over again, for the number of tests in the test directory.

ljacobson64:
Should fix #874

### DIFF
--- a/tests/test_data_nofile.py
+++ b/tests/test_data_nofile.py
@@ -44,7 +44,7 @@ def test_natural_abund_excited_state():
     excited = gnd + 1
     data.natural_abund(gnd)
     # excited state should not be in the map yet
-    assert_equal(data.natural_abund_map.get(excited), 1)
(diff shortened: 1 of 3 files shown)


ljacobson64:
Looks like the CI is currently failing because it's trying to download things from LANL, but LANL's website is "currently performing scheduled maintenance"

ljacobson64:
Never mind, the actual problem is that it's trying to use a HTTPS url when it needs to be HTTP

makeclean:
Awesome, thanks lucas

PR #879, opened 2017-01-27
Doc build updates

ljacobson64:
This PR fixes a few things related to building the documentation.

- Fixes the doc repo URL
- Fixes the bug where documentation generation commits were all named as "Pushed root-level docs at" with no date
- Clears everything out of the pyne.github.com directory before copying stuff in to make sure no old unnecessary files are lying around; this means we also need to add the CNAME and .nojekyll files each time
- Removes an unnecessary file named "index.rst 2"
- Updates the current year to 2017 in the copyright info


scopatz:
Sorry for the delay @ljacobson64!

PR #881, opened 2017-02-10
Small correction to location of r2s source.F90 file.

pshriwise:
Small correction to the location of the `source.F90` file in the  MCNP5 source sampling install instructions. File is located in `pyne/share` rather than `pyne/src`.


ljacobson64:
Thanks @Pshriwise

PR #898, opened 2017-07-03
Serpent branching group constants parsing in pyne/serpent.py

gridley:
Hi all,

These commits add capability to serpent.py to parse its somewhat different group constant generation output. For information on the syntax for having serpent do this, please see here [here](http://serpent.vtt.fi/mediawiki/index.php/Automated_burnup_sequence). The braching must cover all combinations of variations created with the "branch" command, so the parsed data is given as a nested dictionary. This should facilitate keeping data organized. The entries in the dictionary are indexed by first burnup step, then universe ID, then all branch names, and finally the parameter name of interest. An example can be found in the docstring for parse_coe.

In addition, a function for printing the coefficient dictionaries was included. This should display all of the nested structure quite well. Subsequently nested data gets indented, and not all params are printed by default for brevity.

Lastly, an appropriate test and example output file was added to the tests directory, implemented in test_serpent.py.


tests/test_serpent.py
```diff
@@ -82,3 +82,28 @@ def test_parse_det2():
...
+
+    # test some vales. Not all, of course.
+    # check BU step 1, universe 2, fuel0+rod0 branch
+    # P0 scattering matrix for 4 group MSRE
+    assert_array_equal(coe[1]["2"]["fuel0"]["rod0"]["INF_S0"],
+            [3.19826E-01, 5.83480E-03, 6.21411E-09, 0.00000E+00,
+             0.00000E+00, 2.73182E-01, 5.47793E-03, 0.00000E+00, 
+             0.00000E+00, 0.00000E+00, 2.59886E-01, 1.05679E-02,
+             0.00000E+00, 0.00000E+00, 7.47597E-04, 2.73214E-01])
+    assert(coe[1]["2"]["fuel0"]["rod0"]["INF_KINF"]== 1.61559E+00)
```
scopatz: `assert` is not a function

scopatz:
Hi @gridley, thanks for putting this in!  I will make some comments in line.  Once you have addresses them, please let us / me know by commenting in the conversation here.

pyne/serpent.py
```diff
@@ -370,3 +370,164 @@ def parse_det(detfile, write_py=False):
...
+        coe[1]         ["2"]     ["fuel0"]["rod0"]["INF_KINF"]
+        #   ^ dep step  ^ uni ID  ^branches       ^ parameter name
+        
+        all parameters serpent gives are available by default
+
+    Parameters
+    ----------
+    coefile : str or file-like object
+        Path to *.coe file or a *.coe file handle.
+
```
scopatz: no blankline needed here

pyne/serpent.py
```diff
@@ -370,3 +370,164 @@ def parse_det(detfile, write_py=False):
...
+        make up the remaining keys. 
+
+        note: B1 is used for leakage-corrected group constants generated on an infinite
+        lattice. Don't use these values otherwise.
+
+    """
+    if isinstance(coefile, basestring):
+        f = open(coefile, 'r')
+    else:
+        f = coefile
```
scopatz: f is never closed explicitly here.  Also, it is generally inadvisable to leave a file handle open longer than it needs to be.  You should be using a context manager (with-statement) for this.  This will guarantee that the file is always appropriately closed.

pyne/serpent.py
```diff
@@ -370,3 +370,164 @@ def parse_det(detfile, write_py=False):
...
+
+    coe = {}
+    universes = [] # universes GCs were made in
+    layers = None
+
+    while 1:
```
scopatz: It is much more Pythonic to do `while True:`

pyne/serpent.py
```diff
@@ -370,3 +370,164 @@ def parse_det(detfile, write_py=False):
...
+
+        # this line describes branch and coe indices
+        try:
+            l = next(f).split()
```
scopatz: Why not just iterate through the lines?  `for l in f`, rather than `while` and a try except?

pyne/serpent.py
```diff
@@ -370,3 +370,164 @@ def parse_det(detfile, write_py=False):
...
+        except StopIteration:
+            break
+        _,numbranch,coeIndex,nTot,nUni = tuple([int(item) for item in l])
```
scopatz: There are many PEP8 style guide violations in this block of code.  Please use a linter of some sort (pyflakes, pylint, flake8, pep8) or I can go through and point out this issues individually. Thanks!

pyne/serpent.py
```diff
@@ -370,3 +370,164 @@ def parse_det(detfile, write_py=False):
...
+
+        # this line describes what branch you're on
+        l = next(f).split()
```
scopatz: Oh this is why you don't iterate through.  Ignore the above.

pyne/serpent.py
```diff
@@ -370,3 +370,164 @@ def parse_det(detfile, write_py=False):
...
+        # init list to hold branch names if it is still None
+        if layers == None:
```
scopatz: None is a singleton in Python, so `layers is None` is the faster and more appropriate comparison.

pyne/serpent.py
```diff
@@ -370,3 +370,164 @@ def parse_det(detfile, write_py=False):
...
+    """ Pretty prints a dictionary returned
+    by parse_coe. Use this to visualize its
+    nested structure.
+    
+    Parameters:
+    -----------
+    coedict: dict
+        particularly, one returned by parse_coe,
+        although it will work on any nested dict.
+    short:   bool
```
scopatz: numpydoc format requires spaces on both sides of the colon.

pyne/serpent.py
```diff
@@ -370,3 +370,164 @@ def parse_det(detfile, write_py=False):
...
+        found. Set to false for a full STDOUT.
+        
+    Returns:
+    --------
+    None
+    """
+    # blah blah, recursive functions aren't fast
+    # print header:
+    #if lvl==0:
+    #    print("BU i| uni\# | branches ...")
```
scopatz: Please remove extraneous comments

tests/test_serpent.py
```diff
@@ -82,3 +82,28 @@ def test_parse_det2():
...
+
+    # check that all branches are present
+    # in all universes
+    fuelBranches = ["fuel"+str(i) for i in range(15)]
```
scopatz: PEP8 violations in this comment as well.

pyne/serpent.py
gridley: fixed!

pyne/serpent.py
gridley: f is now closed before returning the dictionary. A with-statement was not used in order to concisely keep the option of passing either a file handle or string to read data from.

pyne/serpent.py
gridley: thanks, fixed!

pyne/serpent.py
gridley: The code is more concise this way. Using a for loop, some more if-elif-elif-... logic would be needed. This is because the data file includes a few lines describing the material, branch name, and universe, and then a fixed number of parameters. This fixed number of parameters is given by the nPara variable. Then next() can be used to read until the next branch description. In addition, using a try/except makes the code more robust, because you then must know exactly where the file ends, ensuring that the serpent data was given in the expected format.

gridley:
Oops, had a bit of a "learning experience" with the tests. I'll never force push ever again. Just a bit more to get these back in place...

pyne/serpent.py
scopatz: Ahh good, good

pyne/serpent.py
```diff
@@ -370,3 +379,137 @@ def parse_det(detfile, write_py=False):
...
+                    values.append(float(value[0]))
+                else:
+                    values.append([float(item) for item in value])
+
+            thisCoe = coe[BUi][uni]
+            for branch in theseBranches:
+                thisCoe = thisCoe[branch]
+            thisCoe = dict(zip(names, values))
+
+    f.close()
```
scopatz: f should not be closed here if the user passed in a file handle.  It should only be closed if the user passed in a file path and we opened the handle here.

pyne/serpent.py
```diff
@@ -370,3 +379,137 @@ def parse_det(detfile, write_py=False):
...
+
+    return coe
+
+
+def print_coe(coedict, lvl=0):
+    """ Pretty prints a dictionary returned
+    by parse_coe. Use this to visualize its
+    nested structure.
+
+    Parameters:
```
scopatz: numpydoc does not have a trailing colon in titles.

pyne/serpent.py
```diff
@@ -370,3 +379,137 @@ def parse_det(detfile, write_py=False):
...
+    -----------
+    coedict: dict
+        particularly, one returned by parse_coe,
+        although it will work on any nested dict."
+
+    Returns:
+    --------
+    None
```
scopatz: Returns None is not needed, since this is normal python behaviour

pyne/serpent.py
```diff
@@ -370,3 +379,137 @@ def parse_det(detfile, write_py=False):
...
+    """
+    if type(coedict) == dict:
```
scopatz: type checking like this is bad form in Python.  Replace this with

```py
from collections.abc import Mapping

if isinstance(coedict, Mapping):
```

pyne/serpent.py
gridley: thanks for pointing this out!

pyne/serpent.py
gridley: fixed this, and now with tests passing.

gridley:
OK, should be all fixed up now!

gridley:
Looks like test_cram.py is failing.

scopatz:
I bet the test is failing because it is being tested on Python 2.7.  The consensus at the PyNE meeting was to move to only supporting Python 3.6. Would you be willing to try to switch it over in the `.travis.yml` file?

gridley:
Now, it seems that something is wrong with dagmc when using python3.6 in travis.

scopatz:
Hmm dagmc seems to be running OK.  However, the tests seem to be running with Python 2.7 still. https://travis-ci.org/pyne/pyne/builds/250941577#L3246

scopatz:
These lines probably need to be changed to use `nosetests3`:

https://github.com/pyne/pyne/blob/develop/tests/travis-run-tests.sh#L8

https://github.com/pyne/pyne/blob/develop/tests/travis-run-tests.sh#L10

gridley:
I'm not sure how it's still running with 2.7. If you look in changed files, the .travis should be running 3.6 now. I'll definitely try those nose ones though.

makeclean:
Dont forget that the Docker image on which CI runs does not have python3 (I dont think). So adding 3.7 (or anything for that matter) as an additional test version of python does nothing, since this sets Travis' python version and not the Docker version, i.e. you're still running 2.7. I have an open PR #900 which updates some other things for Docker, I can add python3 if that would be of use?

gridley:
Hm. Seems that nosetests3 wasn't found.

makeclean:
@gridley do you see my above comment? Since the CI uses a Docker image it can only use whats on the image. Nosetests3 is not on the image, nor is Python 3. Is this something that is needed?

gridley:
@makeclean yes, I see it. I'm not familiar with docker though.

katyhuff:
Hi There. I'm curious why this was closed.

gridley:
@katyhuff since it seemed to be a pain to get this to work with the build and test system for PyNE, and the serpent-tools repo provided functionality under a more convenient package. Although, if any PyNE users expressed a desire for this, it'd definitely be worth taking another look at.

gonuke:
Even better would be to convince the serpent-tools people to integrate into PyNE....

PR #899, opened 2017-07-05
Update DAGMC & MOAB 

makeclean:
This PR adds DAGMC version 3 & MOAB v5 support. The Docker image needs to be updated though. This will be coming in another PR - which needs to come first.


scopatz:
Thanks @makeclean! Code looks good but, the build is failing on CI.

gonuke:
I think this was superseded by the PyMOB inclusion - please confirm and close if true.  Thanks @makeclean

PR #900, opened 2017-07-06
Update docker

makeclean:
This PR updates the CI build instructions, adds the dockerfile so anyone can rebuild the docker image. In order to test this someone needs to build the docker image, build pyne and run the tests. PR #899 depends upon the changes found in this PR. Once someone approves of these changes I can push the image to dockerhub.

### DIFF
--- a/docker/DockerFilePyneUbuntu1704Python2
+++ b/docker/DockerFilePyneUbuntu1704Python2
@@ -37,13 +37,17 @@ RUN \
     make check && \ 
     make install && \
     cd .. && \ 
-    rm -rf bld moab
+    rm -rf bld 
 RUN \
     echo "export LD_LIBRARY_PATH=/opt/moab/lib" >> /opt/env.sh && \
     echo 'export PATH=$PATH:/opt/moab/bin' >> /opt/env.sh
 RUN \
     pip install numpy && \
-    pip install cython
+    pip install cython && \
+    pip install tables && \
+    pip install nose && \
+    pip install scipy && \
+    pip install jinja2 
 RUN \
     cd /opt/ && \
     wget https://pypi.python.org/packages/7a/51/c5915ca128da8ab178c3f2a52eae9abcb70ac9b9f5e0d9dd917007730dc8/PyTAPS-1.4.tar.gz && \
--- a/docker/DockerFilePyneUbuntu1704Python2
+++ b/docker/DockerFilePyneUbuntu1704Python2
@@ -39,10 +39,12 @@ RUN \
     rm -rf bld moab
 RUN \
     echo "export LD_LIBRARY_PATH=/opt/moab/lib" >> /opt/env.sh && \
-    echo 'export PATH=$PATH:/opt/moab/bin' >> /opt/env.sh 
+    echo 'export PATH=$PATH:/opt/moab/bin' >> /opt/env.sh
+RUN \
+    pip install numpy && \
+    pip install cython
 RUN \
     cd /opt/ && \
-    pip install numpy && \
     wget https://pypi.python.org/packages/7a/51/c5915ca128da8ab178c3f2a52eae9abcb70ac9b9f5e0d9dd917007730dc8/PyTAPS-1.4.tar.gz && \
     tar -zxf PyTAPS-1.4.tar.gz && \
     mv PyTAPS-1.4 pytaps && \
--- a/docker/DockerFilePyneUbuntu1704Python2
+++ b/docker/DockerFilePyneUbuntu1704Python2
@@ -5,33 +5,38 @@
 #
 
 # pull the base image
-FROM ubuntu:14.04
+FROM ubuntu:17.04
 
 # install
 RUN \
-    apt-get -y update && \
-    apt-get -y install build-essential python-numpy python-scipy cython  && \
-    apt-get -y install python-nose git cmake vim emacs gfortran libblas-dev && \
-    apt-get -y install liblapack-dev libhdf5-dev gfortran python-tables && \
-    apt-get -y install python-matplotlib python-jinja2 autoconf libtool && \
-    apt-get -y install python-pip wget 
+    apt-get -y update && \ 
+    apt-get -y install gcc g++ make openssl
+RUN \
+    apt-get -y install cmake curl git htop man unzip vim wget emacs
 RUN \ 
+    apt-get -y install libhdf5-dev libhdf5-100 gfortran && \
+    apt-get -y install expat libtool autoconf && \
+    apt-get -y install patch texinfo && \
+    apt-get -y install dejagnu bison flex && \
+    apt-get -y install python && \
+    apt-get -y install subversion python-dev python-pip
+# install moab
+RUN \
+    echo "#!/bin/bash" > /opt/env.sh && \
     cd /opt && \
     mkdir moab && \
     git clone https://bitbucket.org/fathomteam/moab && \
     cd moab && \
-    git checkout Version5.0 && \
+    git checkout 4.9.2 && \
     autoreconf -fi && \
     mkdir bld && \
     cd bld && \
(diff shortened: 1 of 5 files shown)


gonuke:
I'm new to this docker CI stuff...  is there a reason we're not using U16.04?

docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+RUN \
+    apt-get -y update && \
+    apt-get -y install build-essential python-numpy python-scipy cython  && \
+    apt-get -y install python-nose git cmake vim emacs gfortran libblas-dev && \
+    apt-get -y install liblapack-dev libhdf5-dev gfortran python-tables && \
+    apt-get -y install python-matplotlib python-jinja2 autoconf libtool && \
+    apt-get -y install python-pip wget 
+RUN \ 
+    cd /opt && \
+    mkdir moab && \
```
gonuke: don't you need to `cd moab` here?

docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+    cd moab && \
+    git checkout Version5.0 && \
+    autoreconf -fi && \
+    mkdir bld && \
+    cd bld && \
+    ../configure --enable-optimize --enable-shared --disable-ahf --with-hdf5=/usr/lib/x86_64-linux-gnu/ --prefix=/opt/moab/ && \
+    make -j4 && \ 
+    make check && \ 
+    make install && \
+    cd .. && \
```
gonuke: don't you need to `cd ../..` here?

docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+    git checkout Version5.0 && \
+    autoreconf -fi && \
+    mkdir bld && \
+    cd bld && \
+    ../configure --enable-optimize --enable-shared --disable-ahf --with-hdf5=/usr/lib/x86_64-linux-gnu/ --prefix=/opt/moab/ && \
+    make -j4 && \ 
+    make check && \ 
+    make install && \
+    cd .. && \ 
+    rm -rf bld moab && \
```
gonuke: only delete `moab` here (assuming the directory changes I've indicated)

docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+    apt-get -y update && \
+    apt-get -y install build-essential python-numpy python-scipy cython  && \
+    apt-get -y install python-nose git cmake vim emacs gfortran libblas-dev && \
+    apt-get -y install liblapack-dev libhdf5-dev gfortran python-tables && \
+    apt-get -y install python-matplotlib python-jinja2 autoconf libtool && \
+    apt-get -y install python-pip wget 
+RUN \ 
+    cd /opt && \
+    mkdir moab && \
+    git clone https://bitbucket.org/fathomteam/moab && \
```
gonuke: perhaps if you clone this repo into a directory called something like `moab-src` it will reduce confusion about which directory you are in.

docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+    cd .. && \ 
+    rm -rf bld moab && \
+# this line patches the current moab version for pytaps comptability
+    sed -i '/IMESH_LIBS/i IMSQ_LIBS = \n MSQ_LIBS = ' lib/iMesh-Defs.inc 
+RUN \
+    echo "export LD_LIBRARY_PATH=/opt/moab/lib" >> /opt/env.sh && \
+    echo 'export PATH=$PATH:/opt/moab/bin' >> /opt/env.sh 
+RUN \
+    cd /opt/ && \
+    pip install numpy && \
```
gonuke: is this because you can't apt-get a new enough version of numpy?

gonuke (changes requested):
I'v PR'ed some changed into this branch that match the comments....
Mostly clean up of the MOAB install but also added a DAGMC install

It worked with PR #899

gonuke:
Is it true that these changes MUST be tested with PR #899?  It seems so to me...

katyhuff:
Regarding #898, can you update this to python 3?

makeclean:
I can, but that leaves me with other questions, should python 2.7 also exist alongside, and therefore we test both 2.7 and python 3 compatibility? Or should we only support python3? Which python3, 3.4 3.5? etc.

katyhuff:
I think the conclusion from the pyne summit was that we should drop python 2 support ( @scopatz , @gonuke  can confirm... I left the meeting early). 

If it's reasonable to have many versions of python supported, I'd say do 3.4, 3.5, 3.6, and onward.

makeclean:
OK, so I have a new docker image with all the python3.5 dependencies met with one exception, PyTAPS - our python interface to MOAB. Now, PyTAPS is as far as I understand orphaned with no further changes to be expected, we also now have a release of MOAB with the pyMOAB interface so future PyNE would probably conform to this interface instead, this is somewhat out of scope to what im doing right now. So as I see it I have 2 options.

1) Continue to try and get beyond the compiling issues I have with PyTAPS and python3, ive already changed the build scripts
2) Ignore testing with PyTAPS and have the CI not built with MOAB (& PyTAPS for meshes) on the understanding that when the new interface is complete someone adds it - PR's that change the mesh interface need therefore to be tested against a different image.

I would be happy to hear opinions

gonuke:
Don't we need to retain python 2.x support until we replace PyTAPS?  Take PR #901 - it specifically relies on PyTAPS

docker/DockerFilePyneUbuntu1704Python2
```diff
@@ -0,0 +1,64 @@
...
+    git checkout 4.9.1 && \
+    autoreconf -fi && \
+    mkdir bld && \
+    cd bld && \
+    ../configure --enable-optimize --enable-shared --enable-dagmc --with-hdf5=/usr/lib/x86_64-linux-gnu/hdf5/serial/ --prefix=/opt/moab/ && \
+    make -j4 && \ 
+    make check && \ 
+    make install && \
+    cd .. && \ 
+    rm -rf bld moab
```
gonuke: If this is attempting to delete the `moab` that is the repo, then I think you're in the wrong directory.  
You `cd ..` from `/opt/moab/moab/bld`

docker/DockerFilePyneUbuntu1704Python3
```diff
@@ -0,0 +1,57 @@
...
+    git checkout 4.9.1 && \
+    autoreconf -fi && \
+    mkdir bld && \
+    cd bld && \
+    ../configure --enable-optimize --enable-shared --enable-dagmc --with-hdf5=/usr/lib/x86_64-linux-gnu/hdf5/serial/ --prefix=/opt/moab/ && \
+    make -j4 && \ 
+    make check && \ 
+    make install && \
+    cd .. && \ 
+    rm -rf bld moab
```
gonuke: similar comment to above about success of removing `moab`

gonuke (commented):
Still looks like one of your `rm` commands will quietly fail.
Not really a blocking issues since it doesn't affect the success/completion of this.

gonuke (approved):

makeclean:
Ok so I think this is actually good, but Im seeing failures from CI, one of them is a compilation issue, which I can fix. But the 2 tests that consistently fail are the index errors from numpy. Could someone please build the current head of develop, ubuntu 16.04 or 17.04 and see if you get failures from nosetests for those two specific tests. Im suspicious that its the numpy/scipy version that I installed into the Docker image

scopatz:
Hi @makeclean - I am seeing the same issues on develop as I am on your branch and that you are seeing.  So your branch here does no harm.  It is safe to merge in my opinion.

makeclean:
I'll do one more push to fix the amalgamate build issue, and then address @gonuke's comments then it should be good to go.

gonuke:
I think this has been superseded by more recent changes.  Are there any features that still aren't included in our CI, @makeclean ?

gonuke:
Closing this as we have replaced everything here with equivalent capability.

PR #901, opened 2017-08-18
R2S fix

makeclean:
This fixes an issue discovered where the r2s routines were not being passed appropriate information from tags in meshes to irradiation setup routines, tet meshes were the only ones affected by this. Changes have also been made in the DAGMC repository such that unstructured meshes will produce the vector tag information expected by r2s.py

### DIFF
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -52,7 +52,9 @@ def mesh_to_fluxin(flux_mesh, flux_tag, fluxin="fluxin.out",
     tag_flux = flux_mesh.mesh.getTagHandle(flux_tag)
 
     # find number of e_groups
-    e_groups = flux_mesh.mesh.getTagHandle(flux_tag)[list(flux_mesh.mesh.iterate(iBase.Type.region,iMesh.Topology.all))[0]]
+    e_groups = flux_mesh.mesh.getTagHandle(flux_tag)[list(
+        flux_mesh.mesh.iterate(iBase.Type.region,
+                               iMesh.Topology.all))[0]]
     e_groups = np.atleast_1d(e_groups)
     num_e_groups = len(e_groups)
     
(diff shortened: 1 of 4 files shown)


gonuke:
This must have been lacking tests - do we have tests now?

makeclean:
Not yet, but soon :)

pyne/alara.py
```diff
@@ -52,12 +52,10 @@ def mesh_to_fluxin(flux_mesh, flux_tag, fluxin="fluxin.out",
     tag_flux = flux_mesh.mesh.getTagHandle(flux_tag)
 
     # find number of e_groups
-    e_groups = flux_mesh.mesh.getTagHandle(flux_tag)[list(
-        flux_mesh.mesh.iterate(iBase.Type.region,
-                               iMesh.Topology.all))[0]]
+    e_groups = flux_mesh.mesh.getTagHandle(flux_tag)[list(flux_mesh.mesh.iterate(iBase.Type.region,iMesh.Topology.all))[0]]
```
scopatz: This line is waay too long.

scripts/expand_tags.py
```diff
@@ -18,7 +18,11 @@ def main():
                               "be seperated by commas without spaces.\n"))
     args = parser.parse_args()
 
-    m = Mesh(structured=True, mesh=args.mesh_file)
+    try:
+        m = Mesh(structured=True, mesh=args.mesh_file)
+    except:
```
scopatz: Exceptions should only catch a the type of error that is expected (eg KeyError). Please qualify this. Thanks!

scripts/expand_tags.py
```diff
@@ -18,7 +18,11 @@ def main():
...
+        print ("Structured mesh not found, trying unstructured mesh")
```
scopatz: This should probably be a warning, or at least go to stderr, rather than being a print.

scopatz:
Thanks for the PR @makeclean! A couple of small style things from me.

scopatz:
Pinging this! Also, will this address the issues on #909?

makeclean:
Ah yes it might well do

makeclean:
So I think this it good to now, adds a new unit test and addresses PR comments.

scopatz:
Thanks @makeclean!

PR #903, opened 2017-08-28
Update data

makeclean:
This PR updates the data location of the decay.tar.gz to point to the storage location on pyne/data. Testing on any branch will fail utils this is merged in.


scopatz:
Thanks @makeclean!

PR #906, opened 2017-09-06
Build System Overhaul - Including RPATH Fix

scopatz:
This basically rewrites the whole build system and flow for pyne. I have hopefully made more readable in the process.  The motivating reason for this is to correctly set the RPATH on the installed files.  This prevents the need for users to set `$LD_LIBRARY_PATH` or similar.  The shared objects *just* work (as far as I could test). The build flow now looks like the somewhat manic:

* setup.py, which calls
* cmake, make, and make install, which calls
* setup_sub.py.

setup.py is used as the name for the standard python installer interface. It is really a wrapper script around CMake (like Cyclus' install.py).  `make install` then calls setup_sub.py (which mainly calls distutils setup() function).  After this, `make install` repairs the RPATH on all shared objects.  

(Other) features include:

* No need for users to set `$LD_LIBRARY_PATH`!
* library is now at `$PREFIX/lib/libpyne.so`
* headers are now at `$PREFIX/include/pyne`
* It seems that the annoying bug with `nuc_data_make` needing to be rerun after each install is gone! I think that this is because I... 
* removed setuptools optional install dependency.
* Amazingly, this fits in less LOC
* updated cmake required version to 3.0.0

Let me know how it goes!

PR #908, opened 2017-09-11
CRAM via Transmutagen

scopatz:
This is an initial, low-frills integration of the [transmutagen](http://www.ergs.sc.edu/transmutagen/) CRAM solver into pyne.  The basic implementation requires the user to bring their own transmutation matrix.  There are tools for building this up in cram.py, including a ready-to-use decay matrix.  Materials now have a new `cram()` method.  

This builds on #906, so that should probably be merged first.


scopatz:
This passes now

PR #913, opened 2017-09-24
Decaygen Significant Figures

scopatz:
This fixes a bug where decay gen was only printing out 8 significant figures (the default) for its constants. Thus results were only accurate to about 1e-7 for decay times in the neighborhood of the half-life.  This issue is fixed by just printing more digits.  Thanks to @asmuerer for pointing out this error.

Additionally, this also removes the rackspace upload part of the script in favor of simply building the tarball.  We are now uploading the tarball to pyne/data.

CC @FlanFlanagan

### DIFF
--- a/src/decaygen.py
+++ b/src/decaygen.py
@@ -327,23 +327,12 @@ def load_default_nucs():
 
 
 
-def upload(ns):
+def build_tarfile(ns):
     import tarfile
-    import pyrax
-    pyrax.set_setting('identity_type', 'rackspace')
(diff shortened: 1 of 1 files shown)


scopatz:
Also see pyne/data#2 for the newly generated file.

FlanFlanagan (approved):
Looks good to me.

scopatz:
Thanks @FlanFlanagan - can you also take a look at pyne/data#2, please? Thanks!

PR #918, opened 2017-10-04
Self shield

FlanFlanagan:
@scopatz


scopatz:
Thanks so much @FlanFlanagan!

PR #920, opened 2017-10-06
Atomic data generation

scopatz:


scopatz:
Some fixes for the code generation of atomic data.

PR #923, opened 2017-10-06
offsets only valid for ground state nuclide

scopatz:
Minor child() fix

### DIFF
--- a/tests/test_rxname.py
+++ b/tests/test_rxname.py
@@ -3,7 +3,7 @@ from __future__ import unicode_literals, division
 import sys
 import warnings
 
-import nose 
+import nose
 from nose.tools import assert_equal, assert_not_equal, assert_raises, raises, \
     assert_in, assert_greater_equal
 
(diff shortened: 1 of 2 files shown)


tests/test_rxname.py
```diff
@@ -24,9 +24,9 @@ def _hash(s):
     return h
 
 def test_hash():
-    rxs = ["a", "hello", "total", "wowza", "z_2na", "z_3na", "absorption", "np", 
-           "n2a", "z_2n2a", "nd", "nt", "nHe3", "nd3a", "nt2a", "z_4n", 
-           "fission_fourth", "z_2np", "z_3np", "n2p", "npa", "n_0", "n_1", "n_2", 
+    rxs = ["a", "hello", "total", "wowza", "z_2na", "z_3na", "absorption", "np",
```
FlanFlanagan, line 27: I've never seen this line before, "hello" and "wowza"?

FlanFlanagan: Not going to hold it up for this, just thought it was interesting!

scopatz: haha yeah, it is just there to test the hash function

PR #924, opened 2017-10-10
More mislabled metastables from ENSDF

scopatz:
CC @asmeurer  @FlanFlanagan


scopatz:
Note that this requires a `nuc_data_make --clean 1`

PR #925, opened 2017-10-13
New k coefficient calculation in decaygen

scopatz:
This should fix some of the issues generating branches that are not stable, but close to stable.  The previous method didn't have strong enough guarantees.

### DIFF
--- a/docs/theorymanual/decay.rst
+++ b/docs/theorymanual/decay.rst
@@ -1,25 +1,25 @@
 .. _theorymanual_decay:
 
 ===============================
-Decay 
+Decay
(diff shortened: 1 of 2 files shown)


scopatz:
Newly generated decayer is at https://github.com/pyne/data/pull/4

scopatz:
Also, this has a bug on the diagonal, hold off for a moment.

scopatz:
Actually, cancel that, this is right.  The error is not this PR.

scopatz:
I think that this is right, but there are still errors in ENSDF...

PR #926, opened 2017-10-13
More misbehaving metastable states

scopatz:
Pm-136 to start

### DIFF
--- a/pyne/ensdf.py
+++ b/pyne/ensdf.py
@@ -854,6 +854,25 @@ def _adjust_metastables(levellist):
             levellist[i] = tuple(row)
 
 
+# State Id, Rx Id : New Half-lives
+_BAD_HALF_LIVES = {
+    # Eu-151 lists a very long half-life (5.364792e+25) even though it
+    # lists no reaction, and thus no children, and no branch ratio.
+    # set to infinity for consistency.
+    (631510000, 0): float('inf'),
+    }
+
+
+def _adjust_half_lives(levellist):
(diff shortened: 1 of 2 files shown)


src/atomicgen.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
```
FlanFlanagan: Is this correct?

scopatz: Yes, you are never supposed to run from system python directly.  This runs with the user's python, which may be system one, but doesn't have to be.

FlanFlanagan:
FAIL: test_data.test_metastable_id
  File "/usr/lib/python3/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/robert/pyne/tests/test_data.py", line 200, in test_metastable_id
    assert_equal(data.metastable_id(310720000, 1), 310720002)
AssertionError: None != 310720002

Getting this error. Does this require the decay PR as well?

scopatz:
You need to do a `nuc_data_make --clean 1`, I think

scopatz:
I think that this is ready to go

scopatz:
After I look at fixing the tests...

scopatz:
This is superceded by #934

scopatz:
Actually, I was wrong, this needs to go in too. CC @FlanFlanagan @asmeurer @makeclean

PR #930, opened 2017-10-24
Fix search path for hdf5 and moab

### LINKED ISSUE #927: New build system requires sudo
I believe this issue is a result of PR #906 when the build system was overhauled. To build and install pyne, the `setup.py` command must be run in `sudo`. I think this is coming down to pyne trying build all shared object libraries (ace.so, dagmc.so, etc) into `/pyne/` in the root folder. It is unclear to me how to fix this based on the new system.

I don't think that pyne should require sudo to be built since some people are don't have administrative priveleges on their computer.

### LINKED ISSUE #928: Pyne having issues finding HDF5 if not using system HDF5
It used to be the case that pyne would automatically find the HDF5 in your `LD_LIBRARY_PATH`, but that appears to have changed in #908. Now it gives an error:

```
-- HDF5: Using hdf5 compiler wrapper to determine C configuration
CMake Error at /groupspace/cnerg/users/jacobson/opt/precompiled/cmake/3.9.4/share/cmake-3.9/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find HDF5 (missing: HDF5_LIBRARIES HDF5_INCLUDE_DIRS
  HDF5_HL_LIBRARIES)
Call Stack (most recent call first):
  /groupspace/cnerg/users/jacobson/opt/precompiled/cmake/3.9.4/share/cmake-3.9/Modules/FindPackageHandleStandardArgs.cmake:377 (_FPHSA_FAILURE_MESSAGE)
  /groupspace/cnerg/users/jacobson/opt/precompiled/cmake/3.9.4/share/cmake-3.9/Modules/FindHDF5.cmake:839 (find_package_handle_standard_args)
  CMakeLists.txt:45 (FIND_PACKAGE)

-- Configuring incomplete, errors occurred!
```

If I add the `-DHDF5_INCLUDE_DIRS=...` definition to my build command, then the error changes:

```
CMake Error at /groupspace/cnerg/users/jacobson/opt/precompiled/cmake/3.9.4/share/cmake-3.9/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find HDF5 (missing: HDF5_LIBRARIES HDF5_INCLUDE_DIRS
  HDF5_HL_LIBRARIES) (found version "1.8.13")
Call Stack (most recent call first):
  /groupspace/cnerg/users/jacobson/opt/precompiled/cmake/3.9.4/share/cmake-3.9/Modules/FindPackageHandleStandardArgs.cmake:377 (_FPHSA_FAILURE_MESSAGE)
  /groupspace/cnerg/users/jacobson/opt/precompiled/cmake/3.9.4/share/cmake-3.9/Modules/FindHDF5.cmake:839 (find_package_handle_standard_args)
  CMakeLists.txt:45 (FIND_PACKAGE)

-- Configuring incomplete, errors occurred!
```

I tried setting the `HDF5_ROOT` environment variable and setting various other build definitions, but none of them worked.

ljacobson64:
This fixes #927 and #928.

EDIT: it only fixes #928

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,8 +14,7 @@ jobs:
           name: Building PyNE
           command: |
             python3 setup.py install --user \
-                -DMOAB_LIBRARY=$HOME/opt/moab/lib \
-                -DMOAB_INCLUDE_DIR=$HOME/opt/moab/include
+                --moab=$HOME/opt/moab
             cd ~
             nuc_data_make --clean 1
 
@@ -40,8 +39,7 @@ jobs:
           name: Building PyNE
           command: |
             python setup.py install --user \
-                -DMOAB_LIBRARY=$HOME/opt/moab/lib \
-                -DMOAB_INCLUDE_DIR=$HOME/opt/moab/include
+                --moab=$HOME/opt/moab
             cd ~
             ~/.local/bin/nuc_data_make --clean 1
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,9 +59,9 @@ MESSAGE("--    HDF5 High Level Libraries: ${HDF5_C_HL_LIBRARIES}")
 
 # Find MOAB library
 find_package(MOAB)
-if(${MOAB_FOUND})
+if(MOAB_FOUND)
   include_directories(${MOAB_INCLUDE_DIRS})
-endif(${MOAB_FOUND})
+endif(MOAB_FOUND)
 
 
 # Find Python
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,7 @@ MESSAGE("--    HDF5 High Level Libraries: ${HDF5_C_HL_LIBRARIES}")
 find_package(MOAB)
 if(MOAB_FOUND)
   include_directories(${MOAB_INCLUDE_DIRS})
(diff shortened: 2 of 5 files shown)


scopatz:
Hi @ljacobson64 - Thanks for putting this in, but I think many of the proposed changes here will break other systems and removing the site packages will make it so that the Python extension modules are not installed and won't be linked correctly.  Let me look at the underlying issues.

ljacobson64:
I have tried `--hdf5` and `--moab` but neither seemed to work. I have not tried `--deps-root` but I'm not sure what I would even set it to since HDF5 and MOAB are in different directories.

With this PR, I'm able to build using `--hdf5` and `--moab` with an empty `LD_LIBRARY_PATH`. I feel this is the easiest way to tell users to specify the locations of HDF5 and MOAB since it does not require them to set any environment variables and it does not require them to set things like `-DMOAB_LIBRARY` and `-DMOAB_INCLUDE_DIR`.

It does look in the `LD_LIBRARY_PATH` to try to find MOAB, but only if `MOAB_ROOT` isn't set. (`--moab` sets `MOAB_ROOT`.)

ljacobson64:
Closing this because Anthony's solution looks to be better

PR #931, opened 2017-10-25
site packages location fix for Python 2

scopatz:
Hopefully this fixes the problem in #927 as reported by @kkiesling

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,8 +74,14 @@ endif(${MOAB_FOUND})
 # Find Python
 find_package(PythonInterp REQUIRED)
 find_package(PythonLibs REQUIRED)
-execute_process(COMMAND "${PYTHON_EXECUTABLE}" -c
-                        "import site; print(site.getsitepackages(['${CMAKE_INSTALL_PREFIX}'])[0])"
+if(PYTHON_VERSION_MAJOR EQUAL 2)
+  set(SITE_PACKAGES_CODE
+      "import site; print site.getusersitepackages().replace(site.USER_BASE, '${CMAKE_INSTALL_PREFIX}', 1)")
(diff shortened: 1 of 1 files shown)


scopatz:
OK This looks like it works on CI!

ljacobson64:
It seems to me like all the `.so` files are getting installed twice now: first by `setup_sub.py` and then again during `make install`. This is why everything still worked when I deleted the CMake lines that installed the libraries: they were already installed. They don't need to be installed twice, would you agree?

i.e. you'll see

```
copying build/lib.linux-x86_64-2.7/pyne/ace.so -> <install_prefix>/lib/python2.7/site-packages/pyne
```

and then later on you'll see

```
-- Installing: <install_prefix>/lib/python2.7/site-packages/pyne/ace.so
```

scopatz:
> They don't need to be installed twice, would you agree?

So actually one of the major benefits of the refactor is that they are installed twice.  The first time that they are installed (by setup_sub.py), they have the wrong RPATH setting. Specifically they have the `CMAKE_BUILD_RATH`.  So if you were to use this version, you  would **need** to have LD_LIBRARY_PATH set to the build path or a similar place where the libs could be found.  This is unavoidable in Python (without a lot of hacks) because Python and setuptools and distutils do not have a mechanism to set the RPATH of extension modules. 

The second time they are installed, they are installed by CMake.  CMake, alternatively, gives you complete control over the RPATH at both build and install.  By reinstalling with CMake *after* the Python install, we are able to easily set the correct RPATH.

scopatz:
Great question, btw @ljacobson64!

ljacobson64:
I see... so there is a method to the madness! I'll merge this and rebase my PR and we can go from there.

scopatz:
Thanks!

PR #933, opened 2017-10-27
Fix binpath in final note about setting PATH

ljacobson64:
The final note displayed after installing pyne that tells users to add pyne's `bin` directory to their path is listing the wrong directory. This PR fixes it.


scopatz:
Thanks @ljacobson64!

PR #934, opened 2017-10-27
Bateman solver handles Almost Stable and Degenerate half-lives

scopatz:
More details in the theory manual.  Also this includes the `moremetabad` commits from #926.  CC @asmeurer @FlanFlanagan

### DIFF
--- a/docs/theorymanual/decay.rst
+++ b/docs/theorymanual/decay.rst
@@ -79,8 +79,9 @@ from Es-254 to Po-210 goes through U-238, which is very close to stable relative
 other nuclides in the chain. This can trigger floating point precision issues, where certain
 terms will underflow or overflow or generate NaNs. Obviously this is a situation to be avoided,
 if at all possible. To handle this sitiuation, let's call :math:`p` the index of the nuclide
-that is almost stable. We can then take the limit of the Bateman equations where
-:math:`\lambda_p \to 0` after we separate out the p-term from the summation:
+that is almost stable. We can then note that the Bateman equations can be reduced by the
+observation that :math:`\lambda_p \ll \lambda_{i\ne p}` after we separate out the p-term
+from the summation:
 
 .. math::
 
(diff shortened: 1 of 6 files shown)


scopatz:
OK! This is good to go!

scopatz:
After this is merged, a new PR should go into pyne/data updating decay.tar.gz, new containers should be built, and the edits to the circle config file should be removed.

docs/theorymanual/decay.rst
```diff
@@ -74,6 +74,84 @@ in front of the summation:
...
+
+   \lim_{\lambda_p\to 0}\frac{N_C(t)}{N_1(0)} = \frac{\gamma}{\lambda_C}\sum_{i\ne p}^C \lim_{\lambda_p\to 0} \left[\lambda_i \frac{\lambda_p}{\lambda_p - \lambda_i}
+                                                        \left(\prod_{j\ne i,p}^C \frac{\lambda_j}{\lambda_j - \lambda_i}\right)
+                                                        e^{-\lambda_i t}\right]
+                           + \lim_{\lambda_p\to 0} \frac{\gamma}{\lambda_C} \lambda_p \left(\prod_{j\ne p}^C \frac{\lambda_j}{\lambda_j - \lambda_p}\right) e^{-\lambda_p t}
+
+   \lim_{\lambda_p\to 0}\frac{N_C(t)}{N_1(0)} = \frac{\gamma}{\lambda_C}\sum_{i\ne p}^C \left[\lambda_i \frac{\lambda_p}{- \lambda_i}
+                                                        \left(\prod_{j\ne i,p}^C \frac{\lambda_j}{\lambda_j - \lambda_i}\right)
+                                                        e^{-\lambda_i t}\right]
+                           + \frac{\gamma}{\lambda_C} \lambda_p \left(\prod_{j\ne p}^C \frac{\lambda_j}{\lambda_j}\right) e^{-\lambda_p t}
```
asmeurer: Is this right? You still have \lambda_p's left in the expression after removing the limit.

asmeurer:
Some of the math you added leads to latex errors in the compiled docs.

docs/theorymanual/decay.rst
```diff
@@ -104,18 +182,50 @@ where the coefficients :math:`k_i` are defined as:
...
+    k_{i\ne C} = -\gamma \prod_{j=1,i\ne j}^{C-1} \frac{\lambda_j}{\lambda_j - \lambda_i}
+
+    k_C = \gamma
+
+
+**Last Nuclide Unstable and pth Almost Stable:**
+
+.. math::
+
+    k_{i\ne p} = -\frac{\gamma\lambda_p}{\lambda_C} \prod_{j\n i,p}^C \frac{\lambda_j}{\lambda_j - \lambda_i}
```
asmeurer: Specifically here, and below, you have `\n` instead of `\ne`.

scopatz: Ahh thanks

scopatz: Should be fixed now

docs/theorymanual/decay.rst
scopatz: Yes, I believe so, because `\lambda_p` isn't zero, it just almost zero and in some places that causes problems for floating point arithmetic.  So the limit is only taken for those reactions.  Said another way `\lambda_p << \lambda_j`

asmeurer: So perhaps instead of using limits, one could write each step with \approx based on \lambda_p << \lambda_j.

scopatz: Yes, that would be better in some cases.  I'll rewrite soon.

scopatz:
Updated some of the language as per @asmeurer's suggestion

FlanFlanagan:
```
FAIL: test_data.test_metastable_id
  File "/usr/lib/python3/dist-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/robert/pyne/tests/test_data.py", line 203, in test_metastable_id
    assert_equal(data.metastable_id(611360000, 1), 611360001)
AssertionError: 611360000 != 611360001

```

Getting this error with a clean build. Is this part of the decay.tar.gz issue?

scopatz:
Did you rebuild nud_data data after you installed? Also, you may need to reinstall after rebuild the nuc_data.

FlanFlanagan:
That fixed it!

scopatz:
Sweet!

PR #936, opened 2017-11-07
Voxel r2s

zxkjack123:
Main changes:

- Remove not used mat_* tag of class mesh in mesh.py

- Add a parameter `sub_voxel` that enable user to choose sub-voxel r2s

- Duplicate neutron flux for each sub-voxel in ALARA fluxin

- Eliminate  fluxes of void sub-voxel from ALARA fluxin

- Add a test function to test ALARA fluxin in pyne/tests/test_alara.py

- Add two files of expected ALARA fluxin in pyne/tests/files_test_alara/


pyne/alara.py
```diff
@@ -48,6 +49,30 @@ def mesh_to_fluxin(flux_mesh, flux_tag, fluxin="fluxin.out",
     reverse : bool
         If true, fluxes will be printed in the reverse order as they appear in
         the flux vector tagged on the mesh.
+    sub_voxel: bool, optional
+        If true, sub-voxel r2s work flow will be sued. Flux of a voxel will
+        be duplicated c times. Where c is the cell numbers of that voxel.
+    cell_fracs : structured array
```
scopatz: This should be listed as optional

pyne/alara.py
```diff
@@ -48,6 +49,30 @@ def mesh_to_fluxin(flux_mesh, flux_tag, fluxin="fluxin.out",
...
+            :cell: int
+                The geometry cell number.
+            :vol_frac: float
+                The volume fraction of the cell withing the mesh ve.
+            :rel_error: float
+                The relative error associated with the volume fraction.
+
+        The array must be sorted with respect to both idx and cell, with
+        cell changing fastest.
+    cell_mats : dict
```
scopatz: This should also be listed as optional

pyne/alara.py
```diff
@@ -70,18 +95,35 @@ def mesh_to_fluxin(flux_mesh, flux_tag, fluxin="fluxin.out",
...
+            for i in range(start, stop, direction):
+                output += "{:.6E} ".format(flux_data[i])
+                # fluxin formatting: create a new line after every 6th entry
+                count += 1
+                if count % 6 == 0:
+                    output += "\n"
+
+            output += "\n\n"
+
+    if sub_voxel:
```
scopatz: The should be an `else:`

scopatz:
LGTM after a couple of modifications!

zxkjack123:
There are 3 changes according to the comments of @scopatz .

pyne/alara.py
```diff
@@ -70,18 +95,35 @@ def mesh_to_fluxin(flux_mesh, flux_tag, fluxin="fluxin.out",
...
+                    flux_data = np.atleast_1d(tag_flux[ve])
+                    for i in range(start, stop, direction):
+                        output += "{:.6E} ".format(flux_data[i])
+                        # fluxin formatting: create a new line
+                        # after every 6th entry
+                        count += 1
+                        if count % 6 == 0:
+                            output += "\n"
+
+                    output += "\n\n"
```
scopatz: Since the body of these two clause is mostly the same, can you maybe please provide a private method (ie a method that begins with an underscore `_`) or similar that performs the work.  This function would then just be called with different arguments.  There is a lot of code duplication here that could be removed :) Thanks!

zxkjack123: Thanks for your suggestion @scopatz . I put these two clause into a private method `_output_flux()` to reduce duplication.

pyne/alara.py
```diff
@@ -704,3 +727,24 @@ def cram(N, t, n_0, order):
...
         raise ValueError(msg)
 
+def _output_flux(ve, tag_flux,output,start,stop,direction):
+    """
+    This function is used to write netron flux into fluxin
+    :param ve: entity
+    :param tag_flux:
+    :param output: string
+    :return: string
+    """
```
scopatz: Please write this docstring in numpydoc format, so that it conforms to the other docstrings throughout pyne. Thanks!

zxkjack123: I rewrote this docstring in numpydoc format.

scopatz:
Thanks @zxkjack123!

PR #938, opened 2017-11-09
subvoxel r2s record_to_geom

zxkjack123:
Major changes:

- `alara.py`: edit function `record_to_geom()`, add code for sub-voxel r2s. write sub-voxel R2S alara input

- `r2s.py`: add a parameter `sub_voxel`

- `test_alara.py`: add a test function `test_record_to_geom_subvoxel()` to test sub-voxel R2S, ALARA input, geom part

- Add two files in `pyne/tests/files_test_alara/` as expect results of test function.


tests/test_alara.py
```diff
@@ -285,6 +285,51 @@ def test_record_to_geom():
...
+
+    cell_fracs[:] = [(0, 11, 0.55, 0.0), (0, 12, 0.45, 0.0), (1, 11, 0.2, 0.0),
+                     (1, 12, 0.3, 0.0), (1, 13, 0.5, 0.0), (2, 11, 0.15, 0.0),
+                     (2, 14, 0.01, 0.0), (2, 15, 0.04, 0.0), (2, 16, 0.8, 0.0),
+                     (3, 11, 0.55, 0.0), (3, 12, 0.45, 0.0)]
+
+    m = Mesh(structured_coords=[[-1, 0, 1], [-1, 0, 1], [0, 1]],
+             structured=True, mats=None)
+
+    record_to_geom(m, cell_fracs, cell_mats, geom, matlib,sub_voxel=True)
```
scopatz: PEP8: single space after comma

scopatz:
This looks good to me @zxkjack123 - but I am not an expert. @makeclean or @gonuke or @kkiesling, do you want to take a look?

pyne/alara.py
```diff
@@ -287,31 +289,51 @@ def record_to_geom(mesh, cell_fracs, cell_mats, geom_file, matlib_file,
...
+                    mixture += '    material {0} 1 {1}\n'.format(key, value)
+
+                mixture += 'end\n\n'
+
+            mat_loading += '    zone_{0}    mix_{1}\n'.format(i,
+                            unique_mixtures.index(ve_mixture))
+    else:
+        sve_count = 0
+        for row in cell_fracs:
+            for i, mat, ve in mesh:
```
gonuke: These loops seem backwards - we probably don't to loop over the whole mesh for each row in `cell_fracs`

pyne/alara.py
```diff
@@ -287,31 +289,51 @@ def record_to_geom(mesh, cell_fracs, cell_mats, geom_file, matlib_file,
...
-
-        if ve_mixture not in unique_mixtures:
-            unique_mixtures.append(ve_mixture)
-            mixture += 'mixture mix_{0}\n'.format(
-                                           unique_mixtures.index(ve_mixture))
-            for key, value in ve_mixture.items():
-                mixture += '    material {0} 1 {1}\n'.format(key, value)
-
-            mixture += 'end\n\n'
+    if not sub_voxel:
```
gonuke: I think this conditional can be at least one layer deeper in the loops, esp. once you have reordered the loops for the subvoxel case.

It may be appropriate to use the conditional in many separate places rather than have two very similar blocks.

pyne/alara.py
gonuke: I just noticed the same in #936

gonuke: Maybe this is related to how we ensure consistent ordering?  but this loop ordering is going to be extremely!!! Inefficient for large mesh:  For a mesh with N voxels, `cell_fracs` will have O(N) rows and mesh will have N entries, so it will be N^2

gonuke (changes requested):
The loop ordering here (and in #936 ) is very inefficient.

pyne/alara.py
zxkjack123: Although these two blocks are both used to materials information to ALARA input file, but actually they have a big different. 

For the normal R2S, we need to name the `mixture` composed of different materials, and assign a id to this `mixture`, however, we don't need to do that in sub-voxel R2S. Because the material of each sub-voxel is the same as corresponding cell. So I write `material.name` to ALARA input file to simplify this process. 

These two blocks look similar but do writing different content. It would make the code more complex if we make the conditional be deeper in the loops, because we need to make conditional judgement for almost every line of the code.

zxkjack123:
`ves = list(mesh.iter_ve())`and `ves[row['idx']]` are used to replace the loop: `for i, _, ve in mesh:`

gonuke (approved):

gonuke:
Thanks @zxkjack123

PR #939, opened 2017-11-10
update for CI

scopatz:
Don't need to bootstrap anymore


scopatz:
CC @pyne/developers

ljacobson64:
Why don't we need this anymore?

ljacobson64:
@scopatz

scopatz:
Sorry I missed this! We don't need this anymore because I updated the docker containers to include the latest bateman solver and data.  Thus we don't need to regenerated it in CI for the tests to pass.

ljacobson64:
I see!

PR #929, opened 2017-10-23
Feature cell fracs

zxkjack123:
Main changes:

- Add a method `tag_cell_fracs()` to `pyne/pyne/mesh.py:Mesh`. 

- Add a function `test_tag_cell_fracs()` in `pyne/tests/test_mesh.py` to test the adding method

Some white space in `pyne/tests/test_mesh.py` are deleted.

PR #940, opened 2017-11-10
update_cell_fracs_tag

zxkjack123:
Major change:

- Use function tag() in mesh.py: class Mesh, to create `cell_number_tag`, `cell_fracs_tag`, `cell_largest_frac_number_tag` and `cell_largest_frac_tag`. Replace the previous method.


pyne/mesh.py
```diff
@@ -775,11 +775,6 @@ def __init__(self, mesh=None, structured=False,
                 setattr(self, name, MaterialMethodTag(mesh=self, name=name,
                         doc=doc))
 
-        # add a property: cell_cell_fracs_tag and set it to None
-        self.cell_number_tag = None
-        self.cell_fracs_tag = None
-        self.cell_largest_frac_number_tag = None
-        self.cell_largest_frac_tag = None
```
scopatz, line 782: I am not sure that it makes sense to delete these, per se.

zxkjack123: The `setattr()` was called at the end of function `self.tag()`. Therefore, `self.tag()` adds a new tag to the mesh. So we don't need to define these four property members in class Mesh. They will be created when we need to use these data, i.e., when the function `tag_cell_fracs()` was called.

On the other hand, if I want to use the `self.tag()` to create these tags, I have to delete them. Because `self.tag()` won't create a tag if the tag name is exist:
```
if name in self.tags:
            raise KeyError('{0} tag already exists on the mesh'.format(name))
```

PS: These changes have passed the test functions in tests/test_mesh.py. As you can see, I didn't change the test_mesh.py in this pull request. They give the identical result as previous method for tagging cell_fracs.

zxkjack123:
Could you please review this pull request and give some comments or suggestions? @gonuke @pshriwise

pyne/mesh.py
```diff
@@ -1259,14 +1246,23 @@ def tag_cell_fracs(self, cell_fracs):
...
-        self.cell_largest_frac_number_tag = cell_largest_frac_number
-        self.cell_largest_frac_tag = cell_largest_frac
-
-
-
 
+        # creat the tags
+        self.tag(name='cell_number_tag', value=voxel_cell_number,
+                 doc='cell numbers of the voxel, -1 used to fill vacancy',
+                 tagtype=IMeshTag, size=max_cell_number, dtype = int)
```
pshriwise: `dtype = int` -> `dtype=int`  small formatting thing

pyne/mesh.py
```diff
@@ -1232,21 +1227,13 @@ def tag_cell_fracs(self, cell_fracs):
...
+
+        # creat tag frame with default value
         cell_largest_frac_number = [-1] * num_vol_elements
         cell_largest_frac = [0.0] * num_vol_elements
-
-        # fill the data
         voxel_cell_number = np.empty(shape=(num_vol_elements,max_cell_number))
         voxel_cell_fracs = np.empty(shape=(num_vol_elements,max_cell_number))
+        # set the data
         for ve in range(num_vol_elements):
```
pshriwise: I think it might be best to use the `iter_ve()` method here rather than range for consistency when setting up this loop. An example of this can be found earlier in this file at line 725 [here](https://github.com/zxkjack123/pyne/blob/012c5afc99b343780cbd06ea076bb422e8e6d0b4/pyne/mesh.py#L725) with a loop over that iterator [here](https://github.com/zxkjack123/pyne/blob/012c5afc99b343780cbd06ea076bb422e8e6d0b4/pyne/mesh.py#L750)

This comment doesn't have to do with this PR, so you can do what you will with this info :smiley:

pyne/mesh.py
pshriwise: @zxkjack123 I'm not sure that you *have* to delete them, because it seems that new tags aren't actually registered as part of the `self.tags` dictionary attribute anywhere in the `mesh.tag` method so this wouldn't be a problem. Did you actually get this error when trying it?

At any rate, until the `tag_cell_fracs` method is called, I don't see a reason to keep them around. Did you have a specific purpose for them being around but set to `None` @scopatz? Their existence can be checked for using the `hasattr` built-in if they're needed, and if they don't exist then your new method can be called to create them.

pshriwise (commented):
I think these are good changes, @zxkjack123. Given that there are many different tag types allowed in a pyne mesh (some of which are not `iMeshTag`'s) it is good to use the more general interface to create them.

pyne/mesh.py
zxkjack123: @pshriwise Actually, I haven't tried to call updated `tag_cell_fracs` with the `self.cell_number_tag`, .... And these tag names are not registered as part of the `self.tags`. I mistook these tag names with the `self.tags`. 

I agree with you that these tags should not be around but set to `None`, as they are necessary only for application of sub-voxel r2s.

pyne/mesh.py
zxkjack123: The `ve` here doesn't represent an entity volume element, it more like the `idx` of the volume element. This `ve` here are [0, 1, 2, ...]. @pshriwise Do you think it's better to replace this `ve` with `i` or `idx`?

pyne/mesh.py
zxkjack123: Format changed according to the comments of @pshriwise .

pyne/mesh.py
pshriwise: Ah, my misunderstanding. Sorry. I think changing the variable might help with clarity here, yes. 

Also I'd recommend taking a look at the NumPy `fill` command. It could replace the first couple lines of this loop I think.

zxkjack123: I replaced the `ve` with `i` in these loop. And the Numpy `fill` command was applied to replace two lines of that loop.

pshriwise: Nice! Looks good.

pyne/mesh.py
```diff
@@ -1229,44 +1224,47 @@ def tag_cell_fracs(self, cell_fracs):
...
-            voxel_cell_number[ve] = [-1] * max_cell_number
-            voxel_cell_fracs[ve] = [0.0] * max_cell_number
-            for (cell, row) in enumerate(cell_fracs[cell_fracs['idx'] == ve]):
-                voxel_cell_number[ve, cell] = row['cell']
-                voxel_cell_fracs[ve, cell] = row['vol_frac']
+        voxel_cell_number = np.empty(shape=(num_vol_elements,max_cell_number),
+                                     dtype=int)
+        voxel_cell_fracs = np.empty(shape=(num_vol_elements,max_cell_number),
+                                    dtype=float)
+        voxel_cell_number.fill(-1)
```
zxkjack123: Numpy `fill` command was used to replace two lines in loop.

pyne/mesh.py
```diff
@@ -1229,44 +1224,47 @@ def tag_cell_fracs(self, cell_fracs):
         num_vol_elements = len(self)
         # Find the maximum cell number in a voxel
         max_cell_number = -1
-        for ve in range(num_vol_elements):
```
zxkjack123: Use `i` to replace `ve` for clarity.

scopatz:
@pshriwise @gonuke @makeclean @ljacobson64 - is there a verdict on this PR?

pyne/mesh.py
gonuke: Why is `i` more clear than `ve`?  Seems the opposite to me.

pyne/mesh.py
```diff
@@ -1229,44 +1224,47 @@ def tag_cell_fracs(self, cell_fracs):
         num_vol_elements = len(self)
         # Find the maximum cell number in a voxel
         max_cell_number = -1
```
gonuke: I missed this in #929, but I recommend calling this `max_num_cells`.  This variable name implies a different meaning: the largest cell ID in the problem.

pyne/mesh.py
```diff
@@ -1229,44 +1224,47 @@ def tag_cell_fracs(self, cell_fracs):
...
-                voxel_cell_number[ve, cell] = row['cell']
-                voxel_cell_fracs[ve, cell] = row['vol_frac']
+        voxel_cell_number = np.empty(shape=(num_vol_elements,max_cell_number),
+                                     dtype=int)
+        voxel_cell_fracs = np.empty(shape=(num_vol_elements,max_cell_number),
+                                    dtype=float)
+        voxel_cell_number.fill(-1)
+        voxel_cell_fracs.fill(0.0)
+        # set the data
+        for i in range(num_vol_elements):
```
gonuke, line 1241: again `ve` is better than `i`

gonuke (changes requested):
This looks like a nice cleanup of the tagging, but I have a few comments about variable name choice.

pyne/mesh.py
pshriwise: I recommended that something other than `ve` be used here, because `ve` is used [in other places](https://github.com/zxkjack123/pyne/blob/0c9c5b874879d0fa415d4ee35386831afd5634e3/pyne/mesh.py#L725) to represent a volume element rather than index to a volume element, which is what `i` is here. Maybe `ve_idx` could work?

I *think* (though I'm not sure, so maybe @zxkjack123 can comment further) that the normal mesh iterator isn't used here because the index is also needed to access parts of data arrays.

zxkjack123: @gonuke @pshriwise 
In the pyne/pyne/mesh.py, the function `__iter__(self)`:
```
def __iter__(self):
        """Iterates through the mesh and at each step yield the volume element
        index i, the material mat, and the volume element itself ve.
        """
        mats = self.mats
        if mats is None:
            for i, ve in enumerate(self.iter_ve()):
                yield i, None, ve
        else:
            for i, ve in enumerate(self.iter_ve()):
                yield i, mats[i], ve
```

`i` is more commonly used as a represent of `idx`, while `ve` represents a volume element entity.  
Also, there are lots of 
```
for i, ve in enumerate(ves):
``` 
or 
```
for i, _, ve in mesh:
```
in PyNE codes. 
In this case, I think it's better to use `ve` only when it represents a volume element entity. And use `i` when it represents an index: `idx`.

zxkjack123:
Variable `max_cell_number` are replaced with `max_num_cells` for clarity.

gonuke:
It seems like all of these changes were in #938 Can you rebase?

On a related note - does anyone know why GH doesn't update the diffs automatically?  I feel like it used to, but the diffs here appear to be relative to the state of `develop` at the time of the branch and not the current state of `develop`

zxkjack123:
Rebased the pull request on current `develop`.

scopatz:
> On a related note - does anyone know why GH doesn't update the diffs automatically? I feel like it used to, but the diffs here appear to be relative to the state of develop at the time of the branch and not the current state of develop

Hmm oddd... Not sure what is going on there

gonuke:
When I look at this branch manually, it is not different from `develop` at all - other than a few whitespace changes

scopatz:
It seems that the rebase rewrote all of the commits that where on develop....

zxkjack123:
@gonuke Some of the commits are also merged into another branch which has merged into PyNE. So, it's true that this pull request didn't make big changes any more. But the system says that this branch is still 3 commits ahead of `develop`, maybe some clean up.

pshriwise:
my guess is that this is going to require an interactive rebase to clean up the history. Perhaps @zxkjack123  and I can take a look at this in person.

zxkjack123:
Most of the commits in this branch are also in another branch that has been merged into pyne. So there are only two whitespace changes from develop now.

gonuke:
The changes originally introduced here ended up being merged in #938

PR #941, opened 2017-11-14
Subvoxel read alara fluxout

zxkjack123:
Note: This pull request from branch `subvoxel_read_alara_fluxout` is dependent on a previous pull request #938 . 
Major changes:
For pyne/pyne/alara.py:
- Edit function `photon_source_hdf5_to_mesh()`: add a new method for sub-voxel r2s photon source treating
- Add private function `_get_non_void_sub_voxel_number()` to calculate the total number of non-void sub-voxel
- Add private function `_get_subvoxel_cell_idx()` to calculate the voxel idx and the cell index in the voxel

For pyne/tests/test_alara.py:
- Add test function `test_photon_source_hdf5_to_mesh_subvoxel()` to test the changes in the `alara.py`.

### DIFF
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -258,17 +258,15 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
                     tag_handles[tags[cond]][ve] = [0] * num_e_groups
         else:
             temp_mesh_data = np.empty(
-                shape=(num_vol_elements, num_e_groups * max_num_cells),
+                shape=(num_vol_elements, max_num_cells, num_e_groups),
                 dtype=float)
             temp_mesh_data.fill(0.0)
             for sve, subvoxel in enumerate(subvoxel_array):
-                    # start index of data
-                    d_s = subvoxel['scid'] * num_e_groups
-                    # end index of data
-                    d_e = (subvoxel['scid'] + 1) * num_e_groups
-                    temp_mesh_data[subvoxel['idx'],d_s:d_e] = matched_data[sve][3][:]
+                temp_mesh_data[subvoxel['idx'],subvoxel['scid'],:] = \
+                    matched_data[sve][3][:]
             for i, _, ve in mesh:
-                tag_handles[tags[cond]][ve] = temp_mesh_data[i,:]
+                tag_handles[tags[cond]][ve] = \
+                    temp_mesh_data[i,:].reshape(max_num_cells * num_e_groups)
 
 def record_to_geom(mesh, cell_fracs, cell_mats, geom_file, matlib_file,
                    sig_figs=6, sub_voxel=False):
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -218,12 +218,12 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
     with tb.open_file(filename) as h5f:
         num_e_groups = len(h5f.root.data[0][3])
     max_num_cells = -1
-    ves = list(mesh.iter_ve())
+    ve0 = next(mesh.iter_ve())
     if sub_voxel:
         num_vol_elements = len(mesh)
         subvoxel_array = _get_subvoxel_array(mesh, cell_mats)
         max_num_cells = \
-                len(mesh.mesh.getTagHandle('cell_number')[ves[0]])
+                len(mesh.mesh.getTagHandle('cell_number')[ve0])
     else:
         max_num_cells = 1
 
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -258,9 +258,9 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
                     tag_handles[tags[cond]][ve] = [0] * num_e_groups
         else:
             temp_mesh_data = np.empty(
-                shape=(num_vol_elements, num_e_groups * max_num_cells))
-            for i in range(num_vol_elements):
-                temp_mesh_data[i] = [0.0] * num_e_groups * max_num_cells
+                shape=(num_vol_elements, num_e_groups * max_num_cells),
+                dtype=float)
+            temp_mesh_data.fill(0.0)
             for sve, row in enumerate(subvoxel_array):
                     # start index of data
                     d_s = row['scid'] * num_e_groups
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -261,12 +261,12 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
                 shape=(num_vol_elements, num_e_groups * max_num_cells),
                 dtype=float)
             temp_mesh_data.fill(0.0)
-            for sve, row in enumerate(subvoxel_array):
+            for sve, subvoxel in enumerate(subvoxel_array):
                     # start index of data
-                    d_s = row['scid'] * num_e_groups
+                    d_s = subvoxel['scid'] * num_e_groups
                     # end index of data
-                    d_e = (row['scid'] + 1) * num_e_groups
-                    temp_mesh_data[row['idx'],d_s:d_e] = matched_data[sve][3][:]
+                    d_e = (subvoxel['scid'] + 1) * num_e_groups
+                    temp_mesh_data[subvoxel['idx'],d_s:d_e] = matched_data[sve][3][:]
             for i, _, ve in mesh:
                 tag_handles[tags[cond]][ve] = temp_mesh_data[i,:]
 
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -829,20 +829,19 @@ def _get_subvoxel_array(mesh, cell_mats):
     """
     cell_number_tag = mesh.mesh.getTagHandle('cell_number')
     subvoxel_array = np.zeros(0, dtype=[(b'svid', np.int64),
-                                                      (b'idx', np.int64),
-                                                      (b'scid', np.int64)])
+                                        (b'idx', np.int64),
+                                        (b'scid', np.int64)])
     temp_subvoxel = np.zeros(1, dtype=[(b'svid', np.int64),
-                                                      (b'idx', np.int64),
-                                                      (b'scid', np.int64)])
+                                       (b'idx', np.int64),
+                                       (b'scid', np.int64)])
     # calculate the total number of non-void sub-voxel
     non_void_sv_num = 0
     for i, _, ve in mesh:
         for c, cell in enumerate(cell_number_tag[ve]):
-            if cell > 0:
-                if len(cell_mats[cell].comp):
-                    temp_subvoxel[0] = (non_void_sv_num, i, c)
-                    subvoxel_array = np.append(subvoxel_array, temp_subvoxel)
-                    non_void_sv_num += 1
+            if cell > 0 and len(cell_mats[cell].comp): #non-void cell
+                temp_subvoxel[0] = (non_void_sv_num, i, c)
+                subvoxel_array = np.append(subvoxel_array, temp_subvoxel)
+                non_void_sv_num += 1
 
     return subvoxel_array
 
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -181,7 +181,8 @@ def photon_source_to_hdf5(filename, chunkshape=(10000,)):
     f.close()
 
 
-def photon_source_hdf5_to_mesh(mesh, filename, tags):
+def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
+                               cell_mats=None):
     """This function reads in an hdf5 file produced by photon_source_to_hdf5
     and tags the requested data to the mesh of a PyNE Mesh object. Any
     combinations of nuclides and decay times are allowed. The photon source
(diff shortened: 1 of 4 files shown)


zxkjack123:
Submit replace `cell_fracs` with `subvoxel_array` aims to use the `cell_number_tag` of a mesh to rebuild cell fraction information to replace the input parameter ``cell_fracs.  So that we don't need to calculate the `cell_fracs` in r2s step2. 

Major changes:
In pyne/pyne/alara.py:
- Delete input parameter `cell_fracs'
- Delete no longer used private functions `_get_non_void_sub_voxel_number()` and `_get_subvoxel_cell_idx()`
- Add private function `_get_subvoxel_array()` to build cell fraction information from mesh
- Update function code, use `subvoxel_array` to replace `cell_fracs`

In pyne/tests/test_alara.py:
- Delete input  parameter `cell_fracs` of function`photon_source_hdf5_to_mesh()` in test function `test_photon_source_hdf5_to_mesh_subvoxel()`

zxkjack123:
This pull request is relay on the #940 , which has been merged into this pull request.

scopatz:
CC @pshriwise @makeclean @gonuke @ljacobson64

pyne/alara.py
```diff
@@ -205,16 +206,34 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags):
...
+        Then the photon_source will be interpreted as sub-voxel photon source.
+    cell_mats : dict, optional
+        cell_mats is required when sub_voxel is True.
+        Maps geometry cell numbers to PyNE Material objects.
     """
+
     # find number of energy groups
     with tb.open_file(filename) as h5f:
         num_e_groups = len(h5f.root.data[0][3])
+    max_cell_number = -1
```
gonuke: As in #940, I think this should be `max_num_cells` instead.

pyne/alara.py
```diff
@@ -205,16 +206,34 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags):
...
+    if sub_voxel:
+        num_vol_elements = len(mesh)
+        subvoxel_array = _get_subvoxel_array(mesh, cell_mats)
+        for i, _, ve in mesh:
+            max_cell_number = \
+                len(mesh.mesh.getTagHandle('cell_number')[ve])
```
gonuke: Since this tag is created with `size=max_cell_number`, won't every volume element have the same length tag?  Therefore, you should only have to test the first volume element.

pyne/alara.py
```diff
@@ -753,3 +811,42 @@ def _output_flux(ve, tag_flux,output,start,stop,direction):
...
+    """
+    This function returns an array of subvoxels.
+    Parameters
+    ----------
+    mesh : PyNE Mesh object
+        The Mesh object for which the geometry is discretized.
+
+    return : subvoxel_array: structured array
+        A sorted, one dimensional array, each entry containing the following
+        fileds:
```
gonuke: typo

pyne/alara.py
```diff
@@ -230,16 +249,33 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags):
...
+                    idx += 1
+                else:
+                    tag_handles[tags[cond]][ve] = [0] * num_e_groups
+        else:
+            temp_mesh_data = np.empty(
+                shape=(num_vol_elements, num_e_groups * max_cell_number))
+            for ve in range(num_vol_elements):
+                temp_mesh_data[ve] = [0.0] * num_e_groups * max_cell_number
+            for sve, row in enumerate(subvoxel_array):
+                for i, _, ve in mesh:
```
gonuke: Loop ordering again...  I think `subvoxel_array` is created by looping through the mesh once already, so maybe it can be done more efficiently here

pyne/alara.py
```diff
@@ -230,16 +249,33 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags):
...
+                    if row['idx'] == i:
+                        # start index of data
+                        d_s = row['scid'] * num_e_groups
+                        # end index of data
+                        d_e = (row['scid'] + 1) * num_e_groups
+                        temp_mesh_data[row['idx'],d_s:d_e] = \
+                            matched_data[sve][3][:]
+            for i, _, ve in mesh:
```
gonuke: If you reverse the loop ordering above, you might not need this loop again, and `temp_mesh_data` might be able to have one fewer dimension

pyne/alara.py
```diff
@@ -287,31 +325,51 @@ def record_to_geom(mesh, cell_fracs, cell_mats, geom_file, matlib_file,
...
-
-        if ve_mixture not in unique_mixtures:
-            unique_mixtures.append(ve_mixture)
-            mixture += 'mixture mix_{0}\n'.format(
-                                           unique_mixtures.index(ve_mixture))
-            for key, value in ve_mixture.items():
-                mixture += '    material {0} 1 {1}\n'.format(key, value)
-
-            mixture += 'end\n\n'
+    if not sub_voxel:
```
gonuke: Not commenting since I think this all comes from another PR...

gonuke (changes requested):
I think we should discuss some of the data structures and loop ordering questions.

I'll need to review again #938 and #940 are merged

pyne/alara.py
zxkjack123: Replaced `fileds` to `fields`.

pyne/alara.py
zxkjack123: I found that the second loop `for i, _, ve in mesh:` is actually not used. So I just delete this second loop.

pyne/alara.py
zxkjack123: Changed according to your comment. Use the first volume element to calculate the `max_num_cells`.

pyne/alara.py
zxkjack123: All the `max_num_cells` are replaced with `max_num_cells`.

zxkjack123:
Changes in the new commits:
1. Replace `max_cell_number` with `max_num_cells`.
2. Delete redundant loop.
3. Fix a typo.

gonuke:
Please rebase this against `develop` and resolve conflicts

zxkjack123:
Conflicts between branch subvoxel_read_alara_fluxout and base branched are solved. @gonuke Is there any other change required?

gonuke:
This also needs to be rebased against `develop`

zxkjack123:
Commits from others are removed form this branch.

pyne/alara.py
```diff
@@ -205,16 +206,33 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags):
...
     # find number of energy groups
     with tb.open_file(filename) as h5f:
         num_e_groups = len(h5f.root.data[0][3])
+    max_num_cells = -1
+    ves = list(mesh.iter_ve())
+    if sub_voxel:
+        num_vol_elements = len(mesh)
+        subvoxel_array = _get_subvoxel_array(mesh, cell_mats)
+        max_num_cells = \
+                len(mesh.mesh.getTagHandle('cell_number')[ves[0]])
```
scopatz: If you only need the first element of ves, you don't need to convert the whole thing to a list and then index. You can just call `next()` on the iterator:  `ve0 = next(mesh.iter_ve())`

scopatz:
Thanks for cleaning this up @zxkjack123

pyne/alara.py
zxkjack123: @scopatz Thanks, I changed this part according to your suggestion.

pyne/alara.py
```diff
@@ -230,13 +248,27 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags):
...
+            idx = 0
+            for i, _, ve in mesh:
+                if matched_data[idx][0] == i:
+                    tag_handles[tags[cond]][ve] = matched_data[idx][3]
+                    idx += 1
+                else:
+                    tag_handles[tags[cond]][ve] = [0] * num_e_groups
+        else:
+            temp_mesh_data = np.empty(
+                shape=(num_vol_elements, num_e_groups * max_num_cells))
```
gonuke: why not shape this as a 3-d array?

pyne/alara.py
```diff
@@ -230,13 +248,27 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags):
...
+            for i in range(num_vol_elements):
+                temp_mesh_data[i] = [0.0] * num_e_groups * max_num_cells
```
gonuke: Isn't this already accomplished by creating this as `np.empty()`

pyne/alara.py
```diff
@@ -230,13 +248,27 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags):
...
+            for sve, row in enumerate(subvoxel_array):
```
gonuke: Is there a better name than `row`? Isn't each `row` a `subvoxel`?

pyne/alara.py
```diff
@@ -230,13 +248,27 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags):
...
+                    # start index of data
+                    d_s = row['scid'] * num_e_groups
```
gonuke: This would be easier with the 3-d array mentioned above.

pyne/alara.py
```diff
@@ -774,3 +806,43 @@ def _output_flux(ve, tag_flux,output,start,stop,direction):
...
+                The index of non-void subvoxel id
+            :idx: int
+                The idx of the voxel
+            :scid: int
+                The cell index of the cell in that voxel
+
+    """
+    cell_number_tag = mesh.mesh.getTagHandle('cell_number')
+    subvoxel_array = np.zeros(0, dtype=[(b'svid', np.int64),
+                                                      (b'idx', np.int64),
```
gonuke: too much indentation here?

pyne/alara.py
```diff
@@ -774,3 +806,43 @@ def _output_flux(ve, tag_flux,output,start,stop,direction):
...
+                                                      (b'scid', np.int64)])
+    temp_subvoxel = np.zeros(1, dtype=[(b'svid', np.int64),
+                                                      (b'idx', np.int64),
+                                                      (b'scid', np.int64)])
+    # calculate the total number of non-void sub-voxel
+    non_void_sv_num = 0
+    for i, _, ve in mesh:
+        for c, cell in enumerate(cell_number_tag[ve]):
+            if cell > 0:
+                if len(cell_mats[cell].comp):
```
gonuke: This is really a test for void, right?  Perhaps we should add a method to the PyNE material object called `nonvoid()`?

pyne/alara.py
```diff
@@ -774,3 +806,43 @@ def _output_flux(ve, tag_flux,output,start,stop,direction):
...
+            :svid: int
+                The index of non-void subvoxel id
+            :idx: int
+                The idx of the voxel
+            :scid: int
+                The cell index of the cell in that voxel
+
+    """
+    cell_number_tag = mesh.mesh.getTagHandle('cell_number')
+    subvoxel_array = np.zeros(0, dtype=[(b'svid', np.int64),
```
gonuke, line 829: What does `shape=0` mean here?  Does it just set up the structure but create an array with no entries?

pyne/alara.py
```diff
@@ -230,13 +248,27 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags):
...
+            for i in range(num_vol_elements):
+                temp_mesh_data[i] = [0.0] * num_e_groups * max_num_cells
+            for sve, row in enumerate(subvoxel_array):
+                    # start index of data
+                    d_s = row['scid'] * num_e_groups
+                    # end index of data
+                    d_e = (row['scid'] + 1) * num_e_groups
+                    temp_mesh_data[row['idx'],d_s:d_e] = matched_data[sve][3][:]
+            for i, _, ve in mesh:
+                tag_handles[tags[cond]][ve] = temp_mesh_data[i,:]
```
gonuke: If you used a 3-D array, you could probably `reshape` it for this step.

gonuke (changes requested):
A few comments here about the shape of your data structure.  This works as is, but it might be more maintainable with a 3-D array...

pyne/alara.py
zxkjack123: When we create this as `np.empty()`, it actually contains numbers like  `-2.41598391e-039,   6.91406097e-310,   6.91404841e-310,`, and there is some number like `-1.85792002e+163`, this will make unexpected error. So, I think set them all to be `0.0` is not a bad choice. 

Also, I will use `numpy.fill` command to replace the loop.

pyne/alara.py
zxkjack123: Yes, this line just create an array frame without entries. Because how many entries are there before the loop, so I create a frame and fill in entries in the loop.

pyne/alara.py
zxkjack123: Yes, that's a test for void. I agree with you that we should add a method for PyNE material. But PyNE material is wrote in Cython, there are several files need to be changed if we add a method. So, I will add that in a separate PR.

zxkjack123:
Made some changes according to the suggestion of @gonuke . 
Major changes:
1. Add a news file
2. Some clean up
3. For the subvoxel tag data part, use 3D array to replace 2D array
4. Use `subvoxel` as loop variable to replace `row` in a loop
5. Rebased on current develop

pyne/alara.py
gonuke: right - maybe it should be `np.zeros` instead?

gonuke (approved):

gonuke:
Thanks @zxkjack123

Issue #919, opened 2017-10-04
PyNE CPP no longer compiles with Clang


makeclean:
On my Mac the CPP part of PyNE no longer compiles, entirely due to cram.c, the error is

```In file included from /Users/davisa/opt/pyne/src/transmuters.cpp:2:
In file included from /Users/davisa/opt/pyne/src/cram.hpp:9:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/complex.h:29:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/ccomplex:21:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/complex:244:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/type_traits:387:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/cstddef:44:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/__nullptr:35:5: error: templates must
      have C++ linkage
    template <class _Tp>
    ^~~~~~~~~~~~~~~~~~~~```

Of course one can use brew to use and install gcc, just an FYI

scopatz:
Hi @makeclean - did you do a fully clean build (ie remove the build dir) when you switched compilers?  That is what I suspect this error is from, since the error is coming from deep within system headers.  Note that cram.hpp does not contain any templated code.

scopatz:
Thanks for reporting!

makeclean:
Yes, from a clean build, here is the full trace

```[  0%] Built target ReplicatePythonSourceTree
[  4%] Building CXX object src/CMakeFiles/pyne.dir/data.cpp.o
[  4%] Building CXX object src/CMakeFiles/pyne.dir/atomic_data.cpp.o
[  4%] Building C object src/CMakeFiles/pyne.dir/cram.c.o
[  5%] Building CXX object src/CMakeFiles/pyne.dir/enrichment.cpp.o
[  6%] Building CXX object src/CMakeFiles/pyne.dir/decay.cpp.o
[...]

PR #944, opened 2017-11-19
removed extern C from source sampling

scopatz:
CC @ljacobson64 @makeclean @FlanFlanagan May address #919

### DIFF
--- a/src/source_sampling.h
+++ b/src/source_sampling.h
@@ -2,25 +2,25 @@
 /// \author Elliott Biondo (biondo\@wisc.edu)
 ///
 /// \brief Mesh-based Monte Carlo source sampling.
-/// 
+///
 /// The Sampler class is used for Monte Carlo source sampling from mesh-based
 /// sources.  The source density distribution and optional biased source density
-/// distribution are defined on a MOAB mesh. Upon instantiation, a Sampler  
+/// distribution are defined on a MOAB mesh. Upon instantiation, a Sampler
 /// object reads this mesh and creates an alias table for randomly sampling
-/// particle birth parameters. The particle_birth member function is supplied 
-/// with 6 pseudo-random numbers and returns the position, energy, and weight 
-/// of a particle upon birth. 
+/// particle birth parameters. The particle_birth member function is supplied
+/// with 6 pseudo-random numbers and returns the position, energy, and weight
+/// of a particle upon birth.
 /// There are three sampling modes: analog, uniform, and user-speficied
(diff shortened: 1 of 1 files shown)


scopatz:
OK this passes test on linux, but testing on Mac with clang would be highly desirable.

makeclean:
IIRC we need that when linking those routines into a fortran based code - eg MCNP

scopatz:
Hmm, well this should have never worked for that :).  This is clearly C++ code, not C code meant for interacting with fortran.  I think another interface layer is needed, if MCNP linkage is needed.

scopatz:
Also, are you all still experiencing issues on Mac?

gonuke:
Hmm... yes - I think these extern's should be in the `cpp` file... not the `hpp` file - and only around certain functions...

gonuke:
hmmm 2.0 ... no - it should be in the hpp, but perhaps more localized to only the functions that will be in the interface for C/FORTRAN

scopatz:
2.0 makes sense to me

scopatz:
If someone wants to put in an alternative fix, that would be awesome!

gonuke:
Perhaps @zxkjack123 can take a look at this since he is working in this part of the code anyway.

It would be useful to add a small FORTRAN test - a short F90 file that calls each of the methods in the interface to demonstrate that they are, in fact, callable.

elliottbiondo:
@gonuke see:

http://pyne.io/usersguide/source_sampling.html#fortran-interface

elliottbiondo:
I can't speak to whether the "extern C" is necessary but I will say that the version of this file prior to this PR did work from Fortran and MCNP without any additional interface layer :)

gonuke:
Not sure I understand exactly what you mean by your comment @elliottbiondo .... 

I don't think we need any additional layer, but that we can/should more carefully localize where the `extern C`.  

Most importantly, I think we need a test for this interface to make sure it doesn't get broken by proposed changes like this.

gonuke:
... and you have pointed us to a good test in the documentation

gonuke:
Uggh.. but we have no infrastructure for tests that required compiled code...

elliottbiondo:
My interface comment was in response to something way earlier in this thread. Not important.

gonuke:
Various users have successfully used this without this change.  The methods are called directly from FORTRAN90, for example, within MCNP and require these delimiters to make the methods available across this language interface.

PR #945, opened 2017-11-19
actually make particle fille return a void*

scopatz:
Here is another error I noticed in relation to #919, CC @makeclean @ljacobson64 @FlanFlanagan


ljacobson64:
Looks good

scopatz:
Thanks @ljacobson64!

PR #959, opened 2017-11-24
CRAM update, makes solver and compiling faster

scopatz:
Updates for new version of transmutagen.


scopatz:
this is superseded by #963

PR #961, opened 2017-11-27
convert half-life in energy width to time in seconds using formula from NuData 2 glossary

pkienzle:


pyne/utils.py
```diff
@@ -52,12 +52,21 @@ class QAWarning(UserWarning):
...
 
 
+# Energy to half-life conversion:  T1/2= ln(2) × (h/2 pi) / energy
+# See http://www.nndc.bnl.gov/nudat2/help/glossary.jsp#halflife
+# NIST CODATA https://physics.nist.gov/cgi-bin/cuu/Value?hbar
+#    h-bar = 1.054 571 800(13) x 1e-34 J 
+#    1 J = 6.241 509 126(38) x 1e18 eV
+HBAR_LN2 = 4.5623775832376968e-16  # h-bar ln(2) in eV s
+energy_conv_dict = {'ev': HBAR_LN2,
+                    'kev': 1e-3 * HBAR_LN2,
```
scopatz: I don't know that this is quite correct.  I don't think that this should have the `ln(2)` per se.  That is only relevant for converting half-lives, but there are other things measured in MeV that could be converted to time.

pyne/utils.py
```diff
@@ -74,7 +83,11 @@ def to_sec(input_time, units):
         Time value in [sec].
 
     """
-    sec_time = input_time * time_conv_dict[units.lower()]
+    units = units.lower()
+    try:
+        sec_time = time_conv_dict[units] * input_time
+    except KeyError:
```
scopatz: Exceptions should only be used for exceptional cases.  Please check containment with `key in dict` or use `dict.get()`

scopatz:
Thanks for putting this in @pkienzle!  Other than the issues I commented on inline, I am not sure that this is really the right way to tackle the problem - and what is there is wrong too! It doesn't make sense to convert something from eV to sec. Some unit conversion is already needed to get it in an acceptable form.  Maybe there needs to be better documentation or a different function entirely for converting half-life widths to seconds? Or the call sites should be changed.

Also, I think this probably needs tests before going in too

pkienzle:
There is a further problem:  in pyne.ensdf when converting uncertainty on time reported as energy, you can't convert the value, plus and minus independently.  

Instead you need something like:
```
converted = scale / value
converted_plus = scale / (value - minus) - converted
converted_minus = converted - scale / (value + plus)
```

So you will need to adjust your internal interfaces appropriately to convert half-life and uncertainty correctly.

scopatz:
Ahh interesting.  All help here is appreciated!

pkienzle:
A quick grep shows that ```to_sec``` is only used in ```ensdf```.  Probably better off just moving the conversion code there where it knows that time is for half-life.

scopatz:
> A quick grep shows that to_sec is only used in ensdf. Probably better off just moving the conversion code there where it knows that time is for half-life.

Yeah, having an ensdf time converter is a reasonable solution :)

pkienzle:
I added some tests, including time as energy with uncertainty.

pyne/utils.py
```diff
@@ -11,72 +11,6 @@
 class QAWarning(UserWarning):
     pass
 
-time_conv_dict = {'as': 1e-18,
```
scopatz: I think that this dict (but with out the energy keys) and the original `to_sec()` fucntion should be restored.  There is no reason to break API here, and  I think that this function may be used by some down stream packages.

tests/test_utils.py
```diff
@@ -12,10 +12,6 @@
 import numpy as np
 
 
-def test_to_sec():
```
scopatz: Please restore this test function as well

pyne/ensdf.py
```diff
@@ -68,6 +67,106 @@ def _to_id(nuc):
         return 0
     return nucid
 
+time_conv_dict = {'as': 1e-18,
```
scopatz: This can probably be removed and then imported from utils.  No reason  to have the data structure twice :)

pyne/ensdf.py
```diff
@@ -68,6 +67,106 @@ def _to_id(nuc):
...
+#    h-bar = 1.054 571 800(13) x 1e-34 J
+#    1 J = 6.241 509 126(38) x 1e18 eV
+HBAR_LN2 = 4.5623775832376968e-16  # h-bar ln(2) in eV s
+energy_conv_dict = {'ev': HBAR_LN2,
+                    'kev': 1e-3 * HBAR_LN2,
+                    'mev': 1e-6 * HBAR_LN2,
+                    }
+
+
+def _to_sec(value, err, units):
```
scopatz: Also, I may recommend a name change here (`_to_sec_with_err()`?) since the interface no longer matches the other tosec function and it also converts energy half-lives.

scopatz:
Thanks for the update! More comments inline!

pkienzle:
I've restored the utils.to_sec function (without the eV -> s conversion) as you requested.

pyne/ensdf.py
```diff
@@ -69,20 +69,72 @@ def _to_id(nuc):
...
+    Returns
+    -------
+    sec_time : float
+        Time value in [sec].
+    sec_err : None or float or (float, float) in [sec].
+        Time uncertainty in [sec], or (plus, minus) if asymmetric uncertainty.
+    """
+    if err is None:
+        plus, minus = 0, 0
+    elif isinstance(err, float):
```
scopatz: This might fail in unexpected ways.  For example, `isinstance(np.float32(10.0), float)` is False.  It seems like it might be better to test if it is a Sequence.  

```python
from collections import Sequence

...

elif isinstance(err, Sequence):
   plus, minus = err
else:
    plus = minus = err
```

scopatz:
Other than the one isinstance check, this looks good to me!

scopatz:
pinging this @pkienzle, it'd be great to get this in.

pyne/ensdf.py
pkienzle: I disagree, since this is an internal function called by the parser which converts error strings to None, float, or (float, float).   Yes, if called directly by user code it is quite possible that an int or a numpy scalar or even a mpmath 500 digit constant, but it won't be in this code.  Slowly introducing PEP484 and mypy typing is a better solution since you only pay for it at build-time, not at run-time.

That said, I will put in a new commit with a np.isscalar() test.

scopatz:
Thanks!

PR #963, opened 2017-11-28
Fast Compile

scopatz:
This makes compilation of PyNE faster by compiling from pre-built assembly, when available (currently only for GCC/GNU). This takes compile times for me locally down from 120 sec for decay.cpp to 3 sec for decay-gnu.s. Don't worry, the decay.cpp and cram.c files are still available as a fall back or if `--slow` is passed in via the command like. 

I also anticipate that this will help with building packages and CI.


scopatz:
ping @pyne/developers

setup.py
```diff
@@ -136,15 +136,37 @@ def download_decay():
     try:
         durl = urlopen(DECAY_URL)
         d = durl.read()
+    except IOError:
+        print('...failed!')
```
katyhuff: Perhaps a more informative error message here?

katyhuff (approved):
I didn't try the fast compile on my machine, but code lgtm. I left one minor comment about an error message.

setup.py
scopatz: Done!

FlanFlanagan:
Much faster! Going to merge this in.

PR #965, opened 2017-11-30
Tag cell fracs update numpyshape

zxkjack123:
There was a bug in `pyne/mesh.py:tag_cell_fracs` . The error message is: `ValueError: could not broadcast input array from shape (N,1) into shape (N)`  where `N` is the number of volume elements. 

The bug appears when the `max_num_cells == 1`. If the `max_num_cells` is greater than 1, then there is no problem. 

So, I modified the code to differentiate two conditions of `max_num_cells` (it's 1 or greater than 1), and add some code to treat the condition when `max_num_cells == 1`. 

Major changes:
1. In `pyne/mesh.py`: add some code for the condition when `max_num_cells == 1`

2. In `tests/test_mesh.py`: add a test function `test_tag_cell_fracs_subvoxel_equal_voxel` to test this condition

### DIFF
--- a/pyne/mesh.py
+++ b/pyne/mesh.py
@@ -1226,17 +1226,18 @@ class Mesh(object):
         max_num_cells = -1
         for i in range(num_vol_elements):
             max_num_cells = max(max_num_cells,
-                                  len(cell_fracs[cell_fracs['idx'] == i]))
+                                len(cell_fracs[cell_fracs['idx'] == i]))
 
-        # creat tag frame with default value
+        # create tag frame with default value
         cell_largest_frac_number = [-1] * num_vol_elements
         cell_largest_frac = [0.0] * num_vol_elements
-        voxel_cell_number = np.empty(shape=(num_vol_elements,max_num_cells),
+        voxel_cell_number = np.empty(shape=(num_vol_elements, max_num_cells),
                                      dtype=int)
-        voxel_cell_fracs = np.empty(shape=(num_vol_elements,max_num_cells),
+        voxel_cell_fracs = np.empty(shape=(num_vol_elements, max_num_cells),
                                     dtype=float)
         voxel_cell_number.fill(-1)
         voxel_cell_fracs.fill(0.0)
+
         # set the data
         for i in range(num_vol_elements):
             for (cell, row) in enumerate(cell_fracs[cell_fracs['idx'] == i]):
@@ -1245,11 +1246,16 @@ class Mesh(object):
             # cell_largest_frac_tag
             cell_largest_frac[i] = max(voxel_cell_fracs[i, :])
             largest_index = \
-                list(voxel_cell_fracs[i, :]).index(cell_largest_frac[i])
+                    list(voxel_cell_fracs[i, :]).index(cell_largest_frac[i])
             cell_largest_frac_number[i] = \
-                int(voxel_cell_number[i, largest_index])
+                    int(voxel_cell_number[i, largest_index])
 
-        # creat the tags
+        # reshape the array
+        if max_num_cells == 1:
+            voxel_cell_number.shape = (num_vol_elements, )
+            voxel_cell_fracs.shape = (num_vol_elements, )
+
+        # create the tags
         self.tag(name='cell_number', value=voxel_cell_number,
                  doc='cell numbers of the voxel, -1 used to fill vacancy',
                  tagtype=IMeshTag, size=max_num_cells, dtype=int)
--- a/tests/test_mesh.py
+++ b/tests/test_mesh.py
@@ -814,6 +814,34 @@ def test_tag_cell_fracs():
         assert_equal(m.cell_largest_frac[i],
                            exp_cell_largest_frac[i])
 
+def test_tag_cell_fracs_subvoxel_equal_voxel():
+    m = Mesh(structured=True,
+             structured_coords=[[0, 0.5, 1], [0, 0.5, 1], [0, 0.5, 1]],
+             mats = None)
+    cell_fracs = np.zeros(8, dtype=[('idx', np.int64),
+                                    ('cell', np.int64),
+                                    ('vol_frac', np.float64),
+                                    ('rel_error', np.float64)])
+
+    cell_fracs[:] = [(0, 1, 1.0, 0.0), (1, 2, 1.0, 0.0), (2, 3, 1.0, 0.0),
+                     (3, 4, 1.0, 0.0), (4, 5, 1.0, 0.0), (5, 6, 1.0, 0.0),
+                     (6, 7, 1.0, 0.0), (7, 8, 1.0, 0.0)]
+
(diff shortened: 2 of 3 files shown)


zxkjack123:
When add data to tags of a mesh with `N` voxels, we usually set the data as a single `int` or `float` if the  size is 1 for each voxel (all the data can be stored in a list with length of `N`), and set the data as a array if the size is bigger than 1 (all the data can be stored in a np.ndarry with shape of `(N, max_num_cells)`). 

For the `cell_number` tag and the `cell_fracs` tag of mesh, the data size for each voxel can both equal to 1 or greater than 1. 

I want to set the tag as a 1D array with length of 1 if the tag size of each voxel equals to 1, but not as a single `int` or `float` data. Because this will make the code more clear and the `cell_number` get consistent data structure. 

However, the `IMeshTag` will broadcast the array of shape `(N, 1)` to shape `(N)`. @gonuke @pshriwise Is there any way to avoid that broadcast?

pyne/mesh.py
```diff
@@ -1231,23 +1231,35 @@ def tag_cell_fracs(self, cell_fracs):
         # creat tag frame with default value
         cell_largest_frac_number = [-1] * num_vol_elements
         cell_largest_frac = [0.0] * num_vol_elements
-        voxel_cell_number = np.empty(shape=(num_vol_elements,max_num_cells),
-                                     dtype=int)
-        voxel_cell_fracs = np.empty(shape=(num_vol_elements,max_num_cells),
-                                    dtype=float)
+        if max_num_cells != 1:
+            voxel_cell_number = np.empty(shape=(num_vol_elements,max_num_cells),
```
scopatz: PEP8 single space after comma

pyne/mesh.py
```diff
@@ -1231,23 +1231,35 @@ def tag_cell_fracs(self, cell_fracs):
         # creat tag frame with default value
         cell_largest_frac_number = [-1] * num_vol_elements
         cell_largest_frac = [0.0] * num_vol_elements
-        voxel_cell_number = np.empty(shape=(num_vol_elements,max_num_cells),
-                                     dtype=int)
-        voxel_cell_fracs = np.empty(shape=(num_vol_elements,max_num_cells),
-                                    dtype=float)
+        if max_num_cells != 1:
```
scopatz: It is usually easier to read if conditionals are positive.  So I would have swapped the two cases and written `if max_num_cells == 1`, but that is not a super blocker here.

scopatz:
Code more or less looks fine to me.  Just a couple of small cleanups.

zxkjack123:
@scopatz Thanks for your comments. I made cleanup to the codes as your suggestion.

pyne/mesh.py
```diff
@@ -1231,24 +1231,35 @@ def tag_cell_fracs(self, cell_fracs):
         # creat tag frame with default value
         cell_largest_frac_number = [-1] * num_vol_elements
         cell_largest_frac = [0.0] * num_vol_elements
-        voxel_cell_number = np.empty(shape=(num_vol_elements,max_num_cells),
-                                     dtype=int)
-        voxel_cell_fracs = np.empty(shape=(num_vol_elements,max_num_cells),
-                                    dtype=float)
+        if max_num_cells == 1:
```
pshriwise: I'm not totally clear on why thy array needs to be one dimensional if the number of cells is 1, but I'm wondering if [this post](https://stackoverflow.com/questions/42882842/whats-the-difference-between-n-and-n-1-in-numpy) might help.

Mostly I'm thinking these can be treated as two dimensional arrays until the very end of the function when the mesh.tag function is called. Right before this they can be reshaped. This would reduce the lines of code above that considerably and keep the indexing the same for most of the function. 

So something like: 

```
# no changes in code up to here

# line 1263
#create the tags
if max_num_cells == 1: 
    voxel_cell_numer.reshape((num_vol_elements,))
    voxel_cell_fracs.reshape((num_vol_elements,))

# previous tag calls as they were before
  self.tag(name='cell_number', value=voxel_cell_number,
                   doc='cell numbers of the voxel, -1 used to fill vacancy',
                  tagtype=IMeshTag, size=max_num_cells, dtype=int)
self.tag(name='cell_fracs', value=voxel_cell_fracs,
                  tagtype=IMeshTag, doc='volume fractions of each cell in the '
                                        'voxel, 0.0 used to fill vacancy',
                  size=max_num_cells, dtype=float)

```
(Sorry, indentation isn't great. In a hurry)

pyne/mesh.py
```diff
@@ -1231,24 +1231,35 @@ def tag_cell_fracs(self, cell_fracs):
...
+            if max_num_cells == 1:
+                cell_largest_frac_number[i] = voxel_cell_number[i]
+                cell_largest_frac[i] = voxel_cell_fracs[i]
+            else:
+                cell_largest_frac[i] = max(voxel_cell_fracs[i, :])
+                largest_index = \
+                    list(voxel_cell_fracs[i, :]).index(cell_largest_frac[i])
+                cell_largest_frac_number[i] = \
+                    int(voxel_cell_number[i, largest_index])
         # creat the tags
```
pshriwise: creat -> create

tests/test_mesh.py
```diff
@@ -814,6 +814,34 @@ def test_tag_cell_fracs():
         assert_equal(m.cell_largest_frac[i],
                            exp_cell_largest_frac[i])
 
+def test_tag_cell_fracs_subvoxel_equal_voxel():
```
pshriwise, line 817: Test looks good!

pshriwise (commented):
Comment on how to reduce changes in code, spelling error fix

zxkjack123:
Codes are simplified according to the comments of @pshriwise .

scopatz:
@pshriwise  - please feel free to merge if it looks good to you!

zxkjack123:
Cleaning up the code. Commits form others are removed.

scopatz:
Also, please add a news entry to this. We don't have documentation for this yet, but it is effectively the same as following the instructions here: http://xon.sh/devguide.html#changelog

gonuke:
LGTM

gonuke (approved):

gonuke:
@scopatz - for news, is there any guidance about 1-to-1 mapping of news entries to PRs.

scopatz:
Hi @gonuke - The rule (for xonsh at least) is that every PR must create or modify a news file.  We recommend that the news file be the same as the branch name, so usually this means there is a 1-to-1 mapping from PR to news file.  However, sometimes branch names are reused or a later PR comes in and modifies what an earlier PR did, so only minor changes to an existing news file are needed.

zxkjack123:
@scopatz A news file was add according to your suggestion.

pshriwise (approved):
LGTM @zxkjack123!

gonuke:
Thanks @zxkjack123

Issue #967, opened 2017-12-03
Suggestion on IMeshTag about (N, 1) shaped array as input value


zxkjack123:
Assuming that we have a mesh with `N` volume elements. Now we want to add an `IMeshTag` to the mesh. 
If the data size for each volume element is greater than 1 (here notate it as `M`), we set the tag from a two dimensional np.ndarray of shape `(N, M)`. If the data size for each volume elements is 1, we usually set the tag from a one dimensional np.ndarry of shape `(N, )`. These two conditions are easy to understand and works fine in PyNE now. 

However, there is another condition that I don't know the exactly data size for each volume element in previous, i.e., the `M` could be 1, 2, 3, ..., etc.. Though the M is greater than 1 in most cases, but it still could be 1. For the purpose of keep the tag data in the same structure, I want to set the tag data from two dimensional np.ndarry of shape `(N, 1)` to the mesh if the `M == 1`. But this will cause error in PyNE: `ValueError: could not broadcast input array from shape (N,1) into shape (N)`

The reason of this error is that, when we add a tag using mesh.tag or IMeshTag with given `value` (a `(N, M)` shaped array, it will call the `__setitem__` method of `class IMeshTag`:    https://github.com/pyne/pyne/blob/4d402fe70ed156d1b34c7ec48dcbb76cc536b2b4/pyne/mesh.py#L396

In this method, `v[...]` is used to receive data from `value`. However, when the `M == 1`, this `v[...]` is a one dimensional array according to below line:
https://github.com/pyne/pyne/blob/4d402fe70ed156d1b34c7ec48dcbb76cc536b2b4/pyne/mesh.py#L415

Thus, the `v[...]` is a one dimensional array of shape `(N, )` while the `value` is a two dimensional array with shape `(N, 1)`. And the error "ValueError: could not broadcast input array from shape (N,1) into shape (N)" jumps out. 
 
I suggest that when we add a tag and set the value as a `(N, 1)` array (tag size is 1), the default `IMeshTag` data for each volume element should be an array but not a number. 
For example, replace the code:
https://github.com/pyne/pyne/blob/4d402fe70ed156d1b34c7ec48dcbb76cc536b2b4/pyne/mesh.py#L415

``` 
v = np.empty(len(key), self.tag.type) if tsize == 1 else \
                np.empty((len(key), tsize), self.tag.type)
```
with the following:
```
if tsize == 1 and len(value.shape) == 1:
    v = np.empty(len(key), self.tag.type)
else:
[...]

PR #969, opened 2017-12-04
IMeshTag about (N, 1) shaped array

zxkjack123:
This pull request contains the modification described in issue #967 . 

Major change:

1. When adding a tag from a (N, 1) array (tag size is 1), the default IMeshTag data for each volume element is set to be an array but not a number.


pshriwise:
I'm curious about what this means for #965, is it still necessary after these changes?

zxkjack123:
@pshriwise Actually, if this pull request merged into PyNE, #965 is no longer necessary. 

Keeping #965 won't make error, but I will modify #965 to delete unnecessary code if this pull request merged first.

pshriwise:
@zxkjack123 good. I think making the change here is best if it doesn't cause tests to fail. @scopatz should probably take a quick look at this as well to make sure this is going into the right location.

That sounds good regarding #965. The formatting improvements and additional test should be kept I think.

zxkjack123:
@pshriwise As you can see, I didn't make changes in test files in this pull request. But I tested this change with the condition of add a tag with array (N, 1). This test function is in #965 . I didn't add that test function here to avoid mess up the pull requests sequences.

pshriwise:
This is what I was referring to. The test in PR #965 :smile: I think we're on the same page.

PR #971, opened 2017-12-05
IMeshTag about (N, 1) shaped array

zxkjack123:
Closed #969 by mistake. This is the identical one as that.
Rebased on develop.

### DIFF
--- a/pyne/mesh.py
+++ b/pyne/mesh.py
@@ -415,10 +415,6 @@ class IMeshTag(Tag):
             v = np.empty((len(key), tsize), self.tag.type)
             if tsize == 1 and len(value.shape) == 1:
                 v.shape = (len(key), )
-            #if tsize == 1 and len(value.shape) == 1:
-            #    v = np.empty(len(key), self.tag.type)
-            #else:
-            #    v = np.empty((len(key), tsize), self.tag.type)
             v[...] = value
             mtag[key] = v
         elif isinstance(key, np.ndarray) and key.dtype == np.bool:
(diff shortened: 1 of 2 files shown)


scopatz (approved):

scopatz:
Also, please add a news entry to this. We don't have documentation for this yet, but it is effectively the same as following the instructions here: http://xon.sh/devguide.html#changelog

pshriwise (approved):
This looks good. Thanks @zxkjack123

pyne/mesh.py
```diff
@@ -412,17 +412,21 @@ def __setitem__(self, key, value):
             mtag[i_ve[1]] = value if tsize == 1 else value[0]
         elif isinstance(key, slice):
             key = list(miter)[key]
-            v = np.empty(len(key), self.tag.type) if tsize == 1 else \
-                np.empty((len(key), tsize), self.tag.type)
+            if tsize == 1 and len(value.shape) == 1:
```
gonuke: Could you use the same trick as in #965 and just reshape this if the size is 1?
```
v = np.empty((len(key), tsize), self.tag.type)
if  tsize == 1 and len(value.shape) == 1:
    v.shape = (len(key), )
```

gonuke (commented):
Looking for less code duplication and consistency with other places where we do this.

pyne/mesh.py
zxkjack123: Codes are changed here according to your suggestion. @gonuke

zxkjack123:
@scopatz A news was added to this PR.

gonuke:
Thanks @zxkjack123

PR #973, opened 2017-12-11
Remove unused code and add news file

zxkjack123:
Removed some unused code in mesh.py. 
These code are introduced in PR #965 for the problem of IMeshTag shape. But the problem of IMeshTag shape is fixed in PR #971 , so this part is no longer needed.


scopatz (approved):

pshriwise (approved):
Looks good!

gonuke:
LGTM  - thanks @zxkjack123

PR #974, opened 2017-12-11
Subvoxel source subroutine analog only

zxkjack123:
This PR added some code to implement the sub-voxel r2s - analog mode. 
Major  changes:
1. Functions and variables definition in `source_sampling.h`, `source_sampling.cpp`, `source_sampling.pyx` and `cpp_source_sampling.pxd`. 
2. Add a return value:`cell_number` for `particle_birth` function. 
3. Add three test function for sub-voxel r2s - analog mode

### DIFF
--- a/src/source_sampling.cpp
+++ b/src/source_sampling.cpp
@@ -32,7 +32,8 @@ void pyne::particle_birth_(double* rands,
                            double* y,
                            double* z,
                            double* e,
-                           double* w) {
+                           double* w,
+                           double* c) {
     std::vector<double> rands2(rands, rands + 6);
     std::vector<double> samp = sampler->particle_birth(rands2);
     *x = samp[0];
@@ -40,6 +41,7 @@ void pyne::particle_birth_(double* rands,
     *z = samp[2];
     *e = samp[3];
     *w = samp[4];
+    *c = samp[5];
 }
 
 std::vector<double> pyne::read_e_bounds(std::string e_bounds_file){
@@ -91,8 +93,18 @@ std::vector<double> pyne::Sampler::particle_birth(std::vector<double> rands) {
   // select mesh volume and energy group
   //
   int pdf_idx =at->sample_pdf(rands[0], rands[1]);
-  int ve_idx = pdf_idx/num_e_groups;
-  int e_idx = pdf_idx % num_e_groups;
+  int ve_idx;
+  int c_idx;
+  int e_idx;
+  if (mode == USER || mode == UNIFORM || mode == ANALOG) {
+    ve_idx = pdf_idx/num_e_groups;
+    e_idx = pdf_idx % num_e_groups;
+  }
+  if (mode == SUBVOXEL_USER || mode == SUBVOXEL_UNIFORM || mode == SUBVOXEL_ANALOG) {
+    ve_idx = -1;
+    e_idx = -1;
+    c_idx = -1;
+  }
 
   // Sample uniformly within the selected mesh volume element and energy
   // group.
@@ -107,6 +119,12 @@ std::vector<double> pyne::Sampler::particle_birth(std::vector<double> rands) {
   samp.push_back(pos[2]);
   samp.push_back(sample_e(e_idx, rands[5]));
   samp.push_back(sample_w(pdf_idx));
+  if (mode == USER || mode == UNIFORM || mode == ANALOG) {
+    samp.push_back(-1.0);
+  }
+  if (mode == SUBVOXEL_USER || mode == SUBVOXEL_UNIFORM || mode == SUBVOXEL_ANALOG) {
+    samp.push_back(1.0);
+  }
   return samp;
 }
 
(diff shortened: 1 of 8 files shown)


src/source_sampling.cpp
```diff
@@ -75,8 +92,9 @@ pyne::Sampler::Sampler(std::string filename,
 std::vector<double> pyne::Sampler::particle_birth(std::vector<double> rands) {
   // select mesh volume and energy group
   //
-  int pdf_idx =at->sample_pdf(rands[0], rands[1]);
-  int ve_idx = pdf_idx/num_e_groups;
+  int pdf_idx = at->sample_pdf(rands[0], rands[1]);
+  int ve_idx = pdf_idx/num_e_groups/max_num_cells;
```
pshriwise: I trust that this is correct, but would you mind explaining what the additional division by `max_num_cells ` in these two variables is for? I'm guessing this is to get an index from a flat array of information?

src/source_sampling.h
```diff
@@ -88,24 +89,24 @@ namespace pyne {
     std::vector<double> prob; /// Probabilities.
     std::vector<int> alias; /// Alias probabilities.
   };
-  
+
   /// Problem modes
-  enum Mode {USER, ANALOG, UNIFORM};
-  
+  enum Mode {USER, ANALOG, UNIFORM, SUBVOXEL_ANALOG, SUBVOXEL_UNIFORM, SUBVOXEL_USER};
```
pshriwise, line 94: The naming of this enumerator concerns me a bit. Is there a way around declaring 3 additional Modes? I'm thinking of a scheme in which you first decide if you're going to be doing subvoxel sampling and then determine if the sampling routine will be analog, uniform, or user-defined. What this might look like is a separate enumerator: 

```
enum Mode {USER, ANALOG, UNIFORM};
enum subMode{DEFAULT, SUBVOXEL};
```
Or something like that...
Then the if statement checks for the mode could go from:

```
  if (mode == USER || mode == UNIFORM || mode == ANALOG) {
    samp.push_back(-1.0);
  }
  if (mode == SUBVOXEL_USER || mode == SUBVOXEL_UNIFORM || mode == SUBVOXEL_ANALOG) {
    samp.push_back(double(cell_number[ve_idx*max_num_cells + c_idx]));
  }
```

could go to 


```
  if ( subMode == SUBVOXEL ) {
      samp.push_back(double(cell_number[ve_idx*max_num_cells + c_idx]));
    }
  else {
      samp.push_back(-1.0);
    }
```

pshriwise (commented):
Kudos for diving into learning how to update the cython files! And the majority of the changes to the code look good to me - just a comment on how the subvoxel mode is checked in various places within the code.

I do wonder, however, if the implementation of the subvoxel sampling is different enough from the existing mode that a new, derived class for subvoxel sampling might be worthwhile at some point in the future.

src/source_sampling.h
pshriwise: This would likely simplify other, similar checks in other places in the code as well.

zxkjack123: Thank your for the suggestion @pshriwise. I agree with you. Another variable SubMode is a good way to simplify the code and make the code easier to understand. I applied this change in the latest commit.

src/source_sampling.cpp
zxkjack123: As you know, when we don't apply sub-voxel r2s, the photon source of each voxel contains `num_e_groups` numbers, so the size of `pdf` is `num_ves*num_e_groups`. In this case, the `ve_idx` is `pdf_idx/num_e_groups`. 

Well, when we apply sub-voxel r2s, the photon source of each voxel contains `max_num_cells*num_e_groups` number, so the size of `pdf` is `num_ves*max_num_cells*num_e_groups`. These data are stored in a `vector`, with the `energy` changes fastest. So the `ve_idx` is `pdf_idx/max_num_cells/num_e_groups`.

src/source_sampling.h
pshriwise: Yes, the changes much cleaner to me now. I didn't put too much thought into the name of that enumerator when writing that example, so you may want to reconsider it's name. Would subvoxel sampling vs. default sampling be a "major" or "minor" mode to you? 

I guess I think of it as there now being two major modes:
  1. Subvoxel Sampling with ANALOG, UNIFORM, and USER options
and 
  2. Default sampling with ANALOG, UNIFORM and USER options

But maybe it's really the other way around. I don't know. :smile:

zxkjack123: I do have another option about the mode names. These modes are actually in two different level, i.e. there are two decisions to determine the mode. First, use sub-voxel r2s or not, we define this using `SubMode = {DEFAULT, SUBVOXEL}`. Second, how to bias the photon source? We define this using `BiasMode = {ANALOG, UNIFORM, USER}`. (This `BiasMode` is the original `Mode`). 

By replacing the original `Mode` with `BiasMode`, and add the parameter `SubMode`, it will be easier to understand the decision process. 

By the way, I think `SubMode = {DEFAULT, SUBVOXEL}` is a good name, easy to understand.

zxkjack123:
Content of the PR was merged into PR #975 .

src/source_sampling.h
pshriwise: Would `SampleMode` not be a little more clear if the other can be named `BiasMode`?

PR #975, opened 2017-12-13
Subvoxel source subroutine sampling

zxkjack123:
PR #974 contains only the subvoxel-analog mode. However, the analog, uniform and user define modes are highly related with each other with `if - else` clauses. 

It's better to put them in one PR. So I closed #974  and put these three (analog, uniform and user define) subvoxel sampling method together in this PR. 

Major changes:
1. Add functions and variable for subvoxel source sampling. 
2. Add test functions to these three sampling mode.

### DIFF
--- a/pyne/source_sampling.pyx
+++ b/pyne/source_sampling.pyx
@@ -213,6 +213,7 @@ cdef class Sampler:
     biased_weights (std::vector< double >) : Birth weights for
         biased sampling.
     at (None) : Alias table used for sampling.
+    source_particle (None) : Source Particle used to store sampling result
     
     
     Methods
@@ -229,6 +230,13 @@ cdef class Sampler:
     sample_w
     sample_xyz
     setup
+    get_src_x
+    get_src_y
+    get_src_z
+    get_src_e
+    get_src_w
+    get_src_c
+    get_src_xyzew
     
     Notes
     -----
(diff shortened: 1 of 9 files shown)


zxkjack123:
Add a class `SourceParticle` to store the sampling results. And add some function to get source particle information to avoid the type conversion of the `cell_number`.

pyne/source_sampling.pyx
```diff
@@ -478,14 +652,185 @@ cdef class Sampler:
...
         rtnval_proxy = np.PyArray_Copy(rtnval_proxy)
         return rtnval_proxy
     
+
+cdef class SourceParticle:
+    """Constructor for class SourceParticle
+    
+    Attributes
+    ----------
+    zzz (double) : The z coordinate of the source particle
```
pshriwise: Are there some attributes missing here perhaps?

I'm also wondering about the naming of these... why `zzz`? Rather than simply `z`?

src/source_sampling.cpp
```diff
@@ -28,14 +37,23 @@ void pyne::particle_birth_(double* rands,
...
     std::vector<double> rands2(rands, rands + 6);
-    std::vector<double> samp = sampler->particle_birth(rands2);
-    *x = samp[0];
-    *y = samp[1];
-    *z = samp[2];
-    *e = samp[3];
-    *w = samp[4];
+//    std::vector<double> samp = sampler->particle_birth(rands2);
+    sampler->particle_birth(rands2);
+    //*x = samp[0];
```
pshriwise: These commented lines can probably be removed (here and in other areas of the code)

src/source_sampling.cpp
```diff
@@ -394,3 +590,47 @@ int pyne::AliasTable::sample_pdf(double rand1, double rand2) {
   int i = (int) n * rand1;
   return rand2 < prob[i] ? i : alias[i];
 }
+
+pyne::SourceParticle::SourceParticle(double x, double y, double z,
```
pshriwise: Nice work with the addition of this class! I think this benefits not only your changes but the entire source sampling routine.

I do wonder if maybe the variable names could be more simple here. Rather than `xxx`, `yyy`, `zzz` why not `x`, `y`, `z`? Maybe there is a convention here I'm not aware of...

pshriwise (commented):
Looks great @zxkjack123. Just a few questions/suggestions.

src/source_sampling.cpp
zxkjack123: `x`, `y`, `z` are used as the input parameters of the constructor of the class `SourceParticle` and also as the input parameters of `particle_birth`. If the `x`, `y`, `z` are then used as the names of the attributes of the source particle, the `SourceParticle` constructor will be something like below:
```
pyne::SourceParticle::SourceParticle(double x, double y, double z,
                                     double e, double w, int c) {
    x = x;
    y = y;
    z = z;
    e = e;
    w = w;
    c = c;
}
```
In other words, names of the attributes and the input arguments will be the same. It may lead to confusion of the attributes and arguments. Therefore, I think it's better to use different names to differentiate the attributes from the arguments for better readability. 
So another set of variable names `xxx`, `yyy`, `zzz`, `erg`, `wgt` are used. They are the variable names used in `MCNP`. 

I think there maybe other method to avoid conflicts between attributes and parameters. For exampler, change  the parameter names. However, that still introduced another set of names.

pshriwise: This is true. And I think it is important to differentiate the argument variables from the attributes of the class. The arguments to the constructor are temporary, however, and won't be seen even when calling the class constructor. The attributes of the class are more permanent and in my opinion should probably have priority in for a clear naming scheme over the constructor arguments.

I'm thinking that even something like this should suffice: 

```
pyne::SourceParticle::SourceParticle(double _x, double _y, double _z,
                                     double _e, double _w, int _c) {
    x = _x;
    y = _y;
    z = _z;
    e = _e;
    w = _w;
    c = _c;
}
```

Separate from this suggestion is the naming of methods on the class. Even though the attribute may be named `xxx` or `x` the method `get_source_xxx` can still be `get_src_x` which makes sense regardless of whether or not you know the MCNP conventions for these variable names.

These are small things, but they may become important in the long run :)

zxkjack123: I found that I can use `this` to differentiate the attributes from the parameters. So the following code may be a better way.  
```
pyne::SourceParticle::SourceParticle(double x, double y, double z,
                                     double e, double w, int c) {
    this->x = x;
    this->y = y;
    this->z = z;
    this->e = e;
    this->w = w;
    this->c = c;
}
```
And I changed the name of `xxx` to `x`, changed the name of `get_src_xxx` to `get_src_x` according to your suggestion.

src/source_sampling.h
```diff
@@ -88,24 +92,53 @@ namespace pyne {
     std::vector<double> prob; /// Probabilities.
     std::vector<int> alias; /// Alias probabilities.
   };
-  
+
+  // class Source particle
+  class SourceParticle {
```
gonuke, line 97: Maybe this should be a struct??  We should check the style guide in case it cares?

src/source_sampling.h
```diff
@@ -118,30 +151,82 @@ namespace pyne {
...
+            std::string cell_fracs_tag_name,
+            std::vector<double> e_bounds,
             std::string bias_tag_name);
+
     /// Samples particle birth parameters
     /// \param rands Six pseudo-random numbers in range [0, 1].
     /// \return A vector containing the x position, y, position, z, point, energy
     ///         and weight of a particle (in that order).
-    std::vector<double> particle_birth(std::vector<double> rands);
+    void particle_birth(std::vector<double> rands);
```
gonuke: return a copy of to the `source_particle`

gonuke: won't need access methods below

src/source_sampling.h
gonuke: Maybe not - instead you need a method `std::vector<double> pack()` that replaces `Sampler::get_src_xyzew()`

src/source_sampling.h
```diff
@@ -150,8 +235,11 @@ namespace pyne {
     // sampling
     std::vector<edge_points> all_edge_points; ///< Four connected points on a VE.
     std::vector<double> biased_weights; ///< Birth weights for biased sampling.
+    std::vector<int> cell_number; ///< Tag cell_number
+    std::vector<double> cell_fracs; ///< Tag cell_fracs
     AliasTable* at; ///< Alias table used for sampling.
-  
+    SourceParticle* source_particle; ///< source particle generated
```
gonuke: There should be no need for a member variable anymore.

src/source_sampling.cpp
```diff
@@ -87,12 +132,18 @@ std::vector<double> pyne::Sampler::particle_birth(std::vector<double> rands) {
...
   samp.push_back(sample_e(e_idx, rands[5]));
   samp.push_back(sample_w(pdf_idx));
-  return samp;
+  if (sub_mode == SUBVOXEL) {
+    samp.push_back(double(cell_number[ve_idx*max_num_cells + c_idx]));
+  } else {
+    samp.push_back(-1.0);
+  }
+  //return samp;
+  source_particle = new SourceParticle(samp[0], samp[1], samp[2], samp[3], samp[4], int(samp[5]));
```
gonuke: Don't allocate a new SourceParticle, just return a SourceParticle object

src/source_sampling.cpp
```diff
@@ -28,14 +37,16 @@ void pyne::particle_birth_(double* rands,
...
+                           double* w,
+                           int* c) {
     std::vector<double> rands2(rands, rands + 6);
-    std::vector<double> samp = sampler->particle_birth(rands2);
-    *x = samp[0];
-    *y = samp[1];
-    *z = samp[2];
-    *e = samp[3];
-    *w = samp[4];
+    sampler->particle_birth(rands2);
```
gonuke: this changes if/when the return type of `particle_birth` changes

src/source_sampling.h
```diff
@@ -118,30 +151,82 @@ namespace pyne {
     ///                       number of energy groups as <src_tag_name> or 1.
     ///                       If 1 (i.e. spatial biasing only), all energy groups
     ///                       within a mesh volume element are sampled equally.
-    Sampler(std::string filename, 
-            std::string src_tag_name, 
-            std::vector<double> e_bounds, 
+    Sampler(std::string filename,
```
gonuke, line 155: instead of overloading the constuctor, pass the `mode` to the constructor and have it setup correctly.

src/source_sampling.cpp
```diff
@@ -10,6 +10,8 @@ void pyne::sampling_setup_(int* mode) {
   if (sampler == NULL) {
     std::string filename ("source.h5m");
     std::string src_tag_name ("source_density");
+    std::string cell_num_tag_name ("cell_num");
```
gonuke, line 13: These are fixed/hard coded, so make part of constructor instead.

src/source_sampling.h
zxkjack123: These constructors are not only used by DAG-MCNP, but also by python when performing tests. 
From DAG-MCNP, we can only pass an integer parameter `mode` into the code, all the other file names are hard coded.
But for the PyNE unit tests, we need to change these file names, otherwise we can't make different unit tests to read different test input files. 
I think we should keep these overloading constructors for the purpose of PyNE test usage

src/source_sampling.cpp
zxkjack123: These constructors are not only used by DAG-MCNP, but also by python when performing tests. 
However, we can't pass `string` type from DAG-MCNP input. But we need to define these `string` type file names for PyNE unit tests. 
So we have to hard code these strings between the DAG-MCNP calling function `pyne::sampling_setup_(int* mode)`  and the specific sampler for a given condition. Thus we can define these strings in PyNE unit tests.

src/source_sampling.h
pshriwise: This is true for the `filename` argument of the constructor, but any of the `*_tag_name` will always be the same because they are hardcoded in the lines above these constructors. To simplify the code here, there can be one constructor that looks something like: 

```
Sampler(std::string filename,
              std::vector<double> e_bounds,
              int mode,
              int sub_mode);
```

or something along those lines.

That's my understanding anyways, but maybe I'm missing something.

zxkjack123:
Update:
1. Rebase on the latest `develop` branch.
2. Return a `SourceParticle` object from `particle_birth`.

src/source_sampling.cpp
gonuke: Now I understand... these are hard-coded only for the FORTRAN use case.  

It is possible to pass strings from FORTRAN.  Since this is called by the `source.F90` function that is also part of this pull request, maybe we should take this opportunity to move the hard-coded strings to there.

They can remain hard-coded because we don't (yet) have a way to get strings from the DAG-MCNP input file.

gonuke:
As I look at this more deeply, I think there are a few design issues we should discuss.

Most importantly, I think there is a way to structure the SUBVOXEL mode so that it is more integrated with the "NORMAL" mode.  There are many places with nested loops that you have copied and pasted and added an additional loop over the `max_num_cells`.  In most (all?) cases, the indexing for the SUBVOXEL case will resolve to be correct for the NORMAL case if `max_num_cells` = 1, so the same loops can be used in both cases.

This was a big effort and very well done, but now that we see how it all comes together, we might have an opportunity to simplify much of it.

src/source_sampling.h
pshriwise: Meant to send the comment above out yesterday. Sorry for the delay.

zxkjack123: That's true. We have 6 constructors now, they are in four types of parameter sets. Actually, what we really need to construct is the `filename`, `e_bounds` and made`. `sub_mode` can be derived from `mode`.   
```
Sampler(std::string filename,
              std::vector e_bounds,
              int mode);
```
That means, we need only one constructor. However, this design will absolutely introduce big changes to the `source_sampleing` and replace the original valid constructors for normal r2s. I don't think we should make these changes before we make sure that `sub_voxel` code runs as we expect.

zxkjack123:
@gonuke Yes, there are many loop copied and pasted, there are many duplicate codes.
The 'SUB_VOXEL` mode was designed to behave the same way as the `DEFAULT` mode when `max_num_cells`=1. We can of course use these codes for both `DEFAUTL` and `SUB_VOXEL` mode. 
These duplicate codes were added to keep the original NORMAL r2s code untouched, prevent me from ruining the original r2s. As I haven't fully valid the `SUB_VOXEL` codes.
I agree that we can and should design a way to simplify these codes. And I will apply the simplification when the validation of `SUB_VOXEL` mode finished.

gonuke:
Perhaps this can move forward if we separate some of the changes into separate PRs:
1. introduce `SourceParticle` return type instead of `std:vector<double>`
2. introduce map of tag names (without subvoxel types for now)
3. change variable names on loops over volume elements and energy groups, and change `mode` to `bias_mode`

zxkjack123:
This PR will be split into several smaller PRs.

scopatz:
> This PR will be split into several smaller PRs.

Always a good idea :)

PR #976, opened 2018-01-08
 Added details about decay heat function units

gwenchee:
The decay heat function units is in MW/material object total mass. These lines were added for the sake of clarity for future users.

### DIFF
--- a/pyne/material.pyx
+++ b/pyne/material.pyx
@@ -608,7 +608,9 @@ cdef class _Material:
 
 
     def decay_heat(self):
-        """This provides the decay heat using the comp of the the Material.
+        """This provides the decay heat using the comp of the the Material. It
+        assumes that the composition of material is given in units of [grams]
+        and returns decay heat in units of [MW/material object total mass].  
 
         Returns
         -------
--- a/pyne/material.pyx
+++ b/pyne/material.pyx
@@ -610,7 +610,7 @@ cdef class _Material:
     def decay_heat(self):
         """This provides the decay heat using the comp of the the Material. It
         assumes that the composition of material is given in units of [grams]
-        and returns decay heat in units of [MW/material object total mass].  
+        and returns decay heat in units of [MW].  
 
         Returns
         -------


pyne/material.pyx
```diff
@@ -608,7 +608,9 @@ cdef class _Material:
 
 
     def decay_heat(self):
-        """This provides the decay heat using the comp of the the Material.
+        """This provides the decay heat using the comp of the the Material. It
+        assumes that the composition of material is given in units of [grams]
+        and returns decay heat in units of [MW/material object total mass].
```
scopatz: I think the decay heat is returned in  units of MW: https://github.com/pyne/pyne/blob/develop/src/material.h#L237

scopatz:
Thanks for putting this in @gwenchee!

pyne/material.pyx
gwenchee: I think line 237 is referring to the activity function, https://github.com/pyne/pyne/blob/develop/src/material.h#L239 says that the decay heat function returns dose per gram, which is synonymous with [MW/material object total mass in grams]. I will correct my PR to include grams.

scopatz: Hi @gwenchee - doxygen/javadoc style docstrings are placed above the function they are documenting, not below.  So 237 is the correct line.  Also, if you see 239 has units of mrem/h, which is more appropriate for dose, than decay heat.

katyhuff:
@gwenchee this looks stalled... I think a pr update was among next steps?

pyne/material.pyx
gwenchee: Yup that makes sense. I changed it to reflect that the unit is [MW]

gwenchee:
I updated my commit to reflect Anthony's comments. It should be ready for merge now.

scopatz:
Hi @gwenchee can your merge from develop please to get the latest commits so that CI will pass

gwenchee:
done! @scopatz

scopatz:
Thanks @gwenchee!

PR #977, opened 2018-01-10
Fast compile is now platform specific

scopatz:
This brings it inline with what was implemented for cyclus/cyclus#1454.

CC @FlanFlanagan @pyne/core

### DIFF
--- a/cmake/PyneMacros.cmake
+++ b/cmake/PyneMacros.cmake
@@ -220,7 +220,7 @@ macro(fast_compile _srcname _gnuflags _clangflags _otherflags)
     message(STATUS "Compiling ${_srcname} fast from assembly ${_asmname}")
     set(_filename "${_asmname}")
   endif()
-  set(PYNE_SRCS "${PYNE_SRCS}" "${_filename}")
+  set(PYNE_SRCS "${_filename}" "${PYNE_SRCS}")
 
   # set some compile flags for the selected file
   if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
--- a/cmake/PyneMacros.cmake
+++ b/cmake/PyneMacros.cmake
@@ -14,6 +14,28 @@ macro(pyne_set_platform)
   endif(APPLE)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D${PYNE_PLATFORM}")
   message("-- Pyne platform defined as: ${PYNE_PLATFORM}")
+
+  # sets a unique platform string
+  if(NOT DEFINED PYNE_PLATFORM_STRING)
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -134,50 +157,60 @@ def ssl_context():
...
+DECAY_H_REP = os.path.join('src', '_decay.h')
+DECAY_CPP_REP = os.path.join('src', '_decay.cpp')
+DECAY_URL = 'http://raw.githubusercontent.com/pyne/data/master/decay.tar.gz'
+
+
+def download_decay():
+    return download_platform(DATA_URL, 'decay', 'src')
+
+
 CRAM_H = os.path.join('src', 'decay.h')
```
gonuke, line 207: These need to be corrected to `cram.[hc]`

setup.py
```diff
@@ -134,50 +157,60 @@ def ssl_context():
...
+    if not status:
+        return False
+    return True
+
+
+DATA_URL = 'http://raw.githubusercontent.com/pyne/data/master'
+DECAY_H = os.path.join('src', 'decay.h')
+DECAY_CPP = os.path.join('src', 'decay.cpp')
+DECAY_H_REP = os.path.join('src', '_decay.h')
+DECAY_CPP_REP = os.path.join('src', '_decay.cpp')
```
gonuke, line 199: should add assembly code that might be downloaded, for cleanup?

gonuke:
Just found this PR which was already here when we worked on #996 - our bad!

There are w few things that aren't covered her, however, particularly with respect to cleaning up the fast compile code in `--clean` mode.

setup.py
gonuke: In general, I expect the set of `cram` files to be similar to the set of `decay` files.

setup.py
gonuke: in #996 we do this with a glob since we don't know what compiler it will be a priori

cmake/PyneMacros.cmake
```diff
@@ -201,7 +220,7 @@ macro(fast_compile _srcname _gnuflags _clangflags _otherflags)
     message(STATUS "Compiling ${_srcname} fast from assembly ${_asmname}")
     set(_filename "${_asmname}")
   endif()
-  set(PYNE_SRCS "${PYNE_SRCS}" "${_filename}")
+  set(PYNE_SRCS "${_filename}" "${PYNE_SRCS}")
```
scopatz, line 223: @gonuke @bam241, This line change is actually kinda of nice to have so that the ASM stuff tries to get built first. Also, in the event of a source compile, it ensures that the long compile starts first, rather than last, for max parallelism.

gonuke:
closing this since we went ahead with #996

PR #981, opened 2018-01-19
WIP: Windows support

ghost:
Note: commits related to resolving flang compiler errors should be removed before merging.


ghost:
@scopatz The main blocker here appears to be that C complex numbers are not supported on Windows.

scopatz:
Hi @xoviat - only the cram files use C complexes.  We can skip those on windows if they are truly unsupported.

scopatz:
Working on a clean up PR of this now, since this breaks on linux.

ghost:
Note: source changes (present here) are required to the Fortran files to work around Flang compiler errors. I have not resolved the compiler errors yet due to CLA issues.

ghost:
C99 complex numbers are unsupported on Windows, but C++ complex numbers are supported. Perhaps if cram was rewritten in C++, these issues would be resolved.

scopatz:
The problem is that C++ complex numbers are slow, which is why cram was written in C99. I think for the time being it would be easiest to just not build cram on Windows. We can think about generating a version in C++ in the future, but it is a pretty low priority for me.

ghost:
Why are C++ complex numbers slow?

ghost:
I haven't benchmarked myself, but I was under the impression that modern compilers could generate optimum code.

scopatz:
At least GCC has special optimization flags (such as `-ftree-sra`) for dealing with small, repeated structs (which is what a complex number is in C99). In C++ they are a template class and the compiler does not pick up on the same optimizations.  

These optimizations are really important. Cram running with clang (which doesn't have these optimizations) is about 4 - 8 times slower than GCC.

scopatz:
Maybe it is more fair to say that GCC with C99 or Fortran has really fast complex numbers that aren't available with other languages or compilers.

scopatz:
Ok, I have some fixes for the windows branch.  Do you want me to PR it into your fork @xoviat or should I PR it here?

ghost:
Just create a new PR.

scopatz:
See #1001

scopatz:
BTW I really appreciate all of your help and effort here!

PR #995, opened 2018-02-16
R2s change loop variables name

zxkjack123:
Major changes:
1. Change loop variables. Use `v` replace `i`, use `e` replace `j`.
2. Rename `mode` into `bias_mode`

### DIFF
--- a/news/r2s_change_loop_variables_name.rst
+++ b/news/r2s_change_loop_variables_name.rst
@@ -1,8 +1,8 @@
 **Added:** None
 
 **Changed:** 
-* Use v (ves) and e(energy) to replace i, j
-* Use bias_mode to replace mode
+* Change loop variables to be v for volume elements and e for energy groups (instead of i & j)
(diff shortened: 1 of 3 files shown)


news/r2s_change_loop_variables_name.rst
```diff
@@ -0,0 +1,13 @@
+**Added:** None
+
+**Changed:** 
+* Use v (ves) and e(energy) to replace i, j
```
gonuke: Perhaps add a little more context here, e.g.
* Change loop variables to be `v` for volume elements and `e` for energy groups (instead of i & j)
* Use `bias_mode` instead of `mode` to allow for additional mode types in future

gonuke:
This looks good.  Just one brief comment about the level of detail in the change log

gonuke:
FYI - I'm trying to study why the tests are failing, before merging this...

gonuke:
Thanks @zxkjack123

PR #996, opened 2018-02-16
Fix CI

bam241:
thx to @CNERG @gonuke @pshriwise @kkiesling @ngranda @zxkjack123 @ypark234

We fixed it.
CI should work again

this download platform specific assembly for fast compile.
we moved the assembly download part in CMake.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,10 @@ import numpy as np
 if '.' not in sys.path:
     sys.path.append(os.getcwd() + '/src')
 
-absexpanduser = lambda x: os.path.abspath(os.path.expanduser(x))
+
+def absexpanduser(x): return os.path.abspath(os.path.expanduser(x))
+
+
 VERSION = '0.5.10'
 IS_NT = os.name == 'nt'
 LOCALDIR = absexpanduser('~/.local')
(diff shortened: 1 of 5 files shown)


bam241:
Could you check it @scopatz ?

gonuke:
It turns out this is basically redundant with #977

bam241:
shall I just close this one ?

setup.py
```diff
@@ -57,7 +57,10 @@
 if '.' not in sys.path:
     sys.path.append(os.getcwd() + '/src')
 
-absexpanduser = lambda x: os.path.abspath(os.path.expanduser(x))
+
+def absexpanduser(x): return os.path.abspath(os.path.expanduser(x))
```
scopatz, line 61: I am not sure, but I think it is against PEP8 to put functions on one line like this.

bam241: my autopep8 did this one..

seems to match this from https://www.python.org/dev/peps/pep-0008/ :
```
Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier.

Yes:

def f(x): return 2*x
No:

f = lambda x: 2*x
```

scopatz: Ahh that is fair :)

scopatz:
Other than the order of the compile issue from #977, this looks good to me!

gonuke:
Thanks for the review @scopatz

PR #998, opened 2018-02-19
R2s fix wrong mode doc

zxkjack123:
Major change:
1. Correct the wrong mode description comment in `pyne/src/source_sampling.h`.


gonuke:
Nice catch! Thanks @zxkjack123

Issue #994, opened 2018-02-15
Remove hard-coded strings from C/FORTRAN API to source_sampling


gonuke:
The current interface hard-codes a number of strings (e.g. filenames and tag names) for `source_sampling` when called from FORTRAN.  It should be straightforward to push this up to the calling code to make PyNE's FORTRAN interface more flexible.

gonuke:
This was marked as `low hanging` because there exists a good model for this in DAGMC's wrappers for MCNP: https://github.com/svalinn/DAGMC/blob/develop/mcnp/mcnp_funcs.cpp#L50

PR #997, opened 2018-02-19
R2s return source particle

zxkjack123:
Major change:
1. Add class `SourceParticle` in `pyne/src/source_sampling`.  
2. Change the return value of `Sampler::particle_birth` from `vector<double>` to `SourceParticle` object.

PR #999, opened 2018-02-22
R2s introduce map of tag names

zxkjack123:
This PR is used to introduce a new constructor for `pyne::Sampler`. 
This new constructor uses a map of <string, string> to store hard coded string in the source file, enables users to define the strings easier. 
This PR partly solves the problem addressed in issue: #994 .
However, the original constructors are currently remained in the code. I think it's better to remove these previous constructors when this PR merged into develop. 

Major changes:
1. Add a constructor  for `pyne::Sampler` in `source_sampling`. 
2. Modify tests functions to use the new added constructor.

### DIFF
--- a/pyne/cpp_source_sampling.pxd
+++ b/pyne/cpp_source_sampling.pxd
@@ -41,7 +41,6 @@ cdef extern from "source_sampling.h" namespace "pyne":
         # attributes
 
         # methods
-        cpp_vector[double] get_xyzew() except +
         double get_x() except +
         double get_y() except +
         double get_z() except +
--- a/pyne/cpp_source_sampling.pxd
+++ b/pyne/cpp_source_sampling.pxd
@@ -60,6 +60,7 @@ cdef extern from "source_sampling.h" namespace "pyne":
         Sampler(std_string, std_string, cpp_vector[double]) except +
         Sampler(std_string, std_string, cpp_vector[double], std_string) except +
         Sampler(std_string, std_string, cpp_vector[double], cpp_bool) except +
+        Sampler(cpp_map, cpp_vector, int) except +
 
         # attributes
 
--- a/pyne/cpp_source_sampling.pxd
+++ b/pyne/cpp_source_sampling.pxd
@@ -60,7 +60,7 @@ cdef extern from "source_sampling.h" namespace "pyne":
         Sampler(std_string, std_string, cpp_vector[double]) except +
         Sampler(std_string, std_string, cpp_vector[double], std_string) except +
         Sampler(std_string, std_string, cpp_vector[double], cpp_bool) except +
-        Sampler(cpp_map[std_string, std_string], cpp_vector[double], int) except +
+        Sampler(std_string, cpp_map[std_string, std_string], cpp_vector[double], int) except +
 
         # attributes
 
--- a/pyne/cpp_source_sampling.pxd
+++ b/pyne/cpp_source_sampling.pxd
@@ -11,6 +11,7 @@
 from libcpp cimport bool as cpp_bool
 from libcpp.string cimport string as std_string
 from libcpp.vector cimport vector as cpp_vector
+from libcpp.map cimport map as cpp_map
 
 cdef extern from "source_sampling.h" namespace "pyne":
 
--- a/pyne/cpp_source_sampling.pxd
+++ b/pyne/cpp_source_sampling.pxd
@@ -60,7 +60,7 @@ cdef extern from "source_sampling.h" namespace "pyne":
         Sampler(std_string, std_string, cpp_vector[double]) except +
         Sampler(std_string, std_string, cpp_vector[double], std_string) except +
         Sampler(std_string, std_string, cpp_vector[double], cpp_bool) except +
-        Sampler(cpp_map, cpp_vector, int) except +
+        Sampler(cpp_map[std_string, std_string], cpp_vector[double], int) except +
 
         # attributes
 
--- a/pyne/source_sampling.pyx
+++ b/pyne/source_sampling.pyx
@@ -189,13 +189,13 @@ cdef class AliasTable:
     pass
 
 
-cdef cpp_vector[double] convert_e_bounds(e_bounds):
-    """convert_e_bounds(rands)
+cdef cpp_vector[double] convert_nparray_to_vector(array):
+    """convert_nparray_to_vector
     Convert python np.ndarray into cpp_vector[double]
     
     Parameters
     ----------
-    e_bounds : np.ndarray
+    array : np.ndarray
     
     Returns
     -------
@@ -203,16 +203,15 @@ cdef cpp_vector[double] convert_e_bounds(e_bounds):
     
     """
 
-    cdef cpp_vector[double] e_bounds_proxy
-    cdef int ie_bounds
-    cdef int e_bounds_size
-    cdef double * e_bounds_data
-    e_bounds = np.array(e_bounds, dtype=np.float64)
-    e_bounds_size = len(e_bounds)
-    e_bounds_data = <double *> np.PyArray_DATA(<np.ndarray> e_bounds)
-    e_bounds_proxy = cpp_vector[double](<size_t> e_bounds_size)
-    memcpy(<void*> &e_bounds_proxy[0], e_bounds_data, sizeof(double) *  e_bounds_size)
-    return e_bounds_proxy
+    cdef cpp_vector[double] array_proxy
+    cdef int array_size
+    cdef double * array_data
+    array = np.array(array, dtype=np.float64)
+    array_size = len(array)
+    array_data = <double *> np.PyArray_DATA(<np.ndarray> array)
+    array_proxy = cpp_vector[double](<size_t> array_size)
+    memcpy(<void*> &array_proxy[0], array_data, sizeof(double) *  array_size)
+    return array_proxy
 
 
 
@@ -302,7 +301,7 @@ cdef class Sampler:
         src_tag_name_bytes = src_tag_name.encode()
         bias_tag_name_bytes = bias_tag_name.encode()
         # convert e_bounds
-        cdef cpp_vector[double] e_bounds_proxy = convert_e_bounds(e_bounds)
+        cdef cpp_vector[double] e_bounds_proxy = convert_nparray_to_vector(e_bounds)
         # construct sampler
         self._inst = new cpp_source_sampling.Sampler(
                 std_string(<char *> filename_bytes),
@@ -339,7 +338,7 @@ cdef class Sampler:
         filename_bytes = filename.encode()
         src_tag_name_bytes = src_tag_name.encode()
         # convert e_bounds
-        cdef cpp_vector[double] e_bounds_proxy = convert_e_bounds(e_bounds)
+        cdef cpp_vector[double] e_bounds_proxy = convert_nparray_to_vector(e_bounds)
         # construct sampler
         self._inst = new cpp_source_sampling.Sampler(
                 std_string(<char *> filename_bytes),
@@ -373,7 +372,7 @@ cdef class Sampler:
         for key, value in tag_names.items():
             cpp_tag_names[key] = value
         # convert e_bounds
-        cdef cpp_vector[double] e_bounds_proxy = convert_e_bounds(e_bounds)
+        cdef cpp_vector[double] e_bounds_proxy = convert_nparray_to_vector(e_bounds)
         # construct sampler
         self._inst = new cpp_source_sampling.Sampler(
                 std_string(<char *> filename_bytes),
--- a/pyne/source_sampling.pyx
+++ b/pyne/source_sampling.pyx
@@ -374,7 +374,7 @@ cdef class Sampler:
                 e_bounds_proxy[ie_bounds] = <double> e_bounds[ie_bounds]
         self._inst = new cpp_source_sampling.Sampler(std_string(<char *> filename_bytes), std_string(<char *> src_tag_name_bytes), e_bounds_proxy, <bint> uniform)
 
-    def _sampler_sampler_5(self, names, e_bounds, mode):
+    def _sampler_sampler_2(self, names, e_bounds, mode):
         """Sampler(self, names, e_bounds, mode)
         
         Constuctor for overall Sampler
@@ -429,6 +429,7 @@ cdef class Sampler:
     
     _sampler_sampler_0_argtypes = frozenset(((0, str), (1, str), (2, np.ndarray), (3, str), ("filename", str), ("src_tag_name", str), ("e_bounds", np.ndarray), ("bias_tag_name", str)))
     _sampler_sampler_1_argtypes = frozenset(((0, str), (1, str), (2, np.ndarray), (3, bool), ("filename", str), ("src_tag_name", str), ("e_bounds", np.ndarray), ("uniform", bool)))
+    _sampler_sampler_2_argtypes = frozenset(((0, dict), (1, np.ndarray), (3, int), ("names", dict), ("e_bounds",  np.ndarray), ("mode", int)))
     
     def __init__(self, *args, **kwargs):
         """Sampler(self, filename, src_tag_name, e_bounds, uniform)
@@ -480,6 +481,9 @@ cdef class Sampler:
         if types <= self._sampler_sampler_1_argtypes:
             self._sampler_sampler_1(*args, **kwargs)
             return
+        if types <= self._sampler_sampler_2_argtypes:
+            self._sampler_sampler_2(*args, **kwargs)
+            return
         # duck-typed dispatch based on whatever works!
         try:
             self._sampler_sampler_0(*args, **kwargs)
@@ -492,6 +496,11 @@ cdef class Sampler:
         except (RuntimeError, TypeError, NameError):
             pass
         raise RuntimeError('method __init__() could not be dispatched')
+        try:
+            self._sampler_sampler_2(*args, **kwargs)
+            return
+        except (RuntimeError, TypeError, NameError):
+            pass
     
     def __dealloc__(self):
         if self._free_inst and self._inst is not NULL:
--- a/pyne/source_sampling.pyx
+++ b/pyne/source_sampling.pyx
@@ -191,7 +191,7 @@ cdef class AliasTable:
 
 def convert_e_bounds(e_bounds):
     """convert_e_bounds(rands)
-    Convert python np.ndarray into vector< double> 
+    Convert python np.ndarray into cpp_vector[double]
     
     Parameters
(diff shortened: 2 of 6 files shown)


zxkjack123:
This PR changes the method that `PyNE` tests call for the Sampler constructors. However, this PR does not change the interface for DAG-MCNP because currently we can't pass strings from DAG-MCNP via `source.F90`. So, there are still some hard coded string in `source_sampling.cpp` before calling the Sampler constructors.

pyne/cpp_source_sampling.pxd
```diff
@@ -54,4 +56,4 @@ cdef extern from "source_sampling.h" namespace "pyne":
 
 
 
-{'cpppxd_footer': '', 'pyx_header': '', 'pxd_header': '', 'pxd_footer': '', 'cpppxd_header': '', 'pyx_footer': ''}
\ No newline at end of file
+{'cpppxd_footer': '', 'pyx_header': '', 'pxd_header': '', 'pxd_footer': '', 'cpppxd_header': '', 'pyx_footer': ''}
```
scopatz: Please remove this line

pyne/source_sampling.pyx
```diff
@@ -372,10 +373,63 @@ cdef class Sampler:
...
+        ----------
+        e_bounds : std::vector< double >
+        
+        names : std::map<std::string, std::string>
+        
+        mode : int
+        
+        Returns
+        -------
+        None
```
scopatz: I don't think that you can have two parameters and returns sections in numpydoc

pyne/source_sampling.pyx
```diff
@@ -372,10 +373,63 @@ cdef class Sampler:
...
+        if isinstance(e_bounds, np.ndarray) and (<np.ndarray> e_bounds).descr.type_num == np.NPY_FLOAT64:
+            e_bounds_data = <double *> np.PyArray_DATA(<np.ndarray> e_bounds)
+            e_bounds_proxy = cpp_vector[double](<size_t> e_bounds_size)
+            for ie_bounds in range(e_bounds_size):
+                e_bounds_proxy[ie_bounds] = e_bounds_data[ie_bounds]
+        else:
+            e_bounds_proxy = cpp_vector[double](<size_t> e_bounds_size)
+            for ie_bounds in range(e_bounds_size):
+                e_bounds_proxy[ie_bounds] = <double> e_bounds[ie_bounds]
+        self._inst = new cpp_source_sampling.Sampler(<cpp_map[std_string, std_string]> cpp_names, <cpp_vector[double]> e_bounds_proxy, <int> mode)
```
scopatz: This line is too long

pyne/source_sampling.pyx
```diff
@@ -427,6 +481,9 @@ cdef class Sampler:
         if types <= self._sampler_sampler_1_argtypes:
             self._sampler_sampler_1(*args, **kwargs)
             return
+        if types <= self._sampler_sampler_2_argtypes:
```
scopatz: Shouldn't this be an `elif`?

pyne/source_sampling.pyx
```diff
@@ -372,10 +373,63 @@ cdef class Sampler:
...
+        cdef cpp_map[std_string, std_string] cpp_names = cpp_map[std_string, std_string]()
+        for key, value in names.items():
+            cpp_names[key] = value
+
+        # e_bounds is a ('vector', 'float64', 0)
+        e_bounds_size = len(e_bounds)
+        if isinstance(e_bounds, np.ndarray) and (<np.ndarray> e_bounds).descr.type_num == np.NPY_FLOAT64:
+            e_bounds_data = <double *> np.PyArray_DATA(<np.ndarray> e_bounds)
+            e_bounds_proxy = cpp_vector[double](<size_t> e_bounds_size)
+            for ie_bounds in range(e_bounds_size):
```
scopatz: There are faster ways to copy array data into a vector than iterating over every element. In fact you have already done the hard part by getting the pointer in `e_bounds_data`. In fact I think that you can do:

```cython
e_bounds_proxy = cpp_vector[double](e_bounds_data, e_bounds_data + <size_t> e_bounds_size)
```

pyne/source_sampling.pyx
```diff
@@ -372,10 +373,63 @@ cdef class Sampler:
...
+                e_bounds_proxy[ie_bounds] = e_bounds_data[ie_bounds]
+        else:
+            e_bounds_proxy = cpp_vector[double](<size_t> e_bounds_size)
+            for ie_bounds in range(e_bounds_size):
+                e_bounds_proxy[ie_bounds] = <double> e_bounds[ie_bounds]
```
scopatz: Same here, no need for the loop

scopatz:
Thanks for putting this in!

pyne/source_sampling.pyx
gonuke: Is this a style question or a functional question?  The `return` in the previous block makes this effectively and `elif` anyway...

src/source_sampling.cpp
```diff
@@ -72,6 +72,26 @@ pyne::Sampler::Sampler(std::string filename,
   setup();
 }
 
+pyne::Sampler::Sampler(std::map<std::string, std::string> names,
```
gonuke: I think it will be best to keep the `filename` separate from the `names` map

pyne/cpp_source_sampling.pxd
zxkjack123: This line is removed in #997 .

pyne/source_sampling.pyx
zxkjack123: There will be an error like this if I use the method you suggested:
```
Error compiling Cython file:
------------------------------------------------------------
...
        if isinstance(e_bounds, np.ndarray) and (<np.ndarray> e_bounds).descr.type_num == np.NPY_FLOAT64:
            e_bounds_data = <double *> np.PyArray_DATA(<np.ndarray> e_bounds)
            e_bounds_proxy = cpp_vector[double](<size_t> e_bounds_size)
            #for ie_bounds in range(e_bounds_size):
            #    e_bounds_proxy[ie_bounds] = e_bounds_data[ie_bounds]
            e_bounds_proxy = cpp_vector[double](e_bounds_data, e_bounds_data + <size_t> e_bounds_size)
                                                            ^
------------------------------------------------------------

/home/zxk/opt/pyne/pyne/source_sampling.pyx:370:61: Cannot assign type 'double *' to 'size_t'
```

pyne/cpp_source_sampling.pxd
scopatz: hmm then you shouldn't modify it here to prevent conflicts.

pyne/source_sampling.pyx
scopatz: I suppose it is a style issue.  If something is logically an elif, it seems like it should be represented as an elif.  There is also probably some tribal performance difference since using an elif is one if-block where as here it is two if-blocks. This is non-blocking to me, just recommended.

pyne/source_sampling.pyx
scopatz: Sorry minor modification needed:

```cython
e_bounds_proxy = cpp_vector[double](<size_t> e_bounds_size)
memcpy(<void*> &e_bounds_proxy[0], e_bounds_data, sizeof(double) * e_bounds_size)
```

or similar. You'll also need to add the following to the top of the file.

```cython
from libc.string cimport memcpy
```

zxkjack123: Thanks for the suggestion. I made the `e_bounds` conversion part as a separate function to reduce duplications.

zxkjack123:
Update: 
1. Make `e_bounds` conversion part as a function to reduce duplication in the code. 
2. Make `filename` a separate input parameter. 
3. Change two `if` to `elif`.

pyne/source_sampling.pyx
```diff
@@ -490,6 +484,20 @@ cdef class Sampler:
     pass
 
 
+def convert_e_bounds(e_bounds):
+    """
+    """
```
scopatz: Function needs a real docstring :)

scopatz: Also this is a Python function (since it isn't there is no return type) so I think that this implictly converts the vector into a Python list on return.

zxkjack123: Sorry, I forgot it.

scopatz: No problem, that is what review is for :smile:

zxkjack123:
Updated some docstrings and did some clean up.

pyne/source_sampling.pyx
```diff
@@ -187,6 +189,31 @@ cdef class AliasTable:
...
+    """convert_e_bounds(rands)
+    Convert python np.ndarray into vector< double> 
+    
+    Parameters
+    ----------
+    e_bounds : np.ndarray
+    
+    Returns
+    -------
+    res1 : std::vector< double >
```
scopatz: This does not return a std::vector because it is a Python function.  Cython is automatically converting the vector into a Python list on return.  You need to make to make this a cdef function and add a return type to have this return a std::vector.  See http://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#python-objects-as-parameters-and-return-values

zxkjack123: I found that we don't need this function anymore as we implemented the properties for all the x, y, z, e, w and c.

zxkjack123:
Update: remove function get_xyzew from `SourceParticle`.

pyne/source_sampling.pyx
scopatz: Ok, if we don't need this function, then  please delete it :)

pyne/source_sampling.pyx
```diff
@@ -267,48 +294,21 @@ cdef class Sampler:
...
-            for ie_bounds in range(e_bounds_size):
-                e_bounds_proxy[ie_bounds] = e_bounds_data[ie_bounds]
-        else:
-            e_bounds_proxy = cpp_vector[double](<size_t> e_bounds_size)
-            for ie_bounds in range(e_bounds_size):
-                e_bounds_proxy[ie_bounds] = <double> e_bounds[ie_bounds]
         bias_tag_name_bytes = bias_tag_name.encode()
-        self._inst = new cpp_source_sampling.Sampler(std_string(<char *> filename_bytes), std_string(<char *> src_tag_name_bytes), e_bounds_proxy, std_string(<char *> bias_tag_name_bytes))
+        # convert e_bounds
+        cdef cpp_vector[double] e_bounds_proxy = convert_e_bounds(e_bounds)
```
scopatz: But I thought it was being used here

zxkjack123: I am sorry. The function removed is `get_xyzew`, but not the `convert_e_bounds`. I mistook these two functions. By the way, the return type is cpp_vector[double], I have changed that in the updated documentation on `convert_e_bounds`.

zxkjack123:
Update: update the documentation on `convert_e_bounds` in `source_sampling.pyx`.

pyne/source_sampling.pyx
scopatz: Ahh ok. Also, the return type **is not** `cpp_vector[double]` because Cython will automatically convert `cpp_vector[double]` to a Python list of Python floats. This is happening because it is `def` (Python) function rather than a `cdef` C/C++ function. See http://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html#standard-library

So what is happening in the code is that the vector is being returned as a list and then the list is being converted back to a vector at the call site. This is exactly what we want to avoid :smile:

zxkjack123: As this is the problem of the `def` and `cdef`. Does that means we can avoid that by define the `convert_e_bounds` as `cdef`?
```
cdef convert_e_bounds(e_bounds):
```

scopatz: yes, but you need to add the return type (and probably the argument type) too:

```cython
cdef cpp_vector[double] convert_e_bounds(e_bounds):
```

zxkjack123: I made the change according to your suggestion, using `cdef` and add return type.

scopatz: Great thanks!

scopatz:
This now looks fine to me.  @gonuke @pshriwise, did you want to comment?

pyne/source_sampling.pyx
```diff
@@ -187,6 +189,31 @@ cdef class AliasTable:
...
+    -------
+    res1 : cpp_vector[double]
+    
+    """
+
+    cdef cpp_vector[double] e_bounds_proxy
+    cdef int ie_bounds
+    cdef int e_bounds_size
+    cdef double * e_bounds_data
+    e_bounds = np.array(e_bounds, dtype=np.float64)
```
gonuke: does this line avoid the complex conditional that is being replaced below?

pyne/source_sampling.pyx
```diff
@@ -187,6 +189,31 @@ cdef class AliasTable:
     pass
 
 
+cdef cpp_vector[double] convert_e_bounds(e_bounds):
```
gonuke: This seems to now essentially be a general purpose function for converting a numpy `ndarray` into a C++ vector - is that right?  Should we rename is such?

Also, is this considered the "best practice" way to do this conversion?

pyne/source_sampling.pyx
```diff
@@ -187,6 +189,31 @@ cdef class AliasTable:
...
+    """convert_e_bounds(rands)
```
gonuke: what does `rands` mean here?

src/source_sampling.cpp
```diff
@@ -74,6 +74,27 @@ pyne::Sampler::Sampler(std::string filename,
...
+  // determine the bias_mode
+  if (mode == 0){
+    bias_mode = ANALOG; 
+  } else if (mode == 1) {
+    bias_mode = UNIFORM;
+  } else if (mode == 2) {
+    bias_mode = USER;
+  }
+
+  // find out the filename, src_tag_name and bias_tag_name
```
gonuke: pendantic point: the `filename` is no longer unpacked here

src/source_sampling.cpp
```diff
@@ -74,6 +74,27 @@ pyne::Sampler::Sampler(std::string filename,
...
+  src_tag_name = tag_names["src_tag_name"];
+  bias_tag_name = tag_names["bias_tag_name"];
```
gonuke: in general, don't we need to check for the existence of this tag?  It might not always exist.

tests/test_source_sampling.py
```diff
@@ -69,7 +71,9 @@ def test_analog_multiple_hex():
     m.src = IMeshTag(2, float)
     m.src[:] = np.ones(shape=(8,2))
     m.mesh.save("sampling_mesh.h5m")
-    sampler = Sampler("sampling_mesh.h5m", "src", np.array([0, 0.5, 1]), False)
+    filename = "sampling_mesh.h5m"
+    tag_names = {"src_tag_name": "src"}
```
gonuke, line 105: I'm surprised that this test passes given the missing `bias_tag_name` key - see comment above.

gonuke (changes requested):
Mostly comments and questions for my own understanding.

One possible robustness issue regarding the `bias_tag_name`

pyne/source_sampling.pyx
zxkjack123: Yes. Sometimes the `e_bounds` is not passed in as a `np.ndarray`. Converting it to `np.ndarray` first can avoid the use of the conditional.

pyne/source_sampling.pyx
zxkjack123: Yes, this function can convert any numpy `ndarray` into a C++ vector. I will rename it in the updates. 
Actually, I am not sure whether this is the "best practice" way to do this conversion, but I am sure it's better than the previous (replaced) method. By the way, the conversion method is suggested by @scopatz .

pyne/source_sampling.pyx
zxkjack123: I am sorry for that. This is a typo, I will remove that `rands`.

src/source_sampling.cpp
zxkjack123: This documentation will be updated.

pyne/source_sampling.pyx
scopatz: Having a conditional or using `np.asarray()` would be better here.  A single conditional is much cheaper than a copy.

pyne/source_sampling.pyx
scopatz: This is pretty close to the best way to copy a numpy array into a vector.

tests/test_source_sampling.py
zxkjack123: This test passes because the `bias_tag_name = tag_names["bias_tag_name"]` happens in C++ part.
In C++, the `map::operator[]` searches the data structure for a value corresponding to the given key, and returns a reference to it. If it can't find one it transparently creates a default constructed element for it, rather than report an error. (In python, it will report an error if the key doesn't exist.)
I have added an input check in the code and add a test function to make sure that it will report an error with corresponding information of which input parameter is missing.

src/source_sampling.cpp
zxkjack123: The input check is added in the update.

zxkjack123:
Update: 
1. Add input check for `tag_names`. 
2. Add a test function to make sure it will report an error at the right place if something missed in `tag_names`. 
3. Update news file.

tests/test_source_sampling.py
```diff
@@ -23,6 +23,38 @@
...
+    m.src = IMeshTag(2, float)
+    m.src[:] = [[2.0, 1.0], [9.0, 3.0]]
+    e_bounds = np.array([0, 0.5, 1.0])
+    m.bias = IMeshTag(2, float)
+    m.bias[:] = [[1.0, 2.0], [3.0, 3.0]]
+    m.mesh.save("sampling_mesh.h5m")
+    filename = "sampling_mesh.h5m"
+
+    # right condition
+    tag_names = {"src_tag_name": "src",
```
gonuke: I don't think this line is needed anymore - the next line replaces it.

gonuke (approved):

gonuke:
Just one small (and minor) change and I'm ready to merge this...

zxkjack123:
Update: Removed an unused line according to the comment of @gonuke .

tests/test_source_sampling.py
zxkjack123: This line has been removed.

zxkjack123:
Is there any further change required? @gonuke @scopatz

scopatz:
This looks good to me, but I am not the expert here, so I will let @gonuke decide.

gonuke:
Thanks @zxkjack123

PR #1000, opened 2018-02-22
R2s fix string match problem

zxkjack123:
In the R2S process, we have to define the decay time in a specific manner, otherwise, there will be an error in r2s step2. For example, we define decay times like `1E3 s` or `3.0 d`, there will be an error. We have to define it like `1000 s`, `3 d`. What's more, we don't know whether there will be an error or not until the step2.

This PR is aimed to solve this problem. Float match is added in the code for better matching. Now we can define decay times in any number format as well as it meets the ALARA unit requirement. For example, decay times like `1E3 s`, `3.0 d` work well now.


pyne/alara.py
```diff
@@ -843,3 +852,57 @@ def _get_subvoxel_array(mesh, cell_mats):
...
+    if unit in ('h', 'hour'):
+        return float(num) * 60 * 60
+    if unit in ('d', 'day'):
+        return float(num) * 60 * 60 * 24
+    if unit in ('w', 'week'):
+        return float(num) * 60 * 60 * 24 * 7
+    if unit in ('y', 'year'):
+        return float(num) * 60 * 60 * 24 * 7 * 52 
+    if unit in ('c', 'century'):
+        return float(num) * 60 * 60 * 24 * 7 * 52 * 100
```
scopatz: First, these should all be `elif` after the first.

pyne/alara.py
```diff
@@ -843,3 +852,57 @@ def _get_subvoxel_array(mesh, cell_mats):
...
     return subvoxel_array
 
+def _convert_unit_to_s(dc):
+    """
+    This function return a float number represent a time in unit of s.
+    Parameters
+    ----------
+    dc : string. Contain a num and an unit. 
+
+    return: a float number
```
scopatz: Please use numpydoc's return section

```
Returns
-------
a float number
```

pyne/alara.py
scopatz: Second, there is a better way to do this.  First put all of your conversion factors in a global (module-level) dict.  Then, you can look up the value from this dict using either indexing or get(). I would write something like:

```python
_TO_SEC = {
  'second': 1.0,
  ...
  'century': 60 * 60 * 24 * 7 * 52 * 100.0,
}
# add abbreviations
_TO_SEC.update({k[0]: v for k, v in _TO_SEC.items()})

def _convert_unit_to_s(dc):
    num, units = dc.split()
    conv = _TO_SEC.get(units, None)
    if conv is None:
        raise ValueError('Invalid unit: {0}'.format(unit))
    return float(num) * conv
```

pyne/alara.py
```diff
@@ -843,3 +852,57 @@ def _get_subvoxel_array(mesh, cell_mats):
...
+    """
+    This function return a float number represent a time in unit of s.
+    Parameters
+    ----------
+    dc : string. Contain a num and an unit. 
+
+    return: a float number 
+    """
+    # get num and unit
+    [num, unit] = dc.split()
```
scopatz: I don't think you need the square brackets on the LHS.  Please remove them.

```
num, unit = dc.split()
```

scopatz:
Thanks for putting this in!

zxkjack123:
Update: Made some clean up according to the suggestions of @scopatz .

scopatz:
@zxkjack123 - this doesn't seem updated other than the removal of the `[]`

zxkjack123:
Update: Changed the unit conversion code according to the suggestion of @scopatz .

pyne/alara.py
```diff
@@ -843,3 +852,64 @@ def _get_subvoxel_array(mesh, cell_mats):
...
+    'h': 60 * 60,
+    'hour': 60 * 60,
+    'd': 60 * 60 * 24,
+    'day': 60 * 60 * 24,
+    'w': 60 * 60 * 24 * 7,
+    'week': 60 * 60 * 24 * 7,
+    'y': 60 * 60 * 24 * 365,
+    'year': 60 * 60 * 24 * 365,
+    'c': 60 * 60 * 24 * 365 * 100,
+    'century': 60 * 60 * 24 * 365 * 100,
```
scopatz: Years should be 365.25 days

zxkjack123:
Update: Change year from 365 days to 365.25 days according to the comment of @scopatz .

pyne/alara.py
zxkjack123: This change has been made in the update.

pyne/alara.py
```diff
@@ -843,3 +852,64 @@ def _get_subvoxel_array(mesh, cell_mats):
...
+    ----------
+    dc : string. Contain a num and an unit. 
+
+    Returns
+    -------
+    a float number
+    """
+    # get num and unit
+    num, unit = dc.split()
+    conv = _TO_SEC[unit]
```
scopatz: this has to be `conv = _TO_SEC.get(unit, None)` to make the following lines work when the unit is not found.

zxkjack123: Thanks for the comment, I have made this change in the update of this PR.

zxkjack123:
Update: Use `_TO_SEC.get(unit, None)` to replace `_TO_SEC[unit]` according to the suggestion of @scopatz .

scopatz:
This looks good to me! Just waiting on CI

Issue #978, opened 2018-01-16
Top-level design of workflow support script(s)


gonuke:
* Create a directory for GT-CADIS files
* Move outline from project description to a README in that directory
* include a link to @biondo's figshare DOI

PR #1006, opened 2018-03-02
GT-CADIS workflow: write PARTISN input file for adjoint  photon transport

### LINKED ISSUE #985: Perform discretization for a user defined Cartesian grid and write input file for deterministic adjoint photon transport
[Step 1]
* Create script analogous to R2S
* One of the initial steps in GT-CADIS workflow
     * Write PARTISN input file for adjoint photon transport

* Generate configuration file that user will fill in
     * Path to hdf5 geometry file
     * Mesh discretization
     * Adjoint source cell ID
     * Source intensity

ngranda:
This PR does the following:

- Fixes #985  (the first step for the GT-CADIS workflow outlined in PR #978)
- Reads material-laden hdf5 geometry file
- Generates mesh to cover entire problem domain based on user-specified values
- Discretizes geometry onto mesh to retrieve material assignment for PARTISN calculation
- Writes PARTISN input file to calculate adjoint photon flux

### DIFF
--- a/scripts/gtcadis.py
+++ b/scripts/gtcadis.py
@@ -16,28 +16,25 @@ config_filename = 'config.ini'
 config = \
 """
 # Optional step to assess all materials in geometry for compatibility with 
-SNILB criteria
+# SNILB criteria
 [step0]
 
 # Prepare PARTISN input file for adjoint photon transport
 [step1]
-# Path to hdf5 geometry file
+# Path to hdf5 geometry file for photon transport
 geom_file:
-# ID of adjoint photon source cell
+# Volume ID of adjoint photon source cell on 
+# DAGMC input [Trelis/Cubit .sat file]
 src_cell:
-# Discretization in x-direction
-# xmin, xmax, number of divisions
-xmesh:
-# Discretization in y-direction
-# ymin, ymax, number of divisions
-ymesh:
-# Discretization in z-direction
-# zmin, zmax, number of divisions
-zmesh:
-# Source intensity
-intensities:
-# Volume of source cell
+# Volume [cm^3] of source cell (detector)
 src_vol:
+# Define mesh that covers entire geometry:
+# xmin, xmax, number of intervals
+xmesh:
+# ymin, ymax, number of intervals
+ymesh:
+# zmin, zmax, number of intervals
+zmesh:
 
 # Calculate T matrix for each material
 [step2]
@@ -49,7 +46,7 @@ src_vol:
 [step4]
 
 # Generate Monte Carlo variance reduction parameters 
-(biased source and weight windows)
+# (biased source and weight windows)
 [step5]
 
 
@@ -61,7 +58,7 @@ def setup():
     with open(config_filename, 'w') as f:
         f.write(config)
     print('File "{}" has been written'.format(config_filename))
-    print('Fill out the fields in these filse then run ">> gtcadis.py step1"')
+    print('Fill out the fields in these files then run ">> gtcadis.py step1"')
 
 def _names_dict():
     names = ['h1', 'd', 'h3', 'he3', 'he4', 'li6', 'li7', 'be9', 'b10', 'b11',
@@ -110,59 +107,61 @@ def _cards():
     return cards
    
 def step1():
+    # Parse config file
     config = ConfigParser.ConfigParser()
     config.read(config_filename)
 
-    hdf5 = config.get('step1', 'geom_file') 
-    cells = [config.get('step1', 'src_cell')]
-    print ('cells', len(cells))
+    # Get user-input from config file
+    geom = config.get('step1', 'geom_file') 
+    cells = [config.getint('step1', 'src_cell')]
+    src_vol = [config.getfloat('step1', 'src_vol')]
     xmesh = config.get('step1', 'xmesh').split(',')
     ymesh = config.get('step1', 'ymesh').split(',')
     zmesh = config.get('step1', 'zmesh').split(',')
-    #intensities = [config.get('step1', 'intensities')]
-    #print (len(intensities))
-    src_vol = [config.getfloat('step1', 'src_vol')]
-    print ('src vol', len(src_vol))
     
-    names_dict = _names_dict()
-    
-    sc = [np.linspace(float(xmesh[0]), float(xmesh[1]), float(xmesh[2])),
-          np.linspace(float(ymesh[0]), float(ymesh[1]), float(ymesh[2])),
-          np.linspace(float(zmesh[0]), float(zmesh[1]), float(zmesh[2]))]
-    
+    # Create structured mesh
+    sc = [np.linspace(float(xmesh[0]), float(xmesh[1]), float(xmesh[2])+1),
+          np.linspace(float(ymesh[0]), float(ymesh[1]), float(ymesh[2])+1),
+          np.linspace(float(zmesh[0]), float(zmesh[1]), float(zmesh[2])+1)]
     mesh = Mesh(structured=True, structured_coords=sc)
-    #hdf5 = "../../geom/photon/photon.h5m"
-    
-    # the first bin has been replaced with 1 for log int.
+   
+    # Generate 42 photon energy bins [eV] 
+    #  First bin has been replaced with 1 for log interpolation
     photon_bins =  [1, 1e4, 2e4, 3e4, 4.5e4, 6e4, 7e4, 7.5e4, 1e5, 1.5e5, 2e5, 3e5, 4e5,
                     4.5e5, 5.1e5, 5.12e5, 6e5, 7e5, 8e5, 1e6, 1.33e6, 1.34e6, 1.5e6, 1.66e6, 2e6,
                     2.5e6, 3e6, 3.5e6, 4e6, 4.5e6, 5e6, 5.5e6, 6e6, 6.5e6, 7e6, 7.5e6, 8e6, 1e7,
                     1.2e7, 1.4e7, 2e7, 3e7, 5e7]
-    # convert to MEV
+    # Convert to MEV
     photon_bins = np.array([x/1E6 for x in photon_bins])
+    # ICRP 74 flux-to-dose conversion factors
     de = np.array([0.01, 0.015, 0.02, 0.03, 0.04, 0.05, 0.06, 0.07, 0.08, 0.1, 0.15, 0.2, 0.3,
                    0.4, 0.5, 0.6, 0.8, 1, 2, 4, 6, 8, 10])
     df = np.array([0.0485, 0.1254, 0.205, 0.2999, 0.3381, 0.3572, 0.378, 0.4066, 0.4399, 0.5172,
                    0.7523, 1.0041, 1.5083, 1.9958, 2.4657, 2.9082, 3.7269, 4.4834, 7.4896,
                    12.0153, 15.9873, 19.9191, 23.76])
-    # convert to Sv/s per photon FLUX (not fluence)
+    # Convert to Sv/s per photon FLUX (not fluence)
     df = np.array([x*1E-12 for x in df])
+    # Convert pointwise data to group data for log interpolation
     photon_spectrum = pointwise_collapse(photon_bins, de, df, logx=True, logy=True)
-    # anything below 0.01 MeV should be assigned the DF value of 0.01 MeV
+    #  Anything below 0.01 MeV should be assigned the DF value of 0.01 MeV
     photon_spectrum[0] = df[0]
+    # Total number of groups is 217 (42 photon + 175 neutron)
     spectra = [np.append(photon_spectrum, np.zeros(175))]
-    print('spectra', len(spectra))
     # The spectrum is normalized by PyNE, so we need to mutliply by the sum of intensities in the spectrum.
     # Additionally, we divide by the volume of the source cell in order to get source density.
     intensities = [np.sum(spectra)/src_vol]
     
-    load(hdf5)
-    source, dg = isotropic_vol_source(hdf5, mesh, cells, spectra, intensities)
-    
+    # Load geometry into DAGMC 
+    load(geom)
+    # Generate isotropic photon volume source
+    source, dg = isotropic_vol_source(geom, mesh, cells, spectra, intensities)
+   
+    # PARTISN input 
     ngroup = 217
     cards = _cards()
+    names_dict = _names_dict()
     
-    write_partisn_input(mesh, hdf5, ngroup, cards=cards, dg=dg, names_dict=names_dict, data_hdf5path="/materials", nuc_hdf5path="/nucid", fine_per_coarse=1)
+    write_partisn_input(mesh, geom, ngroup, cards=cards, dg=dg, names_dict=names_dict, data_hdf5path="/materials", nuc_hdf5path="/nucid", fine_per_coarse=1)
 
 def main():
 
--- a/scripts/gtcadis.py
+++ b/scripts/gtcadis.py
@@ -14,8 +14,8 @@ from pyne.bins import pointwise_collapse
 config_filename = 'config.ini'
 
 config = \
-"""
-# Optional step to assess all materials in geometry for compatibility with 
+    """
+# Optional step to assess all materials in geometry for compatibility with
 # SNILB criteria
 [step0]
 
@@ -23,7 +23,7 @@ config = \
 [step1]
 # Path to hdf5 geometry file for photon transport
 geom_file:
-# Volume ID of adjoint photon source cell on 
+# Volume ID of adjoint photon source cell on
 # DAGMC input [Trelis/Cubit .sat file]
 src_cell:
 # Volume [cm^3] of source cell (detector)
@@ -45,13 +45,12 @@ zmesh:
 # Prepare PARTISN input for adjoint neutron transport
 [step4]
 
-# Generate Monte Carlo variance reduction parameters 
+# Generate Monte Carlo variance reduction parameters
 # (biased source and weight windows)
 [step5]
 
 
 """
-
 
 
 def setup():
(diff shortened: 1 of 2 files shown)


scripts/gtcadis.py
```diff
@@ -0,0 +1,205 @@
...
+
+def setup():
+    with open(config_filename, 'w') as f:
+        f.write(config)
+    print('File "{}" has been written'.format(config_filename))
+    print('Fill out the fields in this file then run ">> gtcadis.py step1"')
+
+
+def _names_dict():
+    names = ['h1', 'd', 'h3', 'he3', 'he4', 'li6', 'li7', 'be9', 'b10', 'b11',
```
gonuke: Reformat this as a dictionary

scripts/gtcadis.py
```diff
@@ -0,0 +1,205 @@
...
+             "block5": {"source": source,
+                        "ith": 1,
+                        "isct": 5}
+             }
+    return cards
+
+
+def step1():
+    # Parse config file
+    config = ConfigParser.ConfigParser()
```
gonuke: move this into `main` as it will be common to all steps

scripts/gtcadis.py
```diff
@@ -0,0 +1,205 @@
...
+    # Convert to MEV
+    photon_bins = np.array([x / 1E6 for x in photon_bins])
+    # ICRP 74 flux-to-dose conversion factors
+    de = np.array([0.01, 0.015, 0.02, 0.03, 0.04, 0.05, 0.06, 0.07, 0.08, 0.1, 0.15, 0.2, 0.3,
+                   0.4, 0.5, 0.6, 0.8, 1, 2, 4, 6, 8, 10])
+    df = np.array([0.0485, 0.1254, 0.205, 0.2999, 0.3381, 0.3572, 0.378, 0.4066, 0.4399, 0.5172,
+                   0.7523, 1.0041, 1.5083, 1.9958, 2.4657, 2.9082, 3.7269, 4.4834, 7.4896,
+                   12.0153, 15.9873, 19.9191, 23.76])
+    # Convert to Sv/s per photon FLUX (not fluence)
+    df = np.array([x * 1E-12 for x in df])
```
gonuke: magic number - what's the basis of this number - define it as a variable for semantics

scripts/gtcadis.py
```diff
@@ -0,0 +1,205 @@
...
+    m.mesh.save("blank_mesh.h5m")
+
+    # Generate 42 photon energy bins [eV]
+    #  First bin has been replaced with 1 for log interpolation
+    photon_bins = [1, 1e4, 2e4, 3e4, 4.5e4, 6e4, 7e4, 7.5e4, 1e5, 1.5e5, 2e5, 3e5, 4e5,
+                   4.5e5, 5.1e5, 5.12e5, 6e5, 7e5, 8e5, 1e6, 1.33e6, 1.34e6, 1.5e6, 1.66e6, 2e6,
+                   2.5e6, 3e6, 3.5e6, 4e6, 4.5e6, 5e6, 5.5e6, 6e6, 6.5e6, 7e6, 7.5e6, 8e6, 1e7,
+                   1.2e7, 1.4e7, 2e7, 3e7, 5e7]
+    # Convert to MEV
+    photon_bins = np.array([x / 1E6 for x in photon_bins])
```
gonuke: magic number - make it a variable

scripts/gtcadis.py
```diff
@@ -0,0 +1,205 @@
...
+    # Create structured mesh
+    sc = [np.linspace(float(xmesh[0]), float(xmesh[1]), float(xmesh[2]) + 1),
+          np.linspace(float(ymesh[0]), float(ymesh[1]), float(ymesh[2]) + 1),
+          np.linspace(float(zmesh[0]), float(zmesh[1]), float(zmesh[2]) + 1)]
+    m = Mesh(structured=True, structured_coords=sc)
+    m.mesh.save("blank_mesh.h5m")
+
+    # Generate 42 photon energy bins [eV]
+    #  First bin has been replaced with 1 for log interpolation
+    photon_bins = [1, 1e4, 2e4, 3e4, 4.5e4, 6e4, 7e4, 7.5e4, 1e5, 1.5e5, 2e5, 3e5, 4e5,
```
gonuke: Just define this in MeV directly?

scripts/gtcadis.py
gonuke: This can be a dictionary provided by the user, perhaps in the config file, or in a separate file referred to by the config file.

scripts/gtcadis.py
```diff
@@ -0,0 +1,205 @@
...
+                   4.5e5, 5.1e5, 5.12e5, 6e5, 7e5, 8e5, 1e6, 1.33e6, 1.34e6, 1.5e6, 1.66e6, 2e6,
+                   2.5e6, 3e6, 3.5e6, 4e6, 4.5e6, 5e6, 5.5e6, 6e6, 6.5e6, 7e6, 7.5e6, 8e6, 1e7,
+                   1.2e7, 1.4e7, 2e7, 3e7, 5e7]
+    # Convert to MEV
+    photon_bins = np.array([x / 1E6 for x in photon_bins])
+    # ICRP 74 flux-to-dose conversion factors
```
gonuke: Move this to some standard reference location in PyNE

scripts/gtcadis.py
```diff
@@ -0,0 +1,205 @@
...
+    de = np.array([0.01, 0.015, 0.02, 0.03, 0.04, 0.05, 0.06, 0.07, 0.08, 0.1, 0.15, 0.2, 0.3,
+                   0.4, 0.5, 0.6, 0.8, 1, 2, 4, 6, 8, 10])
+    df = np.array([0.0485, 0.1254, 0.205, 0.2999, 0.3381, 0.3572, 0.378, 0.4066, 0.4399, 0.5172,
+                   0.7523, 1.0041, 1.5083, 1.9958, 2.4657, 2.9082, 3.7269, 4.4834, 7.4896,
+                   12.0153, 15.9873, 19.9191, 23.76])
+    # Convert to Sv/s per photon FLUX (not fluence)
+    df = np.array([x * 1E-12 for x in df])
+    # Convert pointwise data to group data for log interpolation
+    photon_spectrum = pointwise_collapse(
```
gonuke: variable naming?? this is really a response function since this is all about adjoints

gonuke:
Add a `news` file - see other recent PRs.

scripts/gtcadis.py
```diff
@@ -0,0 +1,205 @@
...
+# Path to hdf5 geometry file for photon transport
+geom_file:
+# Volume ID of adjoint photon source cell on
+# DAGMC input [Trelis/Cubit .sat file]
+src_cell:
+# Volume [cm^3] of source cell (detector)
+src_vol:
+# Define mesh that covers entire geometry:
+# xmin, xmax, number of intervals
+xmesh:
```
gonuke: replace with `origin`, `[x|y|z]mesh` and `[x|y|z]ints` to be similar to MCNP structure (or fold the origin into the *mesh?)

gonuke (changes requested):
Progress!

scripts/gtcadis.py
```diff
@@ -0,0 +1,205 @@
...
+
+import numpy as np
+from pyne.mesh import Mesh
+from pyne.partisn import write_partisn_input, isotropic_vol_source
+from pyne.dagmc import discretize_geom, load
+from pyne import nucname
+from pyne.bins import pointwise_collapse
+
+
+config_filename = 'config.ini'
```
scopatz: Config files are rather old school.  Is there any way the config file could be in a more modern format, such as JSON or YAML?

scripts/gtcadis.py
scopatz: Also, this doesn't take advantage of numpy at all. The line should be, 

```python
df = df * 1e-12
```

scopatz:
Thanks for putting this in @ngranda!

scripts/gtcadis.py
gonuke: Perhaps old-fashioned, but supported natively within Python.  JSON is as well, but is perhaps a little more intimidating for users.

What YAML parser do you recommend?

scopatz: Yeah I agree that native support in python is desirable. I recommend `ruamel.yaml`, since it is pretty ubiquitous at this point.

scripts/gtcadis.py
```diff
@@ -0,0 +1,212 @@
...
+
+def setup():
+    with open(config_filename, 'w') as f:
+        f.write(config)
+    print('File "{}" has been written'.format(config_filename))
+    print('Fill out the fields in this file then run ">> gtcadis.py step1"')
+
+
+def _names_dict():
+    names = {'h1': 'h1', 'd': 'h2', 'h3': 'h3', 'he3': 'he3',
```
gonuke: I think this dictionary is formatted backwards - need to swap keys <-> values

scripts/gtcadis.py
```diff
@@ -0,0 +1,212 @@
...
+        names_dict=names_dict,
+        data_hdf5path="/materials",
+        nuc_hdf5path="/nucid",
+        fine_per_coarse=1)
+
+
+def main():
+
+    # Print blank config file
+    data = yaml.load(config)
```
gonuke: not sure what, if anything, these 2 lines do?

scripts/gtcadis.py
```diff
@@ -0,0 +1,212 @@
...
+    config_file = yaml.dump(data, default_flow_style=False)
+    with open(config_filename, 'r') as f:
+        cfg = yaml.load(f)
+
+    gtcadis_help = ('This script automates the GT-CADIS process of \n'
+                    'producing variance reduction parameters to optimize the\n'
+                    'neutron transport step of the Rigorous 2-Step (R2S) method.\n')
+    setup_help = ('Prints the file "config.ini" to be\n'
```
gonuke: wrong file name in this comment/user help msg

scripts/gtcadis.py
```diff
@@ -0,0 +1,212 @@
...
+        nuc_hdf5path="/nucid",
+        fine_per_coarse=1)
+
+
+def main():
+
+    # Print blank config file
+    data = yaml.load(config)
+    config_file = yaml.dump(data, default_flow_style=False)
+    with open(config_filename, 'r') as f:
```
gonuke: Shouldn't this be in a conditional below - if running `setup` it might fail somehow.

scripts/gtcadis.py
```diff
@@ -0,0 +1,212 @@
...
+                        "kwikrd": 0
+                        },
+             "block5": {"source": source,
+                        "ith": 1,
+                        "isct": 5}
+             }
+    return cards
+
+
+def step1(cfg):
```
gonuke, line 126: perhaps have the calling code pass in `cfg['step1']` to make this a little less verbose

scripts/gtcadis.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # Get user-input from config file
+    geom = cfg['step1']['geom_file']
+    cells = [cfg['step1']['src_cell']]
+    src_vol = [float(cfg['step1']['src_vol'])]
+    xmesh = cfg['step1']['xmesh'].split(',')
```
gonuke: You could get some built-in error checking if you did this:
```
try:
   xmin, xmax, xints = cfg['step1']['xmesh'].split(',')
except:
   print("Too few entries for the mesh intervals in the x-direction (xmesh)")
```

gonuke: You'd then use `xmin`, `xmax`, and `xints` in the `linspace` method below

scripts/gtcadis.py
```diff
@@ -148,7 +141,8 @@ def step1(cfg):
                    0.7523, 1.0041, 1.5083, 1.9958, 2.4657, 2.9082, 3.7269, 4.4834, 7.4896,
                    12.0153, 15.9873, 19.9191, 23.76])
     # Convert to Sv/s per photon FLUX (not fluence)
-    df = np.array([x * 1E-12 for x in df])
+    flux_magnitude = 1E12
```
gonuke: This changed from `1e-12` to `1e12` in the most recent commit - which is correct?

Related: what is the purpose of this anyway?

gonuke (changes requested):
Thanks for the updates - I think there are a few more changes here.  Do you have any tests?

scripts/gtcadis.py
```diff
@@ -0,0 +1,212 @@
...
+                            1.5, 1.66, 2, 2.5, 3, 3.5, 4, 4.5, 5, 5.5, 6, 6.5, 7, 7.5, 8, 10,
+                            12, 14, 20, 30, 50])
+    # ICRP 74 flux-to-dose conversion factors
+    de = np.array([0.01, 0.015, 0.02, 0.03, 0.04, 0.05, 0.06, 0.07, 0.08, 0.1, 0.15, 0.2, 0.3,
+                   0.4, 0.5, 0.6, 0.8, 1, 2, 4, 6, 8, 10])
+    df = np.array([0.0485, 0.1254, 0.205, 0.2999, 0.3381, 0.3572, 0.378, 0.4066, 0.4399, 0.5172,
+                   0.7523, 1.0041, 1.5083, 1.9958, 2.4657, 2.9082, 3.7269, 4.4834, 7.4896,
+                   12.0153, 15.9873, 19.9191, 23.76])
+    # Convert to Sv/s per photon FLUX (not fluence)
+    flux_magnitude = 1E12
```
gonuke: This changed from 1e-12 to 1e12 in the most recent commit - which is correct?

Related: what is the purpose of this anyway?

gonuke:
Ultimately, I think we need to think about moving more things to the config file...

scripts/gtcadis.py
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/env python
+import argparse
+import yaml
```
scopatz, line 3: I suggest `ruamel.yaml`, docs here http://yaml.readthedocs.io/en/latest/

scripts/gtcadis.py
```diff
@@ -0,0 +1,212 @@
...
+             'cu65': 'cu65', 'ganat': 'ga', 'zrnat': 'zr',
+             'nb93': 'nb93', 'mo92': 'mo92', 'mo94': 'mo94',
+             'mo95': 'mo95', 'mo96': 'mo96', 'mo97': 'mo97',
+             'mo98': 'mo98', 'mo100': 'mo100', 'snnat': 'sn',
+             'ta181': 'ta181', 'w182': 'w182', 'w183': 'w183',
+             'w184': 'w184', 'w186': 'w186', 'au197': 'au197',
+             'pb206': 'pb206', 'pb207': 'pb207', 'pb208': 'pb208',
+             'bi209': 'bi209'}
+
+    names_dict = {nucname.id(value): key for key, value in names.iteritems()}
```
scopatz: just use `names.items()`.  `iteritems()` was renamed to `items()` in Python 3

scripts/gtcadis.py
```diff
@@ -0,0 +1,212 @@
...
+
+# Generate Monte Carlo variance reduction parameters
+# (biased source and weight windows)
+step5:
+
+
+"""
+
+
+def setup():
```
scopatz, line 63: Please provide a documentation string for public functions

scripts/gtcadis.py
scopatz: Also, please provide a doc string

scripts/gtcadis.py
```diff
@@ -0,0 +1,212 @@
...
+        cards=cards,
+        dg=dg,
+        names_dict=names_dict,
+        data_hdf5path="/materials",
+        nuc_hdf5path="/nucid",
+        fine_per_coarse=1)
+
+
+def main():
+
```
scopatz: Please provide a docstring

scripts/gtcadis.py
gonuke: For the current use case, is there motivation to add an external dependency (given that not everyone lives in a conda world)?  This `yaml` appears to be relatively standard (I guess I don't really know)

scopatz: either one is fine, just my recommendation!

scripts/gtcadis.py
ngranda: Thanks! I have implemented both comments.

scripts/gtcadis.py
ngranda: done

scripts/gtcadis.py
```diff
@@ -0,0 +1,237 @@
...
+        fine_per_coarse=1)
+
+
+def main():
+    """ This function manages the setup and steps 1-5 for the GT-CADIS workflow.
+    """
+
+    # Print blank config file
+    data = yaml.load(config)
+    config_file = yaml.dump(data, default_flow_style=False)
```
gonuke: `config_file` is never used - is this line necessary?

ngranda: You are correct, I do not need to save the second line in a variable. I have removed this and combined both lines. I do still need yaml.dump as this writes the file. 
yaml.load creates a dictionary object which I need to write the correct file format with yaml.dump.

scripts/gtcadis.py
```diff
@@ -0,0 +1,237 @@
...
+        nuc_hdf5path="/nucid",
+        fine_per_coarse=1)
+
+
+def main():
+    """ This function manages the setup and steps 1-5 for the GT-CADIS workflow.
+    """
+
+    # Write blank config file
+    yaml.dump(yaml.load(config), default_flow_style=False)
```
gonuke: If I understand the documentation correctly, `yaml.dump` returns a string and doesn't write a file.  The blank file gets written in the `setup` method by just writing the already-correctly-formatted string `config`.

I don't think this line does anything

ngranda: You are correct. I removed the line

scripts/gtcadis.py
```diff
@@ -0,0 +1,233 @@
...
+        nuc_hdf5path="/nucid",
+        fine_per_coarse=1)
+
+
+def main():
+    """ This function manages the setup and steps 1-5 for the GT-CADIS workflow.
+    """
+
+    with open(config_filename, 'r') as f:
+        cfg = yaml.load(f)
```
moatazharb: Shouldn't the above two lines come later after 'setup' is run?. This should give an error as the config.yml file hasn't been written yet at this point till 'setup' is run. I guess it should be moved below "if args.command == 'setup': setup()"

scripts/gtcadis.py
```diff
@@ -0,0 +1,233 @@
...
+config_filename = 'config.yml'
+
+config = \
+    """
+# Optional step to assess all materials in geometry for compatibility with
+# SNILB criteria
+step0:
+
+# Prepare PARTISN input file for adjoint photon transport
+step1:
```
moatazharb, line 23: I guess step1: should be 'step1': as it is a nested collection.

gonuke: I am not sure what you mean @moatazharb ?  Do you mean it should have quotes?

scripts/gtcadis.py
```diff
@@ -0,0 +1,234 @@
...
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers(help=gtcadis_help, dest='command')
+
+    setup_parser = subparsers.add_parser('setup', help=setup_help)
+    step1_parser = subparsers.add_parser('step1', help=step1_help)
+
+    args, other = parser.parse_known_args()
+    if args.command == 'setup':
+        setup()
+    with open(config_filename, 'r') as f:
```
gonuke: Now that you've moved this...  There is still a bit of a logic issue, mostly for "cleanliness" and not for correctness.

This will open and read the file always, regardless of which step you are invoking.  If you are invoking `setup` it will read it and do nothing - not the end of the world, but a little messy.

should we only read this file if we need it?

ngranda: Right, I was wondering about that. We would need to read it for every step except for set up. Would it be better to read it in the elif statement for each step?

gonuke: You just have an `else` on the test for `setup`.  Then it will happen every time you are NOT doing `setup`.

ngranda: done

gonuke (approved):

gonuke:
Thanks for all your work on this @ngranda

scripts/gtcadis.py
moatazharb: Yes I meant quotes. Never mind I was trying something and it gave an error so I though it was necessary to have quotes. It's ok without quotes. My bad :(

PR #1009, opened 2018-03-03
Remove alarapy sec conv

zxkjack123:
I added a map `_TO_SEC` in `alara.py` for the purpose of time unit conversion in PR #1000.  But I found that there is a function `to_sec` do the same thing in `untils.py`. So, I modified `to_sec` and use it replace the `_TO_SEC` in `alara.py`. 

Major changes:
1. Add units `week` and `century` in `to_sec`.
2. Use `to_sec` to replace `_TO_SEC` in `alara.py`.

### DIFF
--- a/pyne/utils.py
+++ b/pyne/utils.py
@@ -49,9 +49,15 @@ time_conv_dict = {'as': 1e-18,
                   'd': 86400.0,
                   'day': 86400.0,
                   'days': 86400.0,
+                  'w': 86400.0*7.0,
+                  'week': 86400.0*7.0,
+                  'weeks': 86400.0*7.0,
                   'y': 86400.0*365.25,
                   'year': 86400.0*365.25,
                   'years': 86400.0*365.25,
+                  'c': 86400.0*365.25*100,
+                  'century': 86400.0*365.25*100,
+                  'centuries': 86400.0*365.25*100,
                   }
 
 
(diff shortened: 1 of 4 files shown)


pyne/utils.py
```diff
@@ -71,9 +77,12 @@ def to_sec(input_time, units):
...
     """
-    sec_time = input_time * time_conv_dict[units.lower()]
-    return sec_time
-
+    conv = time_conv_dict.get(units.lower(), None)
+    if conv:
+        sec_time = input_time * conv
+        return sec_time
+    else:
+        raise ValueError('Invalid units: {0}'.format(units))
```
gonuke, line 85: This is a nice addition - can you add a test for this in `test_util.py`, please?

gonuke (changes requested):
This looks like a good improvement.  I've requestd a little more testing, that's all.

pyne/utils.py
zxkjack123: I expanded the test function of `utils.to_sec` for all defined units and an undefined unit (leading to `ValueError` ).

zxkjack123:
Update:
* Add some test cases of `utils.to_sec` for all defined units 
* Add a test case for an undefined unit test that leads to `ValueError`

scopatz:
This looks great to me, thanks!

PR #1010, opened 2018-03-05
Subvoxel add subvoxel

zxkjack123:
This PR adds sub-voxel R2S codes. 
Major changes:
1. Add sub-voxel R2S codes in source_sampling
2. Add test functions for sub-voxel R2S codes

### DIFF
--- a/src/source_sampling.cpp
+++ b/src/source_sampling.cpp
@@ -12,6 +12,15 @@ void pyne::sampling_setup_(int* mode) {
     std::string src_tag_name ("source_density");
     std::string e_bounds_file ("e_bounds");
     std::vector<double> e_bounds = read_e_bounds(e_bounds_file);
+    std::map<std::string, std::string> tag_names;
+    tag_names.insert(std::pair<std::string, std::string> ("src_tag_name",
+          "source_density"));
+    tag_names.insert(std::pair<std::string, std::string> ("bias_tag_name",
+          "biased_source_density"));
+    tag_names.insert(std::pair<std::string, std::string> ("cell_number_tag_name",
+          "cell_number"));
+    tag_names.insert(std::pair<std::string, std::string> ("cell_fracs_tag_name",
+          "cell_fracs"));
     if (*mode == 0) {
       sampler = new pyne::Sampler(filename, src_tag_name, e_bounds, false);
     } else if (*mode == 1) {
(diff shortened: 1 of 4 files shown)


scopatz:
Code LGTM, but it is not my area.

scopatz:
Thanks for putting it in!

src/source_sampling.cpp
```diff
@@ -203,24 +243,51 @@ void pyne::Sampler::mesh_tag_data(moab::Range ves,
...
+  for(int i=0; i<cell_fracs.size(); i++){
+    cell_fracs[i] = 1.0;
+  }
+  if (sub_mode == SUBVOXEL) {
+      // Read the cell_number tag and cell_fracs tag
+      rval = mesh->tag_get_handle(cell_number_tag_name.c_str(),
+                                  cell_number_tag);
+      rval = mesh->tag_get_handle(cell_fracs_tag_name.c_str(),
+                                  cell_fracs_tag);
+      max_num_cells = num_groups(cell_fracs_tag);
```
gonuke, line 281: Isn’t this necessary before setting the size of `cell_fracs`?

gonuke: I see - you resize again below...  maybe you should split this `SUBVOXEL` block into 2 and resize `cell_fracs` in the middle

gonuke (changes requested):
Thank @zxkjack123 - this looks good.  

Is it possible to combine the tests? That is, have the internal parts of each test be the same, but with different initialization and final testing?

zxkjack123:
@gonuke Actually, there isn't a same code part in each test.

Though all the test cases follow the same pattern:
1. Set initial parameters
2. Initial the Sampler
3. Sampling the source particles in the loop, do the tally. There will also be some checks inside the loop.
4. Check the sample results. 

Almost all the initial parameters are different from case to case, because they are designed to test for different conditions. And the check items in the loop are also different. Of course the result check parts are different.

src/source_sampling.cpp
```diff
@@ -107,15 +121,35 @@ pyne::Sampler::Sampler(std::string filename,
...
+      // cell_number_tag_name not found
+      throw std::invalid_argument("cell_number_tag_name not dound");
+    } else {
+      // found cell_number_tag_name
+      cell_number_tag_name = tag_names["cell_number_tag_name"];
+    }
+    // cell_fracs_tag
+    if (tag_names.find("cell_fracs_tag_name") == tag_names.end()) {
+      // cell_fracs_tag_name not found
+      throw std::invalid_argument("cell_fracs_tag_name not dound");
```
scopatz: *found

zxkjack123: Sorry for the typo, I have fixed them in the update.

gonuke (approved):

gonuke:
Thanks @zxkjack123 - as we discussed, we'll refactor tests under a different PR

PR #1011, opened 2018-03-06
Treat mesh cell tag size 1

zxkjack123:
In PR #971 , the problem of writing a (N, 1) shaped numpy array into IMeshTag was solved. However, there remains a problem of reading the tag data. When we write a (N, 1) array into a Tag, it is actually stored by a (N, ) list rather than (N, 1). 

The reason is that, we first initial an IMeshTag by: https://github.com/pyne/pyne/blob/b5fd890986e4179ac0975d19e796208015ba44a0/pyne/mesh.py#L873. And then set the data: https://github.com/pyne/pyne/blob/b5fd890986e4179ac0975d19e796208015ba44a0/pyne/mesh.py#L882

At the initial step, the (N, 1) and (N, ) array have the same `size=1`, and they are all initialized as (N, ). We can't differentiate them without another parameter, for example: `shape=(N, )` or `shape=(N, 1)`. 

Therefore, when we read the tag data, we need to add some special treating in case that it is a  (N, 1) tag. Becasue when the tag is (N, 1), we get a `int` or `double` rather than a list of `int` or list of `double`. 

This PR add this special treat for the (N, 1) shaped tag. 

Major changes:
1. Add special treat for (N, 1) tag in pyne/alara.py
2. Add a test function for a (N, 1) shaped tag

### DIFF
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -224,7 +224,7 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
         subvoxel_array = _get_subvoxel_array(mesh, cell_mats)
         # get max_num_cells
         ve0_cell_number_tag = mesh.mesh.getTagHandle('cell_number')[ve0]
-        if type(ve0_cell_number_tag) in (int, np.int32, np.int64):
+        if isinstance(ve0_cell_number_tag, (int, np.int32, np.int64)):
             max_num_cells = 1
         else:
             max_num_cells = \
@@ -850,7 +850,7 @@ def _get_subvoxel_array(mesh, cell_mats):
     non_void_sv_num = 0
     for i, _, ve in mesh:
         cell_numbers = cell_number_tag[ve]
-        if type(cell_numbers) in (int, np.int32, np.int64):
+        if isinstance(cell_numbers, (int, np.int32, np.int64)):
             cell_numbers = [cell_numbers]
         for c, cell in enumerate(cell_numbers):
             if cell > 0 and len(cell_mats[cell].comp): #non-void cell
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -222,7 +222,12 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
     if sub_voxel:
         num_vol_elements = len(mesh)
         subvoxel_array = _get_subvoxel_array(mesh, cell_mats)
-        max_num_cells = \
+        # get max_num_cells
+        ve0_cell_number_tag = mesh.mesh.getTagHandle('cell_number')[ve0]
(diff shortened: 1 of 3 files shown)


pyne/alara.py
```diff
@@ -222,7 +222,12 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
     if sub_voxel:
         num_vol_elements = len(mesh)
         subvoxel_array = _get_subvoxel_array(mesh, cell_mats)
-        max_num_cells = \
+        # get max_num_cells
+        ve0_cell_number_tag = mesh.mesh.getTagHandle('cell_number')[ve0]
+        if type(ve0_cell_number_tag) in (int, np.int32, np.int64):
```
scopatz: This should probably be `if isinstance(ve0_cell_number_tag, (int, np.int32, np.int64)):`

pyne/alara.py
```diff
@@ -844,7 +849,10 @@ def _get_subvoxel_array(mesh, cell_mats):
     # calculate the total number of non-void sub-voxel
     non_void_sv_num = 0
     for i, _, ve in mesh:
-        for c, cell in enumerate(cell_number_tag[ve]):
+        cell_numbers = cell_number_tag[ve]
+        if type(cell_numbers) in (int, np.int32, np.int64):
```
scopatz: Same, this should be an isisntance check

pyne/alara.py
zxkjack123: Thanks for the suggestion. I made this change in the update.

zxkjack123:
Update: Use `isinstance` to replace the `type` check.

scopatz:
This looks reasonable to me.  @gonuke? Thanks for putting this in @zxkjack123

gonuke:
I haven't looked at this carefully enough myself to decide, but is it possible to use `numpy.atleast_1d()`  to simplify this?

zxkjack123:
@gonuke I don't think so. Becasue there is a 'numpy.atleast_1d()` before setting the value: https://github.com/pyne/pyne/blob/b5fd890986e4179ac0975d19e796208015ba44a0/pyne/mesh.py#L400

pyne/alara.py
```diff
@@ -222,7 +222,12 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
     if sub_voxel:
         num_vol_elements = len(mesh)
         subvoxel_array = _get_subvoxel_array(mesh, cell_mats)
-        max_num_cells = \
+        # get max_num_cells
```
zxkjack123: Use `np.atleast_1d` to replace this

gonuke:
Thanks @zxkjack123 - this looks good.

PR #1013, opened 2018-03-06
Subvoxel r2s workflow support

zxkjack123:
This PR changes some code of `pyne/scripts/r2s.py` to add support for subvoxel r2s workflow. 
Now, this is a normal r2s and subvoxel r2s compatible workflow, with the normal r2s be the default. 

Major changes:
1. Keep the `cell_number`, `cell_fracs` tags in step2
2. Load geom and calculate cell_mats again in step2

This PR doesn't provide any test function because this change can't be tested within the framework of pyne itself. A simple subvoxel r2s calculation is used to test the workflow.

### DIFF
--- a/scripts/r2s.py
+++ b/scripts/r2s.py
@@ -142,9 +142,10 @@ def step1():
         mesh = Mesh(structured=False, mesh=meshtal)
 
     ves = list(mesh.iter_ve())
+    tags_keep = ("cell_number", "cell_fracs",
+                 "cell_largest_frac_number", "cell_largest_frac")
     for tag in mesh.mesh.getAllTags(ves[0]):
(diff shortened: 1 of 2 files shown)


scopatz:
Seems OK to me! Thanks for putting this in!

scripts/r2s.py
```diff
@@ -143,7 +143,9 @@ def step1():
 
     ves = list(mesh.iter_ve())
     for tag in mesh.mesh.getAllTags(ves[0]):
-        mesh.mesh.destroyTag(tag, True)
+        if tag.name not in ("cell_number", "cell_fracs",\
```
gonuke: This list of tags to keep should be in a variable of its own

gonuke:
Thanks @zxkjack123 - this is great progress!

PR #1014, opened 2018-03-12
fix cell_mats in r2s step2

zxkjack123:
Fix a bug in subvoxel r2s workflow. 
`cell_mats` is used as a parameter for both r2s and subvoxel r2s for the purpose of compatibility. 
```
photon_source_hdf5_to_mesh(mesh, h5_file, tags, sub_voxel=sub_voxel,
                                    cell_mats=cell_mats)
```
I made a mistake in PR #1013, where I set the `cell_mats` for only `subvoxel` is `True`. That breaks the normal r2s, because the `cell_mats` is not set.

### DIFF
--- a/news/subvoxel_r2s_workflow_cell_mats.rst
+++ b/news/subvoxel_r2s_workflow_cell_mats.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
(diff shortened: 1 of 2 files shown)


scopatz:
LGTM

PR #1019, opened 2018-03-28
Restruct source sampling tests

zxkjack123:
This PS provides a template that could be used for test cases of srouce_sampling under different modes mesh geometries, src_tags, e_bounds, and bias_tags. By applying this template, we can perform the test by just giving a few parameters, such as `mode`, `cell_fracs`, `src_tag` and `bias_tag`. The template will automatically calculate the expected results, perform the source sampling, and then check the results. 
Major changes:
- Add a template for source sampling tests
- Several support functions used by the template
- An example of using this template to do tests of different conditions


scopatz:
Always happy to have more tests! Thanks @zxkjack123!

PR #1021, opened 2018-04-03
positive ensdf nuc ids in gamma x-rays

### LINKED ISSUE #1020: data.gamma_xrays runtime error
Hi, this is Andreas Vogt.
When calling 
In [8]: data.gamma_xrays(892270000)
I get the following runtime error: 

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-7-4e8f7ba066c4> in <module>()
----> 1 data.gamma_xrays(892270000)
data.pyx in pyne.data.gamma_xrays()
RuntimeError: P��

Other nuclei work just fine, e.g.
In [8]: data.gamma_xrays(902320000)
Out[8]:
[(88.47, 0.002807402708729183),
 (85.43, 0.001718691938284006),
 (100.0, 0.001308041352986811),
 (12.3, 7.139185247404002)]

Is this a matter of the h5 table? I checked that data.gamma_xrays works at least for all even-mass nuclei along the 232Th and 238U decay chains.

Best regards!

scopatz:
should fix #1020 

CC @FlanFlanagan @pyne/core

### DIFF
--- a/news/gammakey.rst
+++ b/news/gammakey.rst
@@ -1,0 +1,15 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fixed issue where some gamma x-rays where throwing ``NotANuclide`` errors
+  because the underlying nuclides were being read & recorded with negative ids.
+  All nuclide ids are now ensured to be positive.
+
+**Security:** None
--- a/pyne/ensdf.py
+++ b/pyne/ensdf.py
@@ -701,7 +701,7 @@ def _parse_decay_dataset(lines, decay_s):
             else:
                 bparent = parent2
             level = 0.0 if level is None else level
-            bdaughter = data.id_from_level(_to_id(daughter), level)
+            bdaughter = abs(data.id_from_level(_to_id(daughter), level))
             betas.append([bparent, bdaughter, dat[0], 0.0, dat[2]])
         bc_rec = _betac.match(line)
         if bc_rec is not None:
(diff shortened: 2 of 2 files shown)


gonuke:
Do we know if the use of negative numbers means anything in ENSDF?

scopatz:
The negative numbers didn't come from ENSDF itself.  Rather they were an artifact of computing level differences, and came soley from PyNE :frowning_face: ENSDF represents nuclides in AAALL form

gonuke:
Shouldn't we take the abs value of `daughter_id` as soon as we set it then? e.g. line 668?

scopatz:
Yeah, good point

scopatz:
updated!

gonuke:
(Feeling like a pest but...)  also for the others? `gp2`, `gparent`, `gdaughter`? 

 or should `_to_id()` be fixed... or `nucname.ensdf_to_id()` called by `_to_id()`?

I don't understand any of this data well enough to know where the right place to fix it is, but taking the `abs()` so far downstream seems like something that will bite us in the future

scopatz:
I think `_to_id()` and `ensdf_to_id()` are actually always producing positive numbers. The problem is with `id_from_level()` and some of the levels being wrong.

scopatz:
Yeah, just checked and id_from_level() produces negative number for some specific reason I don't understand.  But they don't seem relevant to us here.  Just `abs()`'d all call sites in this function.

gonuke:
As far as I can tell, `id_from_level()` returns a negative number if it does not find an energy level close to the requested `level`, where currently the definition of close is "3.0" - for some poorly documented reason in some undocumented units...

I agree that we can ignore the negative here, but I can only assume that one is supposed to check the sign and act on it, while also taking the abs() of that result.

I'll merge this, but I can't say that I like it...  unless you think it's worth a deeper look

scopatz:
Yeah, I am not sure that I like it either, but it does fix the problem in the short to medium term

gonuke:
Merging this because it does address #1020, even though the author and reviewer agree that it may not be the best solution.  Created #1023 to capture this for the future.

gonuke:
Thanks @scopatz

PR #1022, opened 2018-04-04
Source sampling all zero bias check

zxkjack123:
This PR adds a check for bias tag used by the source_sampling. When a bias tag is set to all zeros (wrong bias  tag), throw an error. Otherwise, the weight will be NAN and lead to wrong results.

### DIFF
--- a/news/source_sampling_all_zero_bias_check.rst
+++ b/news/source_sampling_all_zero_bias_check.rst
@@ -1,0 +1,15 @@
+**Added:** 
+
+* Check for bias_tag data. Report error when bias tag data are all zero
+
+* A test function to test the check
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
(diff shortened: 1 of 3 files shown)


scopatz:
LGTM!

gonuke:
Just for my info... this failure mode existed before your subvoxel improvements, right?

zxkjack123:
@gonuke I didn't test before the subvoxel improvements, but I think this failure happens to DEFAULT too. Because this failure happens when the bias tag are all zeros. And there wasn't a check before subvoxel improvements. And there are two test functions in `test_source_sampling.py` can get NAN weight. However, because there NAN weight will not be sampled because of it's zero probability, then there will be no problem.

However, when we set all the bias tag to zero (this is of course a wrong input, will not happen for a real problem), it cannot be normalized. We then get all NAN weights. But there is no error reported in this case. So, this PR is a check for this wrong input.

gonuke:
thanks @zxkjack123

PR #1024, opened 2018-04-10
Subvoxel cell num rejection fix

zxkjack123:
This PR changes the mode arrangement for cell rejection of source sampling. Cell rejection should be on when the mode is SUBVOXEL, i.e. 3, 4, 5.


scopatz:
Thanks @zxkjack123!

PR #1025, opened 2018-04-13
GT-CADIS workflow, step 0 for SNILB criteria check

### LINKED ISSUE #987: Tool to test SNILB for each material
[Step 0]
* Move snilb_survey files into pyne
* Modify to get user-supplied values (currently hard-coded):
      * Materials
      * Neutron spectra
      * Irradiation and decay times
      * Mesh
* Add optional step to gtcadis script to perform sweep on all materials 
* Add another option to print max deviation from eta = 1 for each material

moatazharb:
This PR does the following:

- Fixes #987 (the optional step for GT-CADIS workflow)
Changes to gtcadis.py
================
- Read material-laden hdf5 geometry file
- Write ALARA input files (inp, matlib, fluxin) and run ALARA
- Process produced photon source file and calculates SNILB eta value 
-  Save eta values for each material and decay time in a numpy array and text file

Changes to alara.py
==============
- Added calc_eta function 
- Added supporting functions


pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    phtn_src_file = os.path.join(run_dir, "phtn_src")
+    # Two extra zones are needed; one for the whole spectrum
+    # and one for the zero spectrum. Total number of zones = num_n_group + 2
+    num_n_groups += 2
+    _gt_write_inp(run_dir, data_dir, mats, num_n_groups, flux_magnitudes, 
+                  irr_times, decay_times, input_file, matlib_file,
+                  fluxin_file, phtn_src_file, num_p_groups)
+
+    # Run ALARA
+    sub = subprocess.Popen(['alara',input_file], stderr=subprocess.STDOUT,
```
scopatz: PEP8: space after comma.

scopatz: Also, if you want to run a single blocking subprocess, I would recommend using `subprocess.check_call()` or `check_output()`, rather than a Popen() and communicate.

pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+               eta[m, dt] = 1.0
+           elif tot[m, dt] > zero[m, dt]:
+               # tot and sup > background, eta > 0
+               eta[m, dt] = (sup[m, dt] - zero[m, dt]*175)/(tot[m, dt] - zero[m, dt])
+           else:
+               # tot = background and sup != background, eta = inf >> ste = 1e6
+               eta[m, dt] = 1E6
+
+    # Copy  phtn_src file to main directory to be used for Step 2           
+    os.system('cp ' + phtn_src_file + ' step0_phtn_src')
```
scopatz: Please do not use direct system calls like this, since it won't work on all systems.  Python has a few file copying methods, like the copyfile() function in the `shutil` module

pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    if clean:
+        warn("Deleting intermediate files for Step 0")
```
scopatz: This probably should be a `print(msg, file=sys.stdout)`, rather than a warning.

scopatz:
Thanks for putting this in @moatazharb! Just a couple of small changes.

pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
         raise ValueError('Decay time {0} not found in phtn_src file'.format(idc)) 
 
+def _gt_normalize(neutron_spectrum):
+    """
+    Function that normalizes the neutron spectrum for ALARA fluxin file
+    """
+    tol = 1E-8
+    total = float(np.sum(neutron_spectrum))
+    if abs(total - 1.0) > tol:
+        warn("Normalizing neutron spectrum")
```
gonuke: Why a warning?

pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
                 return phtn_src_dc[i]
         raise ValueError('Decay time {0} not found in phtn_src file'.format(idc)) 
 
+def _gt_normalize(neutron_spectrum):
+    """
+    Function that normalizes the neutron spectrum for ALARA fluxin file
+    """
+    tol = 1E-8
+    total = float(np.sum(neutron_spectrum))
+    if abs(total - 1.0) > tol:
```
gonuke: Why not just normalize?  Seems more important that the total is not 0

pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
                 return phtn_src_dc[i]
         raise ValueError('Decay time {0} not found in phtn_src file'.format(idc)) 
 
+def _gt_normalize(neutron_spectrum):
+    """
+    Function that normalizes the neutron spectrum for ALARA fluxin file
+    """
+    tol = 1E-8
+    total = float(np.sum(neutron_spectrum))
```
gonuke: Do you need the `float`?  Isn’t already a float?

pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+        warn("Normalizing neutron spectrum")
+        neutron_spectrum = [x / total for x in neutron_spectrum]
+    return neutron_spectrum
+
+def _gt_write_matlib(mats, filename):
+    """
+    Function that writes matlib file for ALARA run
+    """
+    s = ""
+    for m, mat in enumerate(mats):
```
gonuke: Why enumerate if you don’t need the index?

pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+        f.write(s)
+
+def _gt_write_inp(run_dir, data_dir, mats, num_n_groups, flux_magnitudes,
+                  irr_times, decay_times, input_file, matlib_file,
+                  fluxin_file, phtn_src_file, num_p_groups):
+    """
+    Function that writes ALARA input file
+    """
+    num_zones = len(mats) * (num_n_groups)
+    s = "geometry rectangular\n\nvolume\n"
```
gonuke: This  might be more readable with a long multiline string and a `template` or `format`

pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    matlib_file = os.path.join(run_dir, "matlib")
+    _gt_write_matlib(mats, matlib_file)
+
+    # Write fluxin file
+    fluxin_file = os.path.join(run_dir, "fluxin")
+    fluxes = []
+    for m in range(num_mats):
+        for n in range(num_n_groups):
+            fluxes.append([neutron_spectrum[n] if x ==
+                           n else 0 for x in range(num_n_groups)])
```
gonuke: This appears to do an extra loop that accomplishes nothing.

pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+        os.makedirs(run_dir)    
+    phtn_src_file = _gt_alara(data_dir, mats, neutron_spectrum, flux_magnitudes, 
+                              irr_times, decay_times, num_p_groups, run_dir)
+    # Parse ALARA output
+    sup = np.zeros(shape=(num_mats, num_decay_times))
+    tot = np.zeros(shape=(num_mats, num_decay_times))
+    zero = np.zeros(shape=(num_mats, num_decay_times))
+    with open(phtn_src_file, 'r') as f:
+        i = 0
+        for line in f.readlines():
```
gonuke, line 1217: This block is very difficult to follow.  Better variable names and/or more comments would be helpful

pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+                    tot[m, dt] = row_sum
+                elif n == num_n_groups + 1:
+                    zero[m, dt] = row_sum
+                else:
+                    sup[m, dt] += row_sum
+                i += 1
+                
+    # Claculate eta
+    eta = np.zeros(shape=(num_mats, num_decay_times))
+    for dt, decay_time in enumerate(decay_times):
```
gonuke: why enumerate when you don't use the data from the list?
`for dt in range(len(decay_times)):`

pyne/alara.py
```diff
@@ -890,5 +892,242 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+       for m, mat in enumerate(mats):
```
gonuke: same here

scripts/gtcadis.py
```diff
@@ -1,50 +1,68 @@
...
+    xints: 
+    ymesh: 
+    yints: 
+    zmesh: 
+    zints: 
 
 # Calculate T matrix for each material
 step2:
+    # Path to material laden geometry (hdf5) file for adjoint neutron transport.
+    geom_file:
```
gonuke: Perhaps there should be an input block that includes data common to all steps?

scripts/gtcadis.py
```diff
@@ -122,19 +139,60 @@ def _cards(source):
...
+    """
+    # Get user input from config file
+    geom = cfg2['geom_file']
+    data_dir = cfg2['data_dir']
+    irr_times = str(cfg2['irr_time']).split(' ')
+    decay_times = str(cfg2['decay_time']).split(' ')
+    num_p_groups = cfg1['p_groups']
+    
+    # Define a flat, 175 group neutron spectrum, with magnitude 1E12 [n/s]
+    neutron_spectrum = [1]*175  # will be normalized
```
gonuke: magic number 175

scripts/gtcadis.py
```diff
@@ -122,19 +139,60 @@ def _cards(source):
...
+    flux_magnitudes = [1.75E14] # 1E12*175
```
gonuke: Is this the mag of the total flux?

gonuke (changes requested):
There are many things for readability and maintainability here.  You also need some tests, I think.
I may have more comments once these changes are implemented.

pyne/alara.py
moatazharb: neutron_spectrum is created in step0 in gtcadis.py. It's just a list of 175 1's as [1]*175. I guess this way the function is more generic in case it'll be used for normalization of other spectrums.

pyne/alara.py
moatazharb: In ALARA fluxin file we will have 175 flux blocks per material (one block per energy group). In each block all groups are zero except for one group. This way we can see that the "for n in .. " is equivalent to looping over the block index while in "flux.append .... for x in ..." we loop over the energy group index within that block and fill only one group with a value and the rest with zero.
An additional outer loop will be added later when we include the option of many decay times.

scripts/gtcadis.py
moatazharb: do you mean the number of neutron groups should be input by the user also?

scripts/gtcadis.py
moatazharb: yes. the spectrum is normalized by the total (175) and in alara input we will have a 1.75e14 flux multiplication factor so this way the group fluxes will be 1e12

scripts/gtcadis.py
moatazharb: This should be the last thing to be done I guess. When all steps are done we will have a list of all the inputs that the script needs and make the common ones generally shared by all steps.

gonuke: Or introduce the common block now, and add things once you recognize that they are common.  It will make the work more incremental

scripts/gtcadis.py
gonuke: Eventually, that may be true... but here I mean something similar:
```
n_groups = 175
neutron_spectrum = [1] * n_groups
```

scripts/gtcadis.py
gonuke: with the above change, this could be
```
group_flux_magnitude = 1e12
flux_magnitude = [n_groups * group_flux_magnitude]
```

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
                 return phtn_src_dc[i]
         raise ValueError('Decay time {0} not found in phtn_src file'.format(idc)) 
 
+def _gt_normalize(neutron_spectrum):
+    """
+    Function that normalizes the neutron spectrum for ALARA fluxin file
```
gonuke: docstring should describe inputs and outputs

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    """
+    total = float(np.sum(neutron_spectrum))
+    if total > 0:
+        # Normalize neutron spectrum
+        neutron_spectrum = [x / total for x in neutron_spectrum]
```
gonuke: you should be able to just do `neutron_spectrum /= total`  (assuming it's an numpy array)

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    return neutron_spectrum
 
+def _gt_write_matlib(mats, filename):
+    """
+    Function that writes matlib file for ALARA run
```
gonuke: docstring should describe inputs and outputs

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+def _gt_write_inp(run_dir, data_dir, mats, num_n_groups, flux_magnitudes,
+                  irr_times, decay_times, input_file, matlib_file,
+                  fluxin_file, phtn_src_file, num_p_groups):
+    """
+    Function that writes ALARA input file
+    """
+    # Two extra zones are added per material; one for the whole spectrum
+    # and one for the zero spectrum. 
+    num_zones = len(mats) * (num_n_groups + 2)
+    inp = Template('geometry rectangular \n\n' \
```
gonuke: If you open this string with a triple quote, you can write it more naturally:

```
inp = Template('''
geometry rectangular

volume
$zone
end

mat_loading
$zone_mat
end

$mix
$flux
output zone...
'''
```

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+                   'truncation 1e-7 \n' \
+                   'impurity 5e-6 1e-3 \n' \
+                   'dump_file $run_dir/dump_file \n')
+    # Zones input
+    zone = ""
+    for z in range(num_zones):
+        zone += "    1.0 zone_{0}\n".format(z)
+    # Zones material assignment
+    zone_mat = ""
+    for z in range(num_zones):
```
gonuke, line 1065: might be more readable with
```
mix_num = int(np.floor(z / float(num_n_groups + 2)))
```
that is used as an argument to the format statement

gonuke: Doesn't `np.floor()` guarantee an `int` already?

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    for m, mat in enumerate(mats):
+        mix += "mixture mix_{0}\n".format(m)
+        mix += "    material {0} 1 1\nend\n\n".format(mat.metadata["name"])
+    # Flux input
+    flux = ""
+    for i, flux_magnitude in enumerate(flux_magnitudes):
+        flux += "flux flux_{0} {1} {2} 0 default\n".format(i, fluxin_file, flux_magnitude)
+    # Photon energy bin structure
+    # 24 bin structure
+    p_groups = ""
```
gonuke: all version of this line are initializes with the same thing, so go ahead and intiailize them that way.

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    if num_p_groups == 24:
+        p_groups += "    photon_source {0}/fendl2.0bin {1} 24 1.00E4 2.00E4 5.00E4 1.00E5\n".format(data_dir, phtn_src_file)
```
gonuke: Store these list of group bounds as lists somewhere and then use those lists here.  You might want a function to handle the formatting of N values per line... or a function to insert newlines at a certain length.  (ALARA doesn't actually care, so you don't need to break the line at all.)

pyne/alara.py
gonuke: Why not
```
tmp_spectrum = np.zeros(num_n_groups)
tmp_spectrum[n] = neutron_spectrum[n]
fluxes.append(tmp_spectrum[n])
```

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+
+    # Write fluxin file
+    fluxin_file = os.path.join(run_dir, "fluxin")
+    fluxes = []
+    for m in range(num_mats):
+        for n in range(num_n_groups):
+            fluxes.append([neutron_spectrum[n] if x ==
+                           n else 0 for x in range(num_n_groups)])
+        fluxes.append(neutron_spectrum) # total spectrum
+        fluxes.append([0]*num_n_groups) # blank spectrum
```
gonuke: `fluxes.append(np.zeros(num_n_groups))`

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    zero = np.zeros(shape=(num_mats, num_decay_times, num_p_groups + 1))
+    with open(phtn_src_file, 'r') as f:
+        # Initiate a block number
+        i = 0
+        for line in f.readlines():
+            l = line.split()
+            if l[0] == "TOTAL" and l[1] != "shutdown":
+                row = [float(x) for x in l[3:]]
+                row_sum = np.sum(row)
+                # Material index will change every (num_n_groups+2)*num_decay_times)
```
gonuke: readability: introduce `entries_per_material = (num_n_groups + 2) * num_decay_times`

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+                m = int(np.floor(float(i)/((num_n_groups+2)*num_decay_times)))
+                dt = i % num_decay_times
+                n = int(np.floor(i/float(num_decay_times))) % (num_n_groups + 2)
+                if n == num_n_groups:
+                    tot[m, dt, -1] = row_sum
```
gonuke: `tot[m, dt, :] = [row row_sum]` (or something like that?)

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+                    zero[m, dt, -1] = row_sum
+                    zero[m, dt, :-1] = row
+                else:
+                    sup[m, dt, -1] += row_sum
+                    sup[m, dt, :-1] += row
+                i += 1
+                
+    # Claculate eta
+    eta = np.zeros(shape=(num_mats, num_decay_times, num_p_groups))
+    eta_sum = np.zeros(shape=(num_mats, num_decay_times))
```
gonuke: why not just use an "extra" p group again to save duplicating the loops

pyne/alara.py
```diff
@@ -890,5 +893,296 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    with open(phtn_src_file, 'r') as f:
+        # Initiate a block number
+        i = 0
+        for line in f.readlines():
+            l = line.split()
+            if l[0] == "TOTAL" and l[1] != "shutdown":
+                row = [float(x) for x in l[3:]]
+                row_sum = np.sum(row)
+                # Material index will change every (num_n_groups+2)*num_decay_times)
+                m = int(np.floor(float(i)/((num_n_groups+2)*num_decay_times)))
```
gonuke: could you take advantage of the fact that you can reshape numpy arrays to avoid all this index math?  I've never done it myself, but seen it used.  Something like:
```
tot = np.zeros(shape=(num_mats * num_decay_times, num_p_groups + 1))
with open....
     i = 0
     for line in f.readlines():
       l = line.split()
       if l[0] == "TOTAL" ....
           row = [float(x) for x in l[3:]]
           tot[i, :] = [row, np.sum(row)]
```
Then, when you're done, you can do `tot.reshape(num_mats, num_decay_times, num_p_groups + 2)`

Of course - you still need all the other arrays and the conditionals...

scripts/gtcadis.py
gonuke: Is this `geom_file` different from the step1 `geom_file`? One is for neutrons and one is for photons?  Maybe name them differently?

scripts/gtcadis.py
```diff
@@ -122,19 +139,86 @@ def _cards(source):
...
+    """
+    # Get user input from config file
+    geom = cfg2['geom_file']
+    data_dir = cfg2['data_dir']
+    irr_times = str(cfg2['irr_time']).split(' ')
+    decay_times = str(cfg2['decay_time']).split(' ')
+    num_p_groups = cfg1['p_groups']
+    
+    # Define a flat, 175 group neutron spectrum, with magnitude 1E12 [n/s]
+    num_n_groups = 175
```
gonuke: shouldn't this be part of the cfg file?

scripts/gtcadis.py
```diff
@@ -122,19 +139,86 @@ def _cards(source):
...
+    flux_magnitudes = [flux_magnitude * num_n_groups] # 1E12*175
+
+    # Get materials from geometry file
+    ml = MaterialLibrary(geom)
+    mats = list(ml.values())
+
+    eta_elements = True
+    if eta_elements:
+        # Calculate eta for each element in the material library
+        elements = []
```
gonuke: This might be better as a set

pyne/alara.py
gonuke: Could possibly do this with the `n_group` dimension as well...

gonuke (changes requested):
I recommend using numpy arrays for all of your list-like data structures that contain numbers - it will likely make many of your operations a little easier

pyne/alara.py
moatazharb: done

pyne/alara.py
moatazharb: done

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+dump_file $run_dir/dump_file
+    """)
+    
+    # Zones input
+    zone = ""
+    for z in range(num_zones):
+        zone += "    1.0 zone_{0}\n".format(z)
+    # Zones material assignment
+    zone_mat = ""
+    for z in range(num_zones):
```
gonuke, line 1065: why not combine this with the previous loop?

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
                 return phtn_src_dc[i]
         raise ValueError('Decay time {0} not found in phtn_src file'.format(idc)) 
 
+def _gt_normalize(neutron_spectrum):
```
gonuke: This is a pretty generic function that might be useful across PyNE (and may already exist elsewhere).

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+        neutron_spectrum /= total
+    return neutron_spectrum
+
+def _gt_write_matlib(mats, filename):
+    """
+    Function that writes ALARA matlib file
+
+    Parameters
+    ----------
+    mats: list
```
gonuke: why not use a `MaterialLibrary` instead of a list of `Materials` and then implement methods in that class to write a whole library in the right format?

gonuke: In it's current form, this hardly needs to be a function.  Instead of building the string `s`, this could be:

```
with open(filename, 'w') as f:
   for mat in mats:
      f.write(mat.alara() + "\n")
```

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    """
+    s = ""
+    # Split flux array into a number of arrays equal to the number
+    # of zones in ALARA input
+    fluxes = np.split(fluxes, num_zones, axis = 0)
+    for j in range(num_zones):
+        flux = np.fliplr(fluxes[j].reshape(1, fluxes[j].size))
+        # Write flux groups in reverse order
+        for i in range(flux.size):
+            s += "{0:.6E} ".format(flux[0, i])
```
gonuke: Seems as though you could resize the numpy array here to take care of this.

Also, ALARA does not require the flux file to only have 6 entries per line.

cadangelo:
I think all requested changes were made.. is there anything I'm missing?

gonuke:
Sorry -- I'm stuck in the middle of a long review of this...

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    p_groups: numpy array
+        Array of photon energy group bounds
 
+    Returns
+    -------
+    p_string: str
+        Formatted photon energy groups structure 
+    """
+    p_string = ""
+    for i, p in enumerate(p_groups):
```
gonuke: This is basically the same formatting as used elsewhere for the flux file - seems like a good place for a method to take care of it.

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+
+    # Write geom file
+    input_file = os.path.join(run_dir, "inp")
+    phtn_src_file = os.path.join(run_dir, "phtn_src")
+    _gt_write_inp(run_dir, data_dir, mats, num_n_groups, flux_magnitudes, 
+                  irr_times, decay_times, input_file, matlib_file,
+                  fluxin_file, phtn_src_file, num_p_groups, p_bins)
+
+    # Run ALARA
+    sub = subprocess.check_output(['alara', input_file], stderr=subprocess.STDOUT)
```
gonuke, line 1163: Do we actually want to run ALARA, or just set it all up?  It could be a long-running job that we don't want to wait for - and it may require large memory. etc...

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    # Flux input
+    flux = ""
+    for i, flux_magnitude in enumerate(flux_magnitudes):
+        flux += "flux flux_{0} {1} {2} 0 default\n".format(i, fluxin_file, flux_magnitude)
+    # Photon energy bin structure
+    p_groups = "    photon_source {0}/fendl2.0bin {1} {2}\n    {3}\n".format(data_dir, phtn_src_file,
+                                                        num_p_groups, _gt_format_p_groups(p_bins))
+    # Irradiation schedule input
+    irr = ""
+    for i, irr_time in enumerate(irr_times):
```
gonuke: Does this require that the length of `irr_times` and `flux_magnitudes` is the same?

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    num_irr_times = len(irr_times)
+
+    # Write matlib file
+    matlib_file = os.path.join(run_dir, "matlib")
+    _gt_write_matlib(mats, matlib_file)
+
+    # Write fluxin file
+    fluxin_file = os.path.join(run_dir, "fluxin")
+    fluxes = np.array([])
+    for m in range(num_mats):
```
gonuke: the block of fluxes for each material is identical, so why form all of them?  Why not one block of fluxes that you use `num_mats` times in `write_fluxin`?

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+        for n in range(num_n_groups):
```
gonuke: not sure why you need to assemble all these fluxes here since the only place it is used is in `_gt_write_fluxin`.  Just do some of this (without needing to build a bug array) in that method.

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    num_mats = len(mats)
+    num_decay_times = len(decay_times)
+    
+    # Run ALARA
+    if not os.path.exists(run_dir):
+        os.makedirs(run_dir)    
+    phtn_src_file = _gt_alara(data_dir, mats, neutron_spectrum, flux_magnitudes, 
+                              irr_times, decay_times, num_p_groups, p_bins, run_dir)
+    # Parse ALARA output
+    sup = np.zeros(shape=(num_mats, num_decay_times, num_p_groups + 1))
```
gonuke: Probably makes sense to have a separate method to parse ALARA results

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+            if l[0] == "TOTAL" and l[1] != "shutdown":
+                row = np.array([float(x) for x in l[3:]])
+                row_sum = np.sum(row)
+                # Material index will change every (num_n_groups + 2) * num_decay_times
+                entries_per_material = (num_n_groups + 2) * num_decay_times
+                m = int(np.floor(float(i) / entries_per_material))
+                dt = i % num_decay_times
+                n = int(np.floor(i/float(num_decay_times))) % (num_n_groups + 2)
+                if n == num_n_groups:
+                    tot[m, dt, :] = np.hstack([row, row_sum])
```
gonuke: isn't `row_sum` a scalar?  Can't you just append it?

gonuke: Also - it appears that you always append it, and the only difference is where to assign it. - move the append outside the conditional

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    phtn_src_file = _gt_alara(data_dir, mats, neutron_spectrum, flux_magnitudes, 
+                              irr_times, decay_times, num_p_groups, p_bins, run_dir)
+    # Parse ALARA output
+    sup = np.zeros(shape=(num_mats, num_decay_times, num_p_groups + 1))
+    tot = np.zeros(shape=(num_mats, num_decay_times, num_p_groups + 1))
+    zero = np.zeros(shape=(num_mats, num_decay_times, num_p_groups + 1))
+    with open(phtn_src_file, 'r') as f:
+        # Initiate a block number
+        i = 0
+        for line in f.readlines():
```
gonuke, line 1217: couldn't this be a set of loops over m, n and dt, in which you throw away lines that don't match your "TOTAL" & "! shutdown" criteria?
It would be more readable.

scripts/gtcadis.py
```diff
@@ -122,31 +142,132 @@ def _cards(source):
...
+
+    # Get materials from geometry file
+    mat_lib = MaterialLibrary(geom)
+    mats = list(mat_lib.values())
+
+    eta_elements = True
+    if eta_elements:
+        # Calculate eta for each element in the material library
+        elements = Set([ ])
+        for m, mat in enumerate(mat_lib.keys()):
```
gonuke: you don't use `mat`

scripts/gtcadis.py
```diff
@@ -122,31 +142,132 @@ def _cards(source):
...
+    decay_times = str(cfg2['decay_time']).split(' ')
+    
+    # Define a flat, 175 group, neutron spectrum with magnitude 1E12 [n/cm^2.s]
+    group_flux_magnitude = 1.0E12
+    neutron_spectrum = group_flux_magnitude * np.ones(num_n_groups)
+    flux_magnitudes = np.array([np.sum(neutron_spectrum)])
+
+    # Get materials from geometry file
+    mat_lib = MaterialLibrary(geom)
+    mats = list(mat_lib.values())
```
gonuke: this conversion between a dictionary and a list is puzzling

scripts/gtcadis.py
```diff
@@ -122,31 +142,132 @@ def _cards(source):
...
+        elements = Set([ ])
+        for m, mat in enumerate(mat_lib.keys()):
+            # Collapse elements in the material
+            mat_collapsed = mats[m].collapse_elements([])
+            element_list = mat_collapsed.comp.keys()
+            elements.update(element_list)
+        # Create PyNE material object per elements    
+        for element in elements:
+            mat_element = Material({element: 1.0})
+            mat_element.metadata['name'] = 'mat:%s' %nucname.name(element)
```
gonuke: use `format` for consistenct?

gonuke (changes requested):
There are a lot of unusual data structure and logic choices that make it hard to follow some of this code

pyne/alara.py
```diff
@@ -890,5 +893,365 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    fluxes = np.split(fluxes, num_zones, axis = 0)
+    for j in range(num_zones):
+        flux = np.fliplr(fluxes[j].reshape(1, fluxes[j].size))
+        # Write flux groups in reverse order
+        for i in range(flux.size):
+            s += "{0:.6E} ".format(flux[0, i])
+            if (i + 1) % 6 == 0:
+                s += '\n'
+        s += '\n\n'
+    with open(fluxin_file, 'w') as f:
```
gonuke: Do the actual writing to the file from the place where you call this function

moatazharb:
With the last commit all the previous comments have been addressed. Besides the comments, the calculation of eta per element has been separated from that of materials present in the geometry as it interferes with impurity threshold on ALARA input and causes activation chains/trees to be treated differently in comparison with the case of only materials. The working of the script has been tested and development of step 2 (calculation of T matrix) is near its end and should be ready soon for a PR.

gonuke:
Please rebase this so that it passes tests and I'll merge it

gonuke (approved):

moatazharb:
Wrong rebase!

moatazharb:
Created a new pull request that include the most updated version of files alara.py and gtcadis.py per the reviews/comments that were done in this pull request. Closing this one to avoid the hustle of redoing a rebase on a wrong pyne/develop.

PR #1026, opened 2018-04-18
Remove alara params decay times

zxkjack123:
Users need to define decay times in both `config.ini` and `alara_params.txt` for running r2s before this PR.
This PR keeps the decay times definition in "config.ini" and remove them from the "alara_params.txt".  
By the way, some names of `alara_geom` are changed into `alara_inp`.
Major changes:
- Remove decay times definition from `alara_params.txt`
- Change`alara_geom` to `alara_inp` at some places

### DIFF
--- a/pyne/r2s.py
+++ b/pyne/r2s.py
@@ -14,7 +14,7 @@ warn(__name__ + " is not yet QA compliant.", QAWarning)
 def irradiation_setup(flux_mesh, cell_mats, alara_params, tally_num=4,
                       geom=None, num_rays=10, grid=False, flux_tag="n_flux",
                       fluxin="alara_fluxin", reverse=False,
-                      alara_inp="alara_geom", alara_matlib="alara_matlib",
+                      alara_inp="alara_inp", alara_matlib="alara_matlib",
                       output_mesh="r2s_step1.h5m", output_material=False,
                       sub_voxel=False):
     """This function is used to setup the irradiation inputs after the first
--- a/pyne/r2s.py
+++ b/pyne/r2s.py
@@ -64,7 +64,7 @@ def irradiation_setup(flux_mesh, cell_mats, alara_params, tally_num=4,
     output_material : bool, optional
         If true, output mesh will have materials as determined by
         dagmc.discretize_geom()
-    decay_times: list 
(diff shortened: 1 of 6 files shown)


pyne/r2s.py
```diff
@@ -14,9 +14,9 @@
 def irradiation_setup(flux_mesh, cell_mats, alara_params, tally_num=4,
                       geom=None, num_rays=10, grid=False, flux_tag="n_flux",
                       fluxin="alara_fluxin", reverse=False,
-                      alara_inp="alara_geom", alara_matlib="alara_matlib",
+                      alara_inp="alara_inp", alara_matlib="alara_matlib",
                       output_mesh="r2s_step1.h5m", output_material=False,
-                      sub_voxel=False):
+                      decay_times=['1 s'], sub_voxel=False):
```
scopatz: You should not put mutable containers (such as a list) as the default value of a kwarg.  Please use a tuple instead.

pyne/r2s.py
```diff
@@ -111,6 +113,14 @@ def irradiation_setup(flux_mesh, cell_mats, alara_params, tally_num=4,
     record_to_geom(m, cell_fracs, cell_mats, alara_inp, alara_matlib,
                    sub_voxel=sub_voxel)
 
+    # write decay times into alara_inp
+    decay_str = 'cooling\n'
+    for dc in decay_times:
+        decay_str = ''.join([decay_str,'    ', dc, '\n'])
```
scopatz: PEP8: single space after comma

scopatz:
Thanks for putting this in.  Generally looks good.  Couple of small issues.

gonuke:
Was this actually a bug previously?  it seems that the decay times wouldn't necessarily match between different steps of R2S before...

zxkjack123:
@gonuke . No, it wasn't a bug. But users had to define the cooling time two times, in both `alara_params.txt` and `config.ini`. However, they are actually the same information. So, I removed the definition of cooling time in `alara_params.txt`. Therefore, we can define the cooling time only once now (in `config.ini`).

gonuke:
Thanks @zxkjack123

PR #1027, opened 2018-04-19
Update r2s doc

zxkjack123:
As the subvoxel r2s is available for pyne now. The R2S documentation should be updated. This PR makes some changes to current r2s docs, adding descriptions according to the updated source_sampling method.


scopatz:
Thanks!

Issue #1032, opened 2018-05-18
Pyne Tutorial Nuclear Data: Half Life Plot


KennellyT:
Hey,

I was going through the PYNE tutorial on nuclear data and I noticed that the Half life plot would not plot. I found the issue was that the 'anums' variable was mapping a function to a list but it would not accept the type inside the list for some reason. I made another variable called 'NuclearName' that made all the  data in f.root.decay.level_list[:]['nuc_id'] an int as seen below:

NuclearName= list(map(lambda x:int(x),f.root.decay.level_list[:]['nuc_id']))

I also changed the variable 'anums' so that it would map nucname.anum to NuclearName, the new list of Nuclear Names (Nuclear ID numbers) as int(s). 'anums' is seen below:
anums = map(nucname.anum, NuclearName)

I also noted that the plot:
plt.semilogy(anums,  f.root.decay.level_list[:]['half_life'], 'ko', ms=1, alpha=0.3)

attempted to plot anums versus the half life but python would not accept plotting a map against a list so I made 'anums' into a list and plotted this as: 

plt.semilogy(list(anums1),  f.root.decay.level_list[:]['half_life'], 'ko', ms=1, alpha=0.3)

I attached figures (1,2) as the prior code and errors, figure 3 as the updated code, and figure 4 as the output. 
I was wondering if there is any stylistic or appearance changes I should make prior to pull request?
Thank you!

![half_life_orginal_5_18](https://user-images.githubusercontent.com/31637359/40258009-cbd29194-5ab5-11e8-82c9-5998a83bacea.png)
![hl_5_18](https://user-images.githubusercontent.com/31637359/40258015-cef1ded4-5ab5-11e8-8b8e-99861de3e679.png)
![half_life_plot_update](https://user-images.githubusercontent.com/31637359/40258028-d42d9640-5ab5-11e8-87a8-b5f147d615bc.png)
![image](https://user-images.githubusercontent.com/31637359/40258149-3cd15e0c-5ab6-11e8-834e-6c2d82aaeaae.png)

scopatz:
I think a pull request here would be great!

KennellyT:
Great! I have just submitted the pull request.

PR #1035, opened 2018-05-19
Update 03-nuclear-data.ipynb

KennellyT:
Changed/fixed half life plot.  
See: [Pyne Tutorial Nuclear Data: Half Life Plot #1032](https://github.com/pyne/pyne/issues/1032)


scopatz:
Thanks for the fix!

PR #1039, opened 2018-05-23
Material_library (examples)

KennellyT:
Converted print to Python 3 version of print
In Python 3, ['KeysView' object does not support indexing](https://stackoverflow.com/questions/45426579/typeerror-keysview-object-does-not-support-indexing)
so I made mats.keys() into a list so it could be indexed:
`list(mats.keys())[:10]`
I am also in the process of collecting/storing a data set composed of the thermophysical properties of matter. If it can be integrated into pyne easily, I will put a pull request on the final product.


scopatz:
Thanks for the fixes @KennellyT!

PR #1042, opened 2018-05-30
Facet voxel traverse tool

zxkjack123:
This PR adds some functions to perform voxel traverse for ray tracing. And then add a function to calculate the voxels intersect with given facet based on the ray tracing.

Based on [A Fast Voxel Traversal Algorithm for Ray Tracing](http://www.cse.yorku.ca/~amana/research/grid.pdf), a modified ray traverse method was used to calculate the voxels intersect with a given ray (line segment). Several rays are used to calculate the voxels that intersect with the facet. A series of rays are constructed based on both the facet and the mesh grid to make sure all the wanted voxels will be found. Combining all the found voxels for each ray, we get the voxels intersect with the facet. 

These voxels will be used to construct the bounding volume for source sampling. But currently these functions are separate from MOAB. Interface of MOAB with PyNE is still in need.

Major changes:

- Functions used to calculate the voxels that intersect with given ray.

- Functions used to calculate the voxels that intersect with given facet (triangle).

- Test functions for the function of ray/facet voxel traverse.


pyne/utils.py
```diff
@@ -190,3 +191,534 @@ def remove(path):
...
         pass
+
+##################################
+### Facet voxel traverse tools ###
+##################################
+
+
+class Point(object):
+    """
+    Class point.
```
scopatz: Can you please be more descriptive here.

pyne/utils.py
```diff
@@ -190,3 +191,534 @@ def remove(path):
...
+    """
+    def __init__(self, x=0.0, y=0.0, z=0.0):
+        self.x = x
+        self.y = y
+        self.z = z
+
+    def __str__(self):
+        """
+        Print point according to the coordinate.
```
scopatz: This doesn't actually print

pyne/utils.py
```diff
@@ -190,3 +191,534 @@ def remove(path):
...
+    d_x = end.x - start.x
+    d_y = end.y - start.y
+    d_z = end.z - start.z
+    dist = math.sqrt(d_x * d_x + d_y * d_y + d_z * d_z)
+    v_x = d_x / dist
+    v_y = d_y / dist
+    v_z = d_z / dist
+    return [v_x, v_y, v_z]
+
+def _find_next_grid_1D(cor, v_1D, bound):
```
scopatz: PEP8: variable and function names should not have uppercase characters

scopatz:
Thanks for putting this in @zxkjack123 - other than a couple of minor issues, this looks good to me!

zxkjack123:
@scopatz I have made some changes according to your comment.

pyne/utils.py
```diff
@@ -190,3 +191,535 @@ def remove(path):
         remove_tree(path, verbose=False)
     else:
         pass
+
+##################################
+### Facet voxel traverse tools ###
+##################################
```
scopatz: hmm it is a little weird to me that these are all in `utils.py` rather than their own module, since this seems like such a self-contained piece of code.  Is there a reason for them living in `utils.py` that I am not seeing?

zxkjack123: These functions are indeed a self-contained piece of code. However, among all these functions, only one function, the `_facet_voxel_traverse`, will be called. 

I am not sure where should I put all these functions. I put them in `utils.py` becasue they just provide an useful function as a tool.

scopatz: I think you should make a new module that has an appropriate name for this capability

zxkjack123:
@scopatz I have moved the voxel traversal algorithm function to module `vta.py`.

news/facet_voxel_traverse_tool.rst
```diff
@@ -0,0 +1,17 @@
+**Added:**
+
+* Functions used to calculate the voxels that intersect with given ray.
```
scopatz: Can you add that these are in the `vta` module please?

pyne/vta.py
```diff
@@ -0,0 +1,535 @@
+from __future__ import division
```
scopatz, line 17: Please provide a module level docstring

tests/test_vta.py
```diff
@@ -0,0 +1,49 @@
...
+
+import nose
+
+from nose.tools import assert_equal, assert_not_equal, assert_raises, raises, \
+                       assert_in, assert_true
+
+from numpy.testing import assert_array_equal, assert_array_almost_equal
+
+from pyne import vta
+import numpy as np
```
scopatz, line 11: PEP8: numpy imports should come before pyne imports.  Please move this line up.

scopatz:
Thanks for updating! Can you also please add a documentation hook to the `docs/pyapi/` directory? It should be called `vta.rst`. You can just copy one of the other files in that directory over and then edit the contents to say vta instead.

katyhuff:
This seems like maybe it's ready to merge. Thoughts @zxkjack123 ?

zxkjack123:
Thanks @katyhuff. I will let this PR to be discussed in a group meeting in next one or two weeks. I'd like to deal with this PR after the discussion in the group meeting.

pyne/vta.py
```diff
@@ -0,0 +1,551 @@
+"""Purpose:
+
+This module provides a modified ray traverse method, which is modifed based on
+http://www.cse.yorku.ca/~amana/research/grid.pdf, to calculate the voxels
```
pshriwise: Some minor grammar here "to calculate which voxels intersect with a given ray (line segment)".

pyne/vta.py
```diff
@@ -0,0 +1,551 @@
...
+intersect with a given ray (line segment).
+Several rays are used to calculate the voxels that intersect with the facet.
+A series of rays are constructed based on both the facet and the mesh grid to
+make sure all the wanted voxels will be found. Combining all the found voxels
+for each ray, we get the voxels intersect with the facet.
```
pshriwise: "we get the voxels intersect with the facet" --> "we get the voxels intersecting with the facet"

pyne/vta.py
```diff
@@ -0,0 +1,551 @@
...
+    return dist
+
+def _is_point_in_mesh(bounds, point):
+    """
+    Check whether a point is in the mesh.
+
+    Parameters:
+    -----------
+    bounds : list
+        Boundaries. List of the boundaries along x, y, and z directions.
```
pshriwise: This is a nested list, expected to be of the format 

`[[x_min, x_max],[y_min, y_max],[z_min,z_max]]`. 

I could see myself also expecting this to work with a nested list of the form 

`[[x_min, y_min, z_min],[x_max, y_max, z_max]]`

I think it's worth being specific about the expected structure of this input parameter in the documentation.

pyne/vta.py
```diff
@@ -0,0 +1,551 @@
...
+    
+    Parameters:
+    -----------
+    tp : Point
+        Temporary point position.
+    end : Point
+        End point
+    v : list
+        Direction vector from start to end: [v_x, v_y, v_z].
+    bounds : list
```
pshriwise: Again, I think this could be a little more descriptive in terms of what is expected to be in this set of nested lists.

pyne/vta.py
```diff
@@ -0,0 +1,551 @@
...
+    n_y_grid = _find_next_grid_1d(tp.y, v[1], bounds[1])
+    if v[0] == 0.0:
+        t_max_x = float('inf')
+    else:
+        t_max_x = (n_x_grid - tp.x) / v[0]
+    if v[1] == 0.0:
+        t_max_y = float('inf')
+    else:
+        t_max_y = (n_y_grid - tp.y) / v[1]
+    return [t_max_x, t_max_y]
```
pshriwise: The doc strings indicate that three directions will be returned, but only two are found here.

pyne/vta.py
```diff
@@ -0,0 +1,551 @@
...
+    --------
+    x_idx : int
+        Modified x_idx.
+    y_idx : int
+        Modified y_idx.
+    z_idx : int
+        Modified z_idx.
+    tp : Point
+        Point after moving along the direction of v.
+    t_temp : float
```
pshriwise: Does this mean an updated length of the ray?

pyne/vta.py
```diff
@@ -0,0 +1,551 @@
...
+        z_idx += 1
+    tp.x += t_min * v[0]
+    tp.y += t_min * v[1]
+    tp.z += t_min * v[2]
+    t_temp += t_min
+    return x_idx, y_idx, z_idx, tp, t_temp
+
+def _move_to_boundary(tp, t_min, v):
+    """
+    Move the point to the nearest boundary when Initialize the problem.
```
pshriwise: "Initialize" --> "initializing"

pyne/vta.py
```diff
@@ -0,0 +1,551 @@
...
+    Return:
+    -------
+    tp : Point
+        Modified point. This point lies on the nearest boundary. Therefore, we
+        moved point from outside the mesh to the mesh boundary (or the extension
+        of the boundary, which still outside the mesh).
+    """
+
+    # calculate the new coordinate
+    tp.x += t_min * v[0]
```
pshriwise: This only updates two dimensions?

pyne/vta.py
```diff
@@ -0,0 +1,551 @@
...
+            # the point is in the mesh
+            # calculate the voxel id
+            x_idx = _find_voxel_idx_1d(bounds[0], tp.x, v[0])
+            y_idx = _find_voxel_idx_1d(bounds[1], tp.y, v[1])
+            z_idx = _find_voxel_idx_1d(bounds[2], tp.z, v[2])
+            if -1 in (x_idx, y_idx, z_idx):
+                # point outside the mesh
+                return voxel_list
+            # add current voxel
+            voxel_list.append([x_idx, y_idx, z_idx])
```
pshriwise: It might be worth considering tuples here. You'll be looping over these values later (presumably) and won't want them to accidentally get changed.

pshriwise (changes requested):
This looks good @zxkjack123! It's a considerable amount of work and you've made it a very modular piece of code. It was very enjoyable to review 🏆 

I have a couple design comments a couple concerns/questions:

  - The `Point` class you've created comes in handy in so many places here! I think it would be worth considering the addition of other classes (such as `Direction` and `Voxel` classes) which will reduce the number of arguments you have to pass from function to function and make the code more clean. Additionally, it will make documentation easier! Instead of having to explain (perhaps repeatedly) that an argument is "a list containing a direction vector u,v,w" you can simply claim it was a `Direction` class object and refer them to the class documentation :)

  - There are some places I've pointed out where I think only 2 of 3 dimensions are being handled. I coule be missing something there but it seems odd and the unit tests are structured in such a way that if only the x & y dimensions are handled they might not catch those problems - so I'm wary of that

Overall, this looks very good. I'll take another look once some of these comments have been addressed.

pyne/vta.py
zxkjack123: I will add the expected structure here. And the expected structure will be:
```
        [[x_min, ..., x_max], [y_min, ..., y_max], [z_min, ..., z_max]]
 ```
The boundaries in each direction could be different (iints, jints and kints in mesh are different). Therefore, the format like:
```
[[x_min, z_min, z_min], ..., [x_max, y_max, z_max]]
``` 
will not work. 

Though we can construct a nested list that contains only the min and max boundaries of a mesh, it is not necessary.

File: pyne/vta.py (hunk 12 above)
Comment (zxkjack123) [reply pyne/vta.py]: Oops, I forgot to update the code here. It should return data in three directions. This will be fixed in next commit.

File: pyne/vta.py (hunk 10 above)
Comment (pshriwise) [reply pyne/vta.py]: True. I was more suggesting that based on the documentation a user (or future developer) might not know how to structure the list if they wanted to use this function.

File: pyne/vta.py (hunk 13 above)
Comment (zxkjack123) [reply pyne/vta.py]: Yes, it is an updated length of the ray. The distance between the start point and updated point position.

File: pyne/vta.py (hunk 14 above)
Comment (zxkjack123) [reply pyne/vta.py]: It will be  fixed in the next commit.

File: pyne/vta.py (hunk 15 above)
Comment (zxkjack123) [reply pyne/vta.py]: Sorry again. Another place I forgot to update. It should update three dimensions.

**Comment by zxkjack123** [comment]
@pshriwise Thank you very much for your comments. You have pointed out that I forgot to update some places of the code, leaving the z direction unchanged. And the unit tests are designed only in x and y directions, therefore, these bugs are not detected by the tests. I will add a 3D unit test to this PR to make sure this PR gets a correct answer first. I will update this PR when I finish this step.

Then I will restructure the code, adding some classes such as direction and voxel to make the code cleaner and easier to read.

File: pyne/vta.py [hunk 17]
```diff
@@ -0,0 +1,566 @@
...
+    voxel_list: list
+        List of voxels that intersect with the facet.
+    """
+    min_step = _cal_min_grid_step(bounds)
+    rays = _create_rays_from_triangle_facet(A, B, C, min_step)
+    voxel_list = []
+    for ray in rays:
+        temp_voxel_list = _ray_voxel_traverse(bounds, ray[0], ray[1])
+        # merge two lists
+        voxel_list = _merge_voxel_list(voxel_list, temp_voxel_list)
```
gonuke: You can use a `set` and then it will take care of duplicates automatically

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+    -------
+    rays : list of rays, [start, end] pairs
+        List of the created rays. Used to represent the facet.
+    """
+    a = _distance(B, C)
+    b = _distance(A, C)
+    c = _distance(A, B)
+    min_edge = min([a, b, c])
+    if a == min_edge:
+        # divide BC
```
gonuke:
```
if a == min_edge:
    target = A
    source = (B,C)
elif b == min_edge:
    target = B
    source = (A,C)
else
    target = C
    source = (A,B)
insert_points = _divide_edge(source, min_step)
rays = _create_rays_from_points(target, insert_points)
```

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+    step : float
+        The minimum grid step. Divider used to divide the line segment.
+
+    Return:
+    -------
+    intert_points : list of points
+        List of points. These points are located on the line segment from start
+        to end.
+    """
+    dist = _distance(start, end)
```
gonuke:
```
num_segments = floor(dist/step) + 2
x_list = numpy.linspace(start.x, end.x,num_segments)
y_list = numpy.linspace(start.y,end.y,num_segments)
z_list - numpy.linspace(start.z,end.z,num_segments)
insert_points = []
for (x,y,z) in zip(x_list,y_list,z_list):
    insert_points.append(Point(x,y,z))
```

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+
+    Return:
+    -------
+    min_step: float
+        The minimum grid step.
+    """
+    min_step = float('inf')
+    for i in range(len(bounds)):
+        for j in range(len(bounds[i])-1):
+            step = bounds[i][j+1] - bounds[i][j]
```
gonuke: add 3rd dimension

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+    bounds : list
+        Boundaries of mesh grid.
+
+    Return:
+    -------
+    min_step: float
+        The minimum grid step.
+    """
+    min_step = float('inf')
+    for i in range(len(bounds)):
```
gonuke: check each dimension/direction independently because the grids are structured

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+        for j in range(len(bounds[i])-1):
+            step = bounds[i][j+1] - bounds[i][j]
+            if min_step > step:
```
gonuke: `min_step = min(min_step,step)`

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+    -----------
+    bounds : list
+        Boundaries of mesh grid.
+
+    Return:
+    -------
+    min_step: float
+        The minimum grid step.
+    """
+    min_step = float('inf')
```
gonuke:
```
for i in range(len(bounds)-1):
    min_step = min(min_step,bounds[i+1]-bounds[i])
``

File: pyne/vta.py (hunk 21 above)
Comment (gonuke) [inline pyne/vta.py]:
```
for bound_list in bounds:
   for bound_idx in range(len(bound_list)-1):
        min_step = min(min_step,bound_list[bound_idx+1]-bound_list[bound_idx])
```

Comment (gonuke) [reply pyne/vta.py]: maybe too clever:
```
for (upper,lower) in zip(bound_list[1:],bound_list[:-1]):
    min_step = min(min_step,upper-lower)
```

File: pyne/vta.py [hunk 24]
```diff
@@ -0,0 +1,566 @@
...
+                tp.z += t_min * v[2]
+    return voxel_list
+ 
+def _cal_min_grid_step(bounds):
+    """
+    Calculate the minimum grid step. Used for scan.
+
+    Parameters:
+    -----------
+    bounds : list
```
gonuke: elaborate on the structure of `bounds`

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+            t_maxs = _cal_max_travel_length_in_current_voxel(tp, end, v, bounds)
+            x_idx, y_idx, z_idx, tp, t_temp = _move_to_next_voxel(
+                x_idx, y_idx, z_idx, tp, t_temp, t_maxs, v)
+        else:
+            # Check whether the ray intersecs the mesh. Calculate the maxinum travel
+            # length of the ray to the outside boundaries of the mesh. If all the
+            # length are negtive, then there is no intersection, return. Ohterwise,
+            # choose the minimum length and move the point to that position.
+            # left, right, back, front, down, up
+            t_maxs = [0.0] * 4
```
gonuke:
```
for dir in [0,1,2]:
   if v[dir] == 0.0:
     t_maxs[dir*2] = float('inf')
     t_maxs[dir*2 +1]=float('inf')
   else:
      t_maxs[dir*2] = (bounds[dir][0] - tp.dir) / v[dir]       # tp.dir is not really defined with your current data structure
      t_maxs[dir*2+1] = (bounds[dir][-1] - tp.dir) / v[dir]
```

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+from __future__ import division
+import math
+
+
+##################################
+### Facet voxel traverse tools ###
+##################################
+
+
+class Point(object):
```
zxkjack123: Make Point iterable.

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+    y_idx : int
+        Modified y_idx.
+    z_idx : int
+        Modified z_idx.
+    tp : Point
+        Point after moving along the direction of v.
+    t_temp : float
+        Updated travel length. Distance from the start point to updated point.
+    """
+    t_min = min(t_maxs)
```
gonuke: look to use `numpy.argmin`

gonuke:
```
update_idx = [0,0,0]
update_dir = nump.argmin(t_maxs)
update_idx[update_dir] = 1
[x_idx,y_idx,z_idx] = [x_idx,y_idx,z_idx] + update_idx
```

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+    -------
+    t_maxs : list
+        The maximum travel lenght in current voxel for 3 directions. Format: 
+        [t_max_x, t_max_y, t_max_z]. These value could be both positive and
+        negtive. And could be 'inf' or '-inf'.
+    """
+    n_x_grid = _find_next_grid_1d(tp.x, v[0], bounds[0])
+    n_y_grid = _find_next_grid_1d(tp.y, v[1], bounds[1])
+    n_z_grid = _find_next_grid_1d(tp.z, v[2], bounds[2])
+    if v[0] == 0.0:
```
gonuke: see below for loop version of this

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+
+    Return:
+    -------
+    n_grid : float
+        The coordinate value of next mesh grid.
+    """
+    n_grid = 0.0
+    flag = False
+    if v_1d > 0:
+        for i in range(len(bound)):
```
gonuke:
```
for bnd_pos in bound:
   if cor < bnd_pos:
       n_grid = bnd_pos
       flag  = True
       break
```

pyne/vta.py
```diff
@@ -0,0 +1,566 @@
...
+        The start point of the line segment.
+    end : Point
+        The end point of the line segment.
+
+    Return:
+    -------
+    dist : float
+        The distance between the start point and the end point.
+    """
+    dist = math.sqrt(math.pow(end.x - start.x, 2) +
```
zxkjack123: Use (end.x - start.x) * (end.x - start.x)

pyne/vta.py
zxkjack123: Maybe just a numpy array of size 3.

pshriwise: One option for this is to make a list of the x,y,z values and index that list. This would allow you to keep the nicely named attributes if you're referring to a single value, but still append to lists or combine points cleanly. 

```
def __iter__(self, key):
    l = list(self.x, self.y, self.z)
    return l[key]
```

zxkjack123:
Major changes:
- Use numpy array to replace the class `Point`.
- Put some variables, such as `x_idx,  y_idx, z_idx` into a list or numpy array. Make them iterable to reduce the code duplication.
- Change the method for rays creating. Divide the shortest triangle edge uniformly.

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+                break
+    else:
+        for i in range(len(bound)-1, -1, -1):
+            if bound[i] < cor <= bound[i+1]:
+                idx = i
+                break
+    # idx could be -1, that means point outside the mesh
+    return idx
+
+def _cal_dir_v(start, end):
```
pshriwise: Could we afford to be a little more verbose here? Even `_cal_dir_v` --> `_calc_vec_dir` provides a more clarity.

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+
+def _find_next_grid_1d(cor, v_1d, bound):
+    """
+    Find the next grid coordinate of a direction.
+
+    Parameters:
+    -----------
+    cor : float
+        Coordinate of current point in the direction.
+    v_1d : float
```
pshriwise: Is there a particular reason this is a float rather than a Bool or Int?

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+    Move the point to next voxel.
+
+    Parameters:
+    -----------
+    idxs: list
+        List of indexes, format: [x_idx, y_idx, z_idx]
+    tp : numpy array
+        An array of size 3. Temporary point position.
+    t_temp: float
+        Temporary travel lenght from the start.
```
pshriwise: lenght --> length

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+    tp : numpy array
+        An array of size 3. Point after moving along the direction of v.
+    t_temp : float
+        Updated travel length. Distance from the start point to updated point.
+    """
+    t_min = min(i for i in t_maxs if i > 0)
+    update_dir = list(t_maxs).index(t_min)
+    update_idx = [0, 0, 0]
+    update_idx[update_dir] = 1
+    idxs = [x + y for x, y in zip(idxs, update_idx)]
```
pshriwise: Should this always be an addition?

pyne/vta.py
pshriwise: It might be nice to provide some extra clarity about this. If I understand correctly, the returned `t_temp` value is the total distance traveled from the start of the ray, but that isn't used or provided as part of this function, which is a bit confusing.

I'll suggest something like "Updated travel length since the beginning of the ray." I'm not sure if that's better really 😞

Also, having different names for outgoing variables might be useful to provide clarity here.

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+    return idxs, tp, t_temp
+
+def _move_to_boundary(tp, t_min, v):
+    """
+    Move the point to the nearest boundary when initializing the problem.
+
+    Parameters:
+    -----------
+    tp : numpy array
+        An array of size 3. Temporary point. Actually, it is also the start point.
```
pshriwise: Not a huge deal but `temporary point` in a lot of the documentation is messing with me. In the context of these individual functions, the "temporary point" is the only point it knows about! I might suggest that `temporary point` --> `point` in a lot of these cases.

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+    v = _cal_dir_v(start, end)
+    t_end = _distance(start, end)
+    t_temp = 0.0
+    while t_temp < t_end:
+        if _is_point_in_mesh(bounds, tp):
+            # the point is in the mesh
+            # calculate the voxel id
+            for dr in range(len(idxs)):
+                idxs[dr] = _find_voxel_idx_1d(bounds[dr], tp[dr], v[dr])
+            if -1 in idxs:
```
pshriwise, line 271: Why this extra check? We test if the point is in the mesh above right?

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+                # point outside the mesh
+                return voxels
+            # add current voxel
+            voxels.add((idxs[0], idxs[1], idxs[2]))
+            t_maxs = _cal_max_travel_length_in_current_voxel(tp, end, v, bounds)
+            idxs, tp, t_temp = _move_to_next_voxel(idxs, tp, t_temp, t_maxs, v)
+        else:
+            # Check whether the ray intersecs the mesh. Calculate the maxinum travel
```
pshriwise: maxinum --> maximum

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+            # length of the ray to the outside boundaries of the mesh. If all the
+            # length are negtive, then there is no intersection, return. Ohterwise,
```
pshriwise: length --> lengths

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+    t_min = min(i for i in t_maxs if i > 0)
+    update_dir = list(t_maxs).index(t_min)
+    update_idx = [0, 0, 0]
+    update_idx[update_dir] = 1
+    idxs = [x + y for x, y in zip(idxs, update_idx)] 
+    tp += t_min * v
+    t_temp += t_min
+    return idxs, tp, t_temp
+
+def _move_to_boundary(tp, t_min, v):
```
pshriwise: Is this function used anywhere? It seems like there is a nice opportunity to extend it's capability and replace some code in the `_ray_voxel_traverse` function (see below).

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+            voxels.add((idxs[0], idxs[1], idxs[2]))
+            t_maxs = _cal_max_travel_length_in_current_voxel(tp, end, v, bounds)
+            idxs, tp, t_temp = _move_to_next_voxel(idxs, tp, t_temp, t_maxs, v)
+        else:
+            # Check whether the ray intersecs the mesh. Calculate the maxinum travel
+            # length of the ray to the outside boundaries of the mesh. If all the
+            # length are negtive, then there is no intersection, return. Ohterwise,
+            # choose the minimum length and move the point to that position.
+            # left, right, back, front, down, up
+            t_maxs = [0.0] * 6
```
pshriwise: The code here is kind of what I'd expect to see in the `_move_to_boundary` function defined above.

Ideally I think there would be two calls here:

```
intersects = _check_mesh_intersection(start, end, bounds) # check if they ray intersects the mesh
if not intersects:
    break
pnt = _move_to_boundary(pnt, bounds)
```

or something like that.

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+        The minimum grid step.
+    """
+    min_step = float('inf')
+    for bound_list in bounds:
+        for bound_idx in range(len(bound_list)-1):
+            min_step = min(min_step,
+                           bound_list[bound_idx+1] - bound_list[bound_idx])
+    return min_step
+
+def _divide_edge(start, end, step):
```
pshriwise: I'll recommend this gets renamed to `_divide_tri_edge` for clarity

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+    """
+    Create rays from points. In which, 'start' is the start point of all the
+    rays, and the point in the `insert_points' are the end points.
+
+    Parameters:
+    -----------
+    start : numpy array
+        An array of size 3, represents the start point.
+    insert_points : list of arrays
+        List of points.
```
pshriwise: "List of ray end points" perhaps?

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+        List of rays (line segments).
+    """
+    rays = []
+    for p in insert_points:
+        rays.append([start, p])
+    return rays
+
+def _create_rays_from_triangle_facet(A, B, C, min_step):
+    """
+    Create rays form triangle for scanning.
```
pshriwise: form --> from

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+    -----------
+    A : numpy array
+        An array of size 3, represents a vertex of the triangle
+    B : numpy array
+        An array of size 3, represents a vertex of the triangle
+    C : numpy array
+        An array of size 3, represents a vertex of the triangle
+    min_step: float
+        The minimum step of the mesh grid. Used to divide the shortest edge
+        of the facet triangle. Use this value could make sure that every voxel
```
pshriwise: This needs some rewording...

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+        source = (A, B)
+
+    insert_points = _divide_edge(source[0], source[1], min_step)
+    rays = _create_rays_from_points(target, insert_points)
+    return rays
+
+
+def _facet_voxel_traverse(A, B, C, bounds):
+    """
+    Facet voxel traverse. Calculate all the voxles that intersect with the
```
pshriwise: Can remove the function name in the docstring here.

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+    facet. Facet is represented with 3 triangle verteice.
```
pshriwise: vertices

pyne/vta.py
pshriwise: This may happen in other places too. Please check it out...

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+    rays = _create_rays_from_triangle_facet(A, B, C, min_step)
+    voxels = set()
+    for ray in rays:
+        temp_voxels = _ray_voxel_traverse(bounds, ray[0], ray[1])
+        # merge two lists
+        voxels = voxels.union(temp_voxels)
+    return voxels
+
+if __name__ == '__main__':
+    my_list = [[1, 1], [2, 2]]
```
pshriwise: I'm not understanding this behavior when `vta.py` is called as main...

tests/test_vta.py
```diff
@@ -0,0 +1,73 @@
...
+from pyne import vta
+
+
+def test_ray_voxel_traverse_2D():
+    x_bound = [0, 1, 2, 3, 4, 5, 6]
+    y_bound = [0, 1, 2, 3, 4, 5, 6]
+    z_bound = [0, 1]
+    bounds = [x_bound, y_bound, z_bound]
+    A = np.array([3, 5.5, 0.5])
```
pshriwise, line 57: I'd recommend changing `A`/`B` to `ray_start`/`ray_end`. In other places A, B, C are used to represent triangle vertices.

tests/test_vta.py
```diff
@@ -0,0 +1,73 @@
...
+    Ray traverse history:
+    Point: [0.3, 0.4, 0.5] -> [0.88, 0.94, 1] -> [0.94, 1, 1.06] -> [1, 1.06, 1.11]
+    Voxel: (0, 0, 0)       -> (0, 0, 1)       -> (0, 1, 1)       -> (1, 1, 1)
+    """
+    x_bound = [0, 1, 2]
+    y_bound = [0, 1, 2]
+    z_bound = [0, 1, 2]
+    bounds = [x_bound, y_bound, z_bound]
+    A = np.array([0.3, 0.4, 0.5])
+    B = np.array([2, 2, 2])
```
pshriwise, line 152: see comment before this one

tests/test_vta.py
```diff
@@ -0,0 +1,73 @@
...
+    x_bound = [0, 1, 2, 3, 4, 5, 6]
+    y_bound = [0, 1, 2, 3, 4, 5, 6]
+    z_bound = [0, 1]
+    bounds = [x_bound, y_bound, z_bound]
+    A = np.array([3, 5.5, 0.5])
+    B = np.array([0.5, 0.5, 0.5])
+    results = vta._ray_voxel_traverse(bounds, A, B)
+    exp_ans = set([(2, 5, 0), (2, 4, 0), (2, 3, 0), (1, 3, 0),
+               (1, 2, 0), (1, 1, 0), (0, 1, 0), (0, 0, 0)])
+#    for i in range(len(exp_ans)):
```
pshriwise: Why are these lines commented?

tests/test_vta.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    x_bound = [0, 1, 2]
+    y_bound = [0, 1, 2]
+    z_bound = [0, 1, 2]
+    bounds = [x_bound, y_bound, z_bound]
+    A = np.array([0.3, 0.4, 0.5])
+    B = np.array([2, 2, 2])
+    results = vta._ray_voxel_traverse(bounds, A, B)
+    exp_ans = set([(0, 0, 0), (0, 0, 1), (0, 1, 1), (1, 1, 1)])
+#    for i in range(len(exp_ans)):
```
pshriwise: Same as above, why commented? Can they be removed?

pshriwise (changes requested):
@zxkjack123 I have a lot of comments, but this was a real pleasure to review. You did a great job drawing boundaries between your functions and it made the higher-level functions easy to understand. I think its really close and it will be a solid addition to PyNE when it's done!

pyne/vta.py
zxkjack123: The `v_1d` here is a number in `[v_x, v_y, v_z]`, it is a float. It can be converted to a `bool` or `int` before the next `if` function. However, the conversion will not simplify the code. For the better understanding of the code, I keep this parameter to be a float.

pyne/vta.py
zxkjack123: Will be revised in next commit.

pyne/vta.py
zxkjack123: Found a bug here, the `update_idx[update_dir]` could be both `1` or `-1`. This is fixed in next commit.

pyne/vta.py
zxkjack123: The comment here should be revised. 
As the mesh has outside boundaries. A point may be exactly on the boundary. The function `_is_point_in_mesh` will give a `True` value if the point is on the boundary. 
However, we don't need to traverse a point that has a direction toward outside. This is the reason of the extra check here.

pyne/vta.py
pshriwise: I think that's ok, yeah. The odd part to me was a float being used to make a binary choice in the code, but I see what you mean. It also comes in handy at the end of the function for setting the sign of INF. 

I'm going to propose something else to simplify the code here though. The for loops can be consolidated like so: 

```
if v_1d < 0: 
    bound = reversed(bound)
    cor = -cor

for bnd in bound:
    if cor < bnd_pos:
        n_grid = bnd_pos
        flag = True
        break
```

The for loop code can always be the same if the iterable is setup correctly.

pyne/vta.py
```diff
@@ -0,0 +1,481 @@
...
+    """
+    Find the next grid coordinate of a direction.
+
+    Parameters:
+    -----------
+    cor : float
+        Coordinate of current point in the direction.
+    v_1d : float
+        Direction value.
+    bound : list of float
```
pshriwise: might be best to go with bounds rather than bound because the variable is a list of values

pyne/vta.py
pshriwise: Then will triangle on the boundary of the mesh not be counted as intersecting with the voxel? Should they be?

I'm feeling like this is a corner case that will be exposed later on...

pyne/vta.py
zxkjack123: No, this function is not used anywhere because it's too simple, only one line. I will remove this function.

pyne/vta.py
zxkjack123: Actually, these two calls are both quite simple. If they are made into functions, there would be only one or two lines.

pshriwise: To me, it's less about the number of lines and more about separating the functionality, which you've done really well in other places in this module.

pshriwise: To put it another way, there are many functions in this module which are only a few lines, but they serve a distinct purpose and make the code easy to follow when used in other places. I think that practice could be repeated here.

zxkjack123:
Major changes:
- Fixed a bug about the voxel index increase step (could be -1), and added a test case for it
- Renamed some names of functions and variables, made them more verbose
-

pyne/vta.py
zxkjack123: I used the `_move_to_boundary` here in the new commit. 
As for the intersection checking function that checks whether the line segment intersects with the mesh is not finished by a single function. This check is finished during the `while` loop. In each single the loop step, the if clause below can't tell whether there will be intersection if it's ture, but there will definitely no intersection if it's false. 
```
if any((x > 0).all() for x in dist_maxs):
```
So, this is not an independent checking function.

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+    Find the voxel id (in a given dimension).
+
+    Parameters:
+    -----------
+    bounds_1d : list
+        Boundary of the mesh in a given direction.
+    cor : float
+        Coordinate of the point in the given direction.
+    vec_1d : float
+        Vector attribute in the given direction.
```
gonuke: It is not clear how this parameter is used.  The docstring might want to describe what the purpose/effect of this parameter is.

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+    Return:
+    -------
+    idx : int
+        Mesh index of the given direction. This value is -1 when the point not
+        in the mesh.
+    ----------
+    """
+    idx = -1
+    if vec_1d > 0:
+        for i in range(len(bounds_1d)-1):
```
gonuke: linear search is inefficient.  A binary search should be straightforward.  Also - does numpy have a `find()` method, or similar that might be effective?

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+    end :
+        An array of size 3. The end point.
+
+    Return:
+    -------
+    vec : numpy array
+        Direction from start to end, an unit vector.
+    """
+    vec = end - start
+    dist = math.sqrt(vec[0] * vec[0] + vec[1] * vec[1] + vec[2] * vec[2])
```
gonuke: once you've decided to import `numpy` might be better to use `np.sqrt()`. In this case, it's probably equivalent, but it may remove the `math` dependency

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+    vec = np.divide(vec, dist)
+    return vec
+
+def _find_next_grid_1d(cor, vec_1d, bounds_1d):
+    """
```
gonuke, line 108: can't this reuse `_find_voxel_idx_1d`?

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+
+    Return:
+    -------
+    dist : float
+        The distance between the start point and the end point.
+    """
+    dist = math.sqrt((end[0] - start[0]) * (end[0] - start[0]) +
+                     (end[1] - start[1]) * (end[1] - start[1]) +
+                     (end[2] - start[2]) * (end[2] - start[2]))
+    return dist
```
gonuke: `dist = np.linalg.norm(end-start)`

pyne/vta.py
gonuke: `numpy.searchsorted()`

pyne/vta.py
gonuke: `return np.divide(end-start,np.linalg.norm(end-start))`

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+    -------
+    vec : numpy array
+        Direction from start to end, an unit vector.
+    """
+    vec = end - start
+    dist = math.sqrt(vec[0] * vec[0] + vec[1] * vec[1] + vec[2] * vec[2])
+    vec = np.divide(vec, dist)
+    return vec
+
+def _find_next_grid_1d(cor, vec_1d, bounds_1d):
```
zxkjack123, line 107: Similar to find_index_1D.

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+    Return:
+    -------
+    dist_maxs : numpy array
+        The maximum travel lenght in current voxel for 3 directions. Format: 
+        [dist_max_x, dist_max_y, dist_max_z]. These value could be both
+        positive and negtive. And could be 'inf' or '-inf'.
+    """
+    n_x_grid = _find_next_grid_1d(point[0], vec[0], bounds[0])
+    n_y_grid = _find_next_grid_1d(point[1], vec[1], bounds[1])
+    n_z_grid = _find_next_grid_1d(point[2], vec[2], bounds[2])
```
gonuke, line 163:
```
next_grid = np.array[n_x_grid, n_y_grid, n_z_grid]
return np.divide(next_grid - point, vec)
```

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+    Returns:
+    --------
+    idxs: list
+        Updated indexes.
+    updated_point : numpy array
+        An array of size 3. Updated point positon.
+    dist_temp : float
+        Updated travel length since the beginning of the ray."""
+
+    dist_min = min(i for i in dist_maxs if i > 0)
```
gonuke:
```suggestion
    dist_min = min(dist_maxs[dist_maxs>0])
```

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+                dist_maxs[dr*2] = np.divide(bounds[dr][0] - point[dr],
+                                            vec[dr])
+                dist_maxs[dr*2+1] = np.divide(bounds[dr][-1] - point[dr],
+                                              vec[dr])
+            if any((x > 0).all() for x in dist_maxs):
+                # current point outside the mesh
+                # there maybe intersects to the mesh
+                dist_min = min(i for i in dist_maxs if i > 0)
+                # Move the point to the boundary, calculate the new coordinate
+                point = _move_to_boundary(point, dist_min, vec)
```
gonuke: `point = point + min_dist_bdry * dir_vec`

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+    """
+    if any((x == 1.0).all() for x in vec):
+        # this is a axis aligned ray
+        for i in range(len(vec)):
+            if vec[i] == 1.0:
+                continue
+            else:
+               if any((x == start[i]).all() for x in bounds[i]):
+                   return True
+        return False
```
gonuke:
```
for idx in np.argwhere(a!=1.0):
   if start[i[0]] in bounds[i[0]]:
       return True
```

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+        Boundaries of mesh grid.
+
+    Return:
+    -------
+    min_step: float
+        The minimum grid step.
+    """
+    min_step = float('inf')
+    for bound_list in bounds:
+        for bound_idx in range(len(bound_list)-1):
```
gonuke: `min_step = min(min_step, min(bound_list[2:] - bound_list[:-1]))`

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+    vec = _calc_vec_dir(start, end)
+    dist_temp = 0.0
+    insert_points = []
+    # +2 here to make sure it's a conservative result
+    num_segment = math.floor(dist/step) + 2
+    x_list = np.linspace(start[0], end[0], num_segment)
+    y_list = np.linspace(start[1], end[1], num_segment)
+    z_list = np.linspace(start[2], end[2], num_segment)
+    for (x, y, z) in zip(x_list, y_list, z_list):
+        insert_points.append([x, y, z])
```
gonuke:
```suggestion
        insert_points.append(np.array([x, y, z]))
```

pyne/vta.py
```diff
@@ -0,0 +1,518 @@
...
+        source = (B, C)
+    elif b == min_edge:
+        target = B
+        source = (A, C)
+    else:
+        target = C
+        source = (A, B)
+
+    insert_points = _divide_tri_edge(source[0], source[1], min_step)
+    rays = _create_rays_from_points(target, insert_points)
```
gonuke:
```suggestion
    rays = [ (target, point) for point in insert_points ]
```

gonuke (changes requested):
Here are some suggested changes based on better use of numpy

zxkjack123:
Major changes:
- Simplified the code using some functions in numpy

pshriwise:
Just a heads up for @zxkjack123 that I haven't forgotten about this. Just a very busy week. I will do my best to look over the updated changes within the next few days.

In the meantime can someone with write access to the repo (@gonuke?)  restart the CI builds. It looks like they failed due to a timeout of an ace dataset download.

pshriwise:
Oops I spoke too soon... @zxkjack123 it looks like the Python 3 build is failing due to a divide by zero error in the `vta.py` tests...

pyne/vta.py
```diff
@@ -0,0 +1,421 @@
...
+
+    Return:
+    -------
+    True : bool
+        Ray is on a boundary surface
+    False : bool
+        Ray is not on a boundary surface
+    """
+    if 1.0 in vec:
+        for idx in np.argwhere(vec!=1.0):
```
zxkjack123:
```suggestion
        for idx in np.argwhere(vec!=1.0).flatten():
```

zxkjack123:
Major changes:
- Add a function for triangle-box intersection check based on separate axis theorem
- Add a test function for this triangle-box intersection check

pyne/vta.py
```diff
@@ -0,0 +1,514 @@
...
+    """
+
+    # variables names used in the code
+    # e, unit vector of axis. e0=(1, 0, 0), e1=(0, 1, 0), e2=(0, 0, 1).
+    # v, vertices of the triangles. v0, v1, and v2.
+    # f, edge vector of the triangle.  f0=v1-v0, f1=v2-v1, f2=v0-v2
+    # a, axis for test in bullet 3. aij = ei x fj, i.e., a00 = e0 x f0
+
+    # define bais axis
+    e = np.zeros(shape=(3, 3))
```
gonuke: `e = np.eye(3,3)`

pyne/vta.py
```diff
@@ -0,0 +1,514 @@
...
+    e[0] = np.array([1, 0, 0])
+    e[1] = np.array([0, 1, 0])
+    e[2] = np.array([0, 0, 1])
+    # Translate triangle as conceptually moving AABB to origin
+    v = np.zeros(shape=(3, 3))
+    v[0] = triangle[0] - box_center
```
gonuke: ??? `v = triangle - box_center`  ??

pyne/vta.py
```diff
@@ -0,0 +1,514 @@
...
+    f = np.zeros(shape=(3, 3))
+    f[0] = triangle[1] - triangle[0]
+    f[1] = triangle[2] - triangle[1]
+    f[2] = triangle[0] - triangle[2]
+
+    ## region Test axes a00..a22 (category 3)
+    for i in range(0, 3):
+        for j in range(0, 3):
+            a = np.cross(e[i], f[j])
+            p0 = np.dot(v[0], a)
```
gonuke: ?? `p = np.mult(v,a)`

pyne/vta.py
```diff
@@ -0,0 +1,514 @@
...
+            p2 = np.dot(v[2], a)
+            r = np.dot(box_extents, np.absolute(a))
+            if min(p0, p1, p2) > r or max(p0, p1, p2) < -r:
+                return False
+    ## endregion
+
+    ## region Test the three axes corresponding to the face normals of AABB b (category 1)
+    # Exit if...
+    # x direction
+    if max(v[:, 0]) < -box_extents[0] or min(v[:, 0]) > box_extents[0]:
```
gonuke: loop over directions (0,1,2)

zxkjack123:
Update:
- Simplify the function tri-box intersection check via vectorizing some variables

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+    True : when point in the mesh.
+    False : when point outside the mesh.
+    """
+    if bounds[0][0] <= point[0] <= bounds[0][-1] and \
+       bounds[1][0] <= point[1] <= bounds[1][-1] and \
+       bounds[2][0] <= point[2] <= bounds[2][-1]:
+        return True
+    else:
+        return False
+
```
moatazharb, line 52: Why not check coordinates for being outside the boundaries in their respective directions?. All "AND" statements will be executed unless a "False" is found which means that all statements will be checked if they result in True. I suggest switching to "NOT OR" statements. This way if at least one coordinate is outside the mesh then there will be no need to continue checking the rest of the directions. 
Note: I might not have succeeded in explaining what I mean but you can check "short circuit" of logical operators.
```suggestion
    if not (bounds[0][0] <= point[0] <= bounds[0][-1]) or \
      not (bounds[1][0] <= point[1] <= bounds[1][-1]) or \
      not (bounds[2][0] <= point[2] <= bounds[2][-1]):
        return False
    else:
        return True
```

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+    Return:
+    -------
+    idx : int
+        Mesh index of the given direction. This value is -1 when the point not
+        in the mesh.
+    ----------
+    """
+    idx = -1
+    bounds_1d = np.array(bounds_1d)
```
moatazharb: What is the format of bounds_1d?. Is it from the nested list you mentioned in "_is_point_in_mesh" function?. If so then you may consider making the nested list into an array with each row containing the boundaries along one dimension;
bounds = np.array([[x_min, ..., x_max],
                                [y_min, ..., y_max],
                                [z_min, ..., z_max]]) 
This way bounds_1d = bounds[0] (or [1] or [2] according yo which direction is considered) and it is a numpy array so you wouldn't need to cast bounds_1d as an array here.

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+        idx = np.searchsorted(bounds_1d, cor, side='right')
+    else:
+        idx = np.searchsorted(bounds_1d, cor, side='left')
+    if idx == 0 or idx == len(bounds_1d):
+    # idx could be -1, that means point outside the mesh
+        return -1
+    else:
+        return idx - 1
+
+def _calc_vec_dir(start, end):
```
moatazharb, line 88: One line!, probably no need for a function since it's too simple. Even if it's used many times in many functions, it is still one line and no need to be included in a separate function. Up to you :)

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+        Boundary values of current direction.
+
+    Return:
+    -------
+    n_grid : float
+        The coordinate value of next mesh grid.
+    """
+    voxel_idx = _find_voxel_idx_1d(bounds_1d, cor, vec_1d)
+    if voxel_idx == -1:
+        return vec_1d * float('inf')
```
moatazharb: Why not np.inf?. [https://stackoverflow.com/questions/42315541/difference-between-np-inf-and-floatinf](https://stackoverflow.com/questions/42315541/difference-between-np-inf-and-floatinf)

pyne/vta.py
zxkjack123: Though it's only one line, but I think the function name `calc_vec_dir` are easier to understand than the code. So, I will keep this function.

pyne/vta.py
zxkjack123: Applied this change. Numpy array is better than the nested list in this case.

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+    dist_maxs : numpy array
+        The maximum travel lenght in current voxel for 3 directions. Format: 
+        [dist_max_x, dist_max_y, dist_max_z]. These value could be both
+        positive and negtive. And could be 'inf' or '-inf'.
+    """
+    n_x_grid = _find_next_grid_1d(point[0], vec[0], bounds[0])
+    n_y_grid = _find_next_grid_1d(point[1], vec[1], bounds[1])
+    n_z_grid = _find_next_grid_1d(point[2], vec[2], bounds[2])
+    next_grid = np.array([n_x_grid, n_y_grid, n_z_grid])
+    return np.divide(next_grid - point, vec)
```
moatazharb: Why divide by the unit vector "vec"?. If I understand correctly, you calculated the distance from the point to the next grid boundaries and you need to get the components that correspond to a vector from the point to the boundaries along "vec", so I guess you need the projection of "next_grid - point" on "vec!

Also create a variable and add a comment above to describe what you're calculating then return it
```suggestion
    # Calculate .......
    dist_maxs = np.divide(next_grid - point, vec)
    return dist_maxs
```

moatazharb: Can you add a comment above the np.divide calculation to describe it and also create a variable and return it?.
```suggestion
    # Divide by direction cosines to calculate the distance along vec
    dist_maxs = np.divide(next_grid - point, vec)
    return dist_maxs
```

moatazharb: I think also that the function should return the minimum of all three distances as this would be the maximum travel distance in the voxel along vec before reaching a boundary.

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+
+    dist_min = min(dist_maxs[dist_maxs>0])
+    update_dir = list(dist_maxs).index(dist_min)
+    update_idx = [0, 0, 0]
+    # vec[update_dir] will not be 0
+    if vec[update_dir] > 0:
+        update_idx[update_dir] = 1
+    elif vec[update_dir] < 0:
+        update_idx[update_dir] = -1
+    idxs = [x + y for x, y in zip(idxs, update_idx)]
```
moatazharb: Why not just:
```suggestion
    # vec[update_dir] will not be 0
    if vec[update_dir] > 0:
        idxs[update_dir] += 1
    elif vec[update_dir] < 0:
        idxs[update_dir] -= 1 
```

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+        End point.
+
+    Return:
+    -------
+    voxles : list of tuples
+        List of voxel indices. These voxels intersect with the line segment.
+    """
+    # Initialization
+    # Find the voxel that start point in
+    point = np.copy(start)
```
moatazharb, line 256: maybe
```suggestion
    point = start[:]
```

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+    idxs = [-1, -1, -1]
+    voxels = set()
+    vec = _calc_vec_dir(start, end)
+    # if the ray on a boundary, then return empty set
+    if _is_ray_on_boundary(start, vec, bounds):
+        return set()
```
moatazharb: for consistency, why not just return "voxels" since this is the first check after creating the set it'll be empty as you needed it to be if the if statement is True.
```suggestion
        return voxels
```

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+    dist_end = np.linalg.norm(end-start)
+    dist_temp = 0.0
+    while dist_temp < dist_end:
```
moatazharb, line 265: There is a possible problem here; dist_temp not being exactly = dist_end in which case the while loop will not terminate. 
A better way to do this is to (add) check the the index of the voxel of the current point against the end point.
Another way is to use a comparison within some rounding limit like numpy.isclose.

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+            idxs, point, dist_temp = _move_to_next_voxel(
+                idxs, point, dist_temp, dist_maxs, vec)
+        else:
+            # Check whether the ray intersecs the mesh. Calculate the maximum travel
+            # length of the ray to the outside boundaries of the mesh. If all the
+            # lengths are negtive, then there is no intersection, return. Ohterwise,
+            # choose the minimum length and move the point to that position.
+            # left, right, back, front, down, up
+            dist_maxs = [0.0] * 6
+            for dr in [0, 1, 2]:
```
moatazharb: For consistency, why not
```suggestion
            for dr in range(len(idxs):
```

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+                dist_maxs[dr*2] = np.divide(bounds[dr][0] - point[dr],
+                                            vec[dr])
+                dist_maxs[dr*2+1] = np.divide(bounds[dr][-1] - point[dr],
+                                              vec[dr])
+            if any((x > 0).all() for x in dist_maxs):
+                # current point outside the mesh
+                # there maybe intersects to the mesh
+                dist_min = min(i for i in dist_maxs if i > 0)
```
moatazharb:
```suggestion
                dist_min = min(dist_maxs[dist_maxs > 0])
```

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+        The triangle does not intersect with the box.
+    """
+
+    # variables names used in the code
+    # e, unit vector of axis. e0=(1, 0, 0), e1=(0, 1, 0), e2=(0, 0, 1).
+    # v, vertices of the triangles. v0, v1, and v2.
+    # f, edge vector of the triangle.  f0=v1-v0, f1=v2-v1, f2=v0-v2
+    # a, axis for test in bullet 3. aij = ei x fj, i.e., a00 = e0 x f0
+
+    # define bais axis
```
moatazharb: typo, "basis"

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+    e = np.eye(3, 3)
```
moatazharb, line 478: Maybe call it unit_vec or axes_unit_vecs. Similarly for "v", "f", and "a". It makes it easier to follow when the name of the variable says something about it.

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+    Return:
+    -------
+    True : bool
+        The triangle intersects with the box.
+    False : bool
+        The triangle does not intersect with the box.
+    """
+
+    # variables names used in the code
+    # e, unit vector of axis. e0=(1, 0, 0), e1=(0, 1, 0), e2=(0, 0, 1).
```
moatazharb: wouldn't it be better to call it ex, ey, ez since it's aligned witht the axes?

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+    # Translate triangle as conceptually moving AABB to origin
+    v = triangle - box_center
+    # Compute edge vectors for triangle
+    f = np.zeros(shape=(3, 3))
+    f[0] = triangle[1] - triangle[0]
+    f[1] = triangle[2] - triangle[1]
+    f[2] = triangle[0] - triangle[2]
+
+    ## region Test axes a00..a22 (category 3)
+    for i in range(0, 3):
```
moatazharb: just range(3) should be fine

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+        for j in range(0, 3):
+            a = np.cross(e[i], f[j])
```
moatazharb, line 492: Add description before each of these products. like: # Calculating the normal to edge-axis plane, etc.

pyne/vta.py
```diff
@@ -0,0 +1,500 @@
...
+            p = np.dot(v, a)
+            r = np.dot(box_extents, np.absolute(a))
+            if min(p) > r or max(p) < -r:
```
moatazharb, line 498: Add description of what this condition means!.

pyne/vta.py
zxkjack123: The `dist_temp` here is the distance  from the start point of the ray to the intersection point of the ray with the mesh boundary. The `dist_temp` could be `inf` if the ray goes out of the mesh. So, the `dist_temp` will always be greater than the `dist_end` (a definite value) in several iterations. Therefore, the loop will terminate.

pyne/vta.py
zxkjack123: Well, the codes that use the separate axis theorem (SAT) are converted from the other people's codes which are based on several papers related to SAT. In the se papers/codes, the `e`, `a`, `f` and `v` are used.
In my opinion, keep these variables similar to the papers introducing SAT and the codes that apply SAT to the `box-triangle` check would be a better choice.

pyne/vta.py
zxkjack123: I changed them to `e[0], e[1], e[2]`.

zxkjack123:
Close this PR. Because the margin for speed up particle birth is too small. I tested with a R2S photon transport using FNG model, the CPU time of particle birth takes about only 1.5% when #1118 is applied. It's worthless to do source sampling with bounding box.

PR #1053, opened 2018-06-15
read_wwinp() check photon bins != 0

kkiesling:
It is possible that the first lines of a wwinp file specify a number of photon bins to be 0 (which means it does not exist in the file). ADVANTG writes out wwinp files like this. Currently, when using `read_wwinp()` if there is a number there, it assumes the values exist in the file. This pr just adds a check to make sure the number of photon energy bins is not 0 even if a number exists there (just like the neutron bin check in the lines above).


scopatz:
Thanks @kkiesling!

PR #1058, opened 2018-07-12
Fixing an error with _find_phsrc_dc function in alara.py

moatazharb:
This PR fixes an error with _find_phsrc_dc function in alara.py that arises sometimes when writing photon sources to hdf5 files in r2s step2.
For each decay time read from the config file, r2s.py step2 calls photon_source_hdf5_to_mesh function that creates a list of all decay times read from phtn_src.h5 file and then loops over that list to find the matching decay time to write the source file for that time.
The issue is demonstrated by the following example:
- Input decay times from r2s config file: 
decay_times: 0 s,1e2 s,1e4 s,1e6 s,1e8 s,1e10 s
- List of decay times retrieved from phtn_src.h5 in photon_source_hdf5_to_mesh function:
['0 s', '1e+10 s', '1e+08 s', 'shutdown', '100 s', '1e+06 s', '10000 s']

It can be seen that an exact match using string comparison cannot be found for most of the decay times. The two sources of errors arise when an exact match cannot be found; 
1- looping over the list to convert units to seconds and compare where an error is raised for "shutdown"
2- trying to find the nearest value where dividing by the first value in the list, '0', causes an error.

This PR fixes these issues.

### DIFF
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -869,25 +869,39 @@ def _convert_unit_to_s(dc):
 
 def _find_phsrc_dc(idc, phtn_src_dc):
     """
-    This function return a string representing a time in phsrc_dc.
+    This function returns a string representing a time in phsrc_dc.
+
     Parameters
     ----------
-    idc : string. Represent a time, input decay time
-    phtn_src_dc : list of string. Decay times in phtn_src file.
+    idc : string
+        Represents a time, input decay time
+    phtn_src_dc : list of strings
+        Decay times in phtn_src file
 
-    return : odc, string. output mathched decay time.
-    """ 
+    Returns
+    -------
+    string from phtn_src_dc list that mathches idc
+    """
+    # Check the existence of idc in phtn_src_dc list.
     if idc in phtn_src_dc:
-        # if idc exist in phtn_src_dc, directly return idc
         return idc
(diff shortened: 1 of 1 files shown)


zxkjack123:
@moatazharb Thanks, I forgot the zero cooling time. But there is another problem, does the 'zero cooling time' equivalent to 'shutdown'? If 'zero cooling time' is the 'shutdown', maybe we should define '0 s' as an error input and raise an error information at the step1.

moatazharb:
@zxkjack123 Indeed '0 s' = 'shutdown'. It's just how it is treated in r2s now that the result for 'shutdown' is not included when writing the photon sources unless you specify that with a '0 s' entry in decay times on the config file. It's a matter of whether the user needs the source at shutdown or not?. are we gonna write the source by default or upon request as specified in decay times input in config file?. If this is to change and we treat a 0 s entry as an error then we need to carefully edit the r2s as this will not appear only in _find_phsrc_dc function but also in other functions where the h5m files are written.
Since anyway you'll need the user to specify that he wants the source at shutdown then '0 s' seems a good option and will enable numerical treatment of all decay times.

zxkjack123:
@moatazharb LGTM.

scopatz:
Thanks @moatazharb @zxkjack123!

PR #1065, opened 2018-10-08
Use pyne data

pshriwise:
This PR fixes CI by relying on files served from the `pyne/data` repository. 

One thing **NOT** included here is a fix for builds of PyNE relying on the origen benchmark data for hdf5 v1.8.13. I couldn't find those in either of the PyNE docker images.


zxkjack123:
This PR passed the CI checks. Great!

PR #1069, opened 2018-11-06
Fix subvoxel r2s total photon intensities

zxkjack123:
Major changes:
- Add a function to write the total photon intensities for subvoxel r2s
- Add a test function for the new function

### DIFF
--- a/pyne/r2s.py
+++ b/pyne/r2s.py
@@ -161,7 +161,7 @@ def photon_sampling_setup(mesh, phtn_src, tags):
     photon_source_hdf5_to_mesh(mesh, h5_file, tags)
 
 
-def total_photon_source_intensity(m, tag_name):
+def total_photon_source_intensity(m, tag_name, sub_voxel=False):
     """This function reads mesh tagged with photon source densities and returns
     the total photon emission desinty.
 
@@ -171,7 +171,9 @@ def total_photon_source_intensity(m, tag_name):
        The mesh-based photon emission density distribution in p/cm3/s.
     tag_name : str
        The name of the tag on the mesh with the photon emission density information.
-
+    sub_voxel: bool, optional
+        If true, sub-voxel r2s work flow will be used.
+                         
     Returns
     -------
     intensity : float
@@ -180,8 +182,18 @@ def total_photon_source_intensity(m, tag_name):
 
     sd_tag = m.mesh.getTagHandle(tag_name)
     intensity = 0.
-    for idx, _, ve in m:
-        vol = m.elem_volume(ve)
-        ve_data = sd_tag[ve]
-        intensity += vol*np.sum(ve_data)
+    if sub_voxel:
+        max_num_cells = len(m.cell_fracs[:][0])
+        for idx, _, ve in m:
+            ve_data = sd_tag[ve]
+            num_e_groups = len(ve_data) / max_num_cells
+            for svid in range(max_num_cells):
+                vol = m.elem_volume(ve) * m.cell_fracs[idx][svid]
+                sv_data = ve_data[num_e_groups*svid:num_e_groups*(svid+1)]
+                intensity += vol * np.sum(sv_data)
+    else:
+        for idx, _, ve in m:
+            vol = m.elem_volume(ve)
+            ve_data = sd_tag[ve]
+            intensity += vol*np.sum(ve_data)
     return intensity
--- a/pyne/r2s.py
+++ b/pyne/r2s.py
@@ -183,12 +183,13 @@ def total_photon_source_intensity(m, tag_name, sub_voxel=False):
     sd_tag = m.mesh.getTagHandle(tag_name)
     intensity = 0.
     if sub_voxel:
-        max_num_cells = len(m.cell_fracs[:][0])
+        cell_fracs = m.cell_fracs[:]
+        max_num_cells = len(cell_fracs[0])
         num_e_groups = len(sd_tag[list(m.iter_ve())[0]]) / max_num_cells
         for idx, _, ve in m:
             ve_data = sd_tag[ve]
             for svid in range(max_num_cells):
-                vol = m.elem_volume(ve) * m.cell_fracs[idx][svid]
+                vol = m.elem_volume(ve) * cell_fracs[idx][svid]
                 sv_data = ve_data[num_e_groups*svid:num_e_groups*(svid+1)]
                 intensity += vol * np.sum(sv_data)
     else:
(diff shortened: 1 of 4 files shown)


pshriwise:
The PR is for a fix. @zxkjack123 can you tell us a little bit about what the bug was for context?

tests/test_r2s.py
```diff
@@ -308,6 +308,26 @@ def test_total_photon_source_intensity():
     intensity = total_photon_source_intensity(m, "source_density")
     assert_equal(intensity, 58)
 
+def test_total_photon_source_intensity_subvoxel():
+    m = Mesh(structured = True, structured_coords=[[0, 1, 2],[0, 1, 3], [0, 1]])
```
pshriwise:
```suggestion
    m = Mesh(structured=True, structured_coords=[[0, 1, 2],[0, 1, 3], [0, 1]])
```

tests/test_r2s.py
```diff
@@ -308,6 +308,26 @@ def test_total_photon_source_intensity():
     intensity = total_photon_source_intensity(m, "source_density")
     assert_equal(intensity, 58)
 
+def test_total_photon_source_intensity_subvoxel():
```
pshriwise, line 311: I think I understand this, but a comment block explaining the setup for this test would be very helpful if anyone needs to modify it in the future.

pshriwise:
After looking at the code a little more I'll add my take on the fix here for posterity: 

It looks like this is fixing how photon source intensities are written to the source mesh for subvoxel mode by weighting each cell-based source contribution by the cell volume fraction to avoid over reporting the total source intensity. Is that right @zxkjack123?

tests/test_r2s.py
```diff
@@ -308,6 +308,26 @@ def test_total_photon_source_intensity():
...
+                     (2, 13, 0.5, 0.0), (2, 11, 0.5, 0.0),
+                     (3, 12, 0.5, 0.0), (3, 13, 0.5, 0.0)]
+    m.tag_cell_fracs(cell_fracs)
+    m.source_density = IMeshTag(4, float)
+    m.source_density[:] = [[0.0, 0.0, 1.0, 1.0],
+                           [2.0, 2.0, 3.0, 3.0],
+                           [4.0, 4.0, 5.0, 5.0],
+                           [6.0, 6.0, 7.0, 7.0]]
+    intensity = total_photon_source_intensity(m, "source_density", True)
+    assert_equal(intensity, 46.0)
```
pshriwise: Can we set a variable for this? (`expected_intensity`)

zxkjack123:
Thanks for your reminder @pshriwise .  The bug is that I forgot to change the total photon intensities calculation method. Therefore, in the subvoxel mode, it still uses the method for default r2s, leading to a wrong photon source intensities.
This error is not significant in small scale test cases that have small quantities of mesh voxels (less than 4% in my test problem named `demo_collimator`, mesh size: 16*5*1), but the error will increase in large scale problem (for example, in the calculation of FNG with mesh size: 30*30*45, the error is 160%).

zxkjack123:
@pshriwise That's true. In subvoxel mode, the volume fraction should be multiplied when calculate the total photon source intensities.

tests/test_r2s.py
pshriwise: Also, setting this variable using the math that is being done in the `total_photon_source_intensity` would provide a lot of clarity here. 
```
# expected intensity - each represents a mesh voxel, each value an energy group
expected_intensity = 0.5 + 0.5
expected_intensity += 4.0 + 6.0
expected_intensity += 4.0 + 5.0
expected_intensity += 12.0 + 14.0
```
(I hope I'm understanding that correctly.)

pshriwise (changes requested):
Just a few small comments about clarity. Over all this looks good though!

tests/test_r2s.py
zxkjack123: Yes, your understanding is right. I added these in the update commit.

tests/test_r2s.py
```diff
@@ -308,6 +308,39 @@ def test_total_photon_source_intensity():
...
+    # Set up the source density with energy group number of 2
+    m.source_density = IMeshTag(4, float)
+    m.source_density[:] = [[0.0, 0.0, 1.0, 1.0],
+                           [2.0, 2.0, 3.0, 3.0],
+                           [4.0, 4.0, 5.0, 5.0],
+                           [6.0, 6.0, 7.0, 7.0]]
+    intensity = total_photon_source_intensity(m, "source_density", True)
+    #expected intensity: each line represents a mesh voxel
+    #for each subvoxel: voxel_vol * cell_fracs * photon_intensity
+    expected_intensity = 1 * 0.5 * (0.0 + 0.0) + 1 * 0.5 * (1.0 + 1.0)
```
pshriwise, line 337: Nice. This is extremely helpful. Thanks!

pyne/r2s.py
```diff
@@ -180,8 +182,18 @@ def total_photon_source_intensity(m, tag_name):
 
     sd_tag = m.mesh.getTagHandle(tag_name)
     intensity = 0.
-    for idx, _, ve in m:
-        vol = m.elem_volume(ve)
-        ve_data = sd_tag[ve]
-        intensity += vol*np.sum(ve_data)
+    if sub_voxel:
```
gonuke, line 185: Look at reorganizing this conditional and loop

zxkjack123: @gonuke I found that the `max_num_cells` can not be moved outside the conditional statement because the mesh doesn't have a `cell_fracs` tag under r2s mode.

zxkjack123:
Update:
- Move the `num_e_groups` outside the for loop

zxkjack123:
I found that the earlier version of this PR would use a huge amount of memory during the function `total_photon_source_intensities`. For example, when I run a mesh with size of 30\*30\*30, `max_num_cells` is 12, the peak memory consumption could reach about 90GB.

And the memory consumption is gradually increasing in the loops under subvoxel. I think it is becasue of the `m.cell_fracs[idx][svid]` in the loop. The `m.cell_fracs[idx]` will call a [`__getitem__` ](https://github.com/pyne/pyne/blob/dc750cad9cf4f2e938e46a8d2920390417951307/pyne/mesh.py#L370) `IMeshTag`. There is a loop to get the value, it's alos time consuming.

Before the latest commit, there will be `MemoryError` in r2s step2 for a mesh size of 30\*30\*45. Well, the latest commit enables the calculation of r2s step2 of mesh size 30\*30\*45 (tested) . 

By the way, I don't know the exact reason of the huge memory consuming, but the latest commit does fix the problem.

gonuke (approved):

gonuke:
Thanks @zxkjack123

gonuke:
Thanks @zxkjack123

Issue #1070, opened 2018-11-09
feature request: Additional Tallies


shimwell:
It would be great (for me at least) if the src/tally.cpp allowed some additional tallies such as HEAT, DPA, TBR and others as discussed on the DAGMC issue tracker [link](https://github.com/svalinn/DAGMC/issues/413).

A fork with a new branch for implementing this feature [link](https://github.com/Shimwell/pyne/blob/feature_extra_tallies/src/tally.cpp) has been made, I am hoping to commit this to the Pyne repo if people think it is a nice feature to add?

shimwell:
Tests for DPA, TBR and HEAT have been added to tally_test.py and a pull request has been issued.
[https://github.com/pyne/pyne/pull/1071](https://github.com/pyne/pyne/pull/1071)

PR #1071, opened 2018-11-13
Feature extra tallies

shimwell:
Created pull request to address [this](https://github.com/pyne/pyne/issues/1070) issue / feature request 

Extra tallies for TBR, DPA and HEAT have been added to the tally.cpp and associated files

### DIFF
--- a/src/tally.cpp
+++ b/src/tally.cpp
@@ -8,6 +8,7 @@
 
 #ifndef PYNE_IS_AMALGAMATED
   #include "tally.h"
+  #include "material.h"
   #include "particle.h"
 #endif
 
--- a/src/tally.cpp
+++ b/src/tally.cpp
@@ -12,9 +12,9 @@
 #endif
 
 enum entity_type_enum {VOLUME, SURFACE}; // Enumeration for entity types
-enum tally_type_enum  {FLUX, CURRENT};   // Enumeration for tally types
+enum tally_type_enum  {FLUX, CURRENT, TBR, DPA, HEAT};   // Enumeration for tally types
 
-const std::string tally_type_enum2string[] = {"Flux", "Current"};
+const std::string tally_type_enum2string[] = {"Flux", "Current", "TBR", "DPA", "HEAT"};
 const std::string entity_type_enum2string[] = {"Volume", "Surface"};
 
 
@@ -44,9 +44,9 @@ pyne::Tally::Tally() {
 
 // Default constructor
 pyne::Tally::Tally(std::string type, std::string part_name, 
-		   int ent, std::string ent_type, 
-		   std::string ent_name, std::string tal_name,
-		   double size, double norm ) {
+       int ent, std::string ent_type, 
+       std::string ent_name, std::string tal_name,
+       double size, double norm ) {
 
   // Empty Tally Constructor
   tally_type = type;
@@ -65,6 +65,17 @@ pyne::Tally::~Tally() {
 
 
 /*--- Method definitions ---*/
+// is valid functions
+bool pyne::Tally::is_valid(std::string s) {
+ 
+ // todo find some fancy c++ way
+ int num_tallies = sizeof(tally_type_enum2string)/sizeof(tally_type_enum2string[0]);
+ for ( int i = 0 ; i != num_tallies ; i++ ) {
+   if ( s == tally_type_enum2string[i] ) return true;
+ }
+ return false;
+}
+
 //
 void pyne::Tally::from_hdf5(char * filename, char *datapath, int row) {
   std::string fname(filename);
@@ -74,7 +85,7 @@ void pyne::Tally::from_hdf5(char * filename, char *datapath, int row) {
 
 //
 void pyne::Tally::from_hdf5(std::string filename, std::string datapath, 
-			    int row) { 
+          int row) { 
   // line of data to acces
   int data_row = row;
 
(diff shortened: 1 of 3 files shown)


src/tally.cpp
```diff
@@ -65,6 +65,17 @@ pyne::Tally::~Tally() {
 
 
 /*--- Method definitions ---*/
+// is valid functions
+bool pyne::Tally::is_valid(std::string s) {
+ 
+ // todo find some fancy c++ way
+ int num_tallies = sizeof(tally_type_enum2string)/sizeof(tally_type_enum2string[0]);
```
gonuke, line 72: I think if you make `tally_type_enum2string` a `std::vector` of `std::string` your life will be easier.

src/tally.cpp
```diff
@@ -44,9 +44,9 @@ pyne::Tally::Tally() {
 
 // Default constructor
 pyne::Tally::Tally(std::string type, std::string part_name, 
-		   int ent, std::string ent_type, 
-		   std::string ent_name, std::string tal_name,
-		   double size, double norm ) {
+       int ent, std::string ent_type,
```
gonuke, line 47: The style guide indicates that indentation should align with the first argument on the previous line.

src/tally.cpp
```diff
@@ -423,37 +440,59 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
+          output << "FM" << tally_index << "4 " << " 1 " << mat_num << " 205 " << std::endl;
+
+    } else if (tally_type.find("DPA") != std::string::npos ) {
+        output << "F" << tally_index << "4:" << particle_token << " " << entity_id << std::endl;
+        if(mat_num == "?" ) { std::cout << "mat_number unset for a need fm card" << std::endl;}
+          output << "FM" << tally_index << "4 " << " 1 " << mat_num << " 444 " << std::endl;
+
+    } else if (tally_type.find("HEAT") != std::string::npos ) {
+        output << "F" << tally_index << "6:" << particle_token << " " << entity_id << std::endl;
+        output << "FM" << tally_index << "6 " << normalization << std::endl;
```
gonuke, line 495: Earlier options only normalize if `normalization > 1`.  Should it be the same here?

src/tally.cpp
```diff
@@ -423,37 +440,59 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
     } else if (tally_type.find("Current") != std::string::npos) {
-      // makes no sense in mcnp
+      // volume current tally makes no sense in mcnp
+
+    } else if (tally_type.find("TBR") != std::string::npos && particle_token == "n" ) {
+        output << "F" << tally_index << "4:" << particle_token << " " << entity_id << std::endl;
+        if(mat_num == "?" ) { std::cout << "mat_number unset for a need fm card" << std::endl;}
+          output << "FM" << tally_index << "4 " << " 1 " << mat_num << " 205 " << std::endl;
+
+    } else if (tally_type.find("DPA") != std::string::npos ) {
```
gonuke, line 488: Check for neutron here, as with TBR

src/tally.h
```diff
@@ -49,6 +51,8 @@ namespace pyne
 
     ~Tally (); /// default destructor
 
+    // is the proposed name a valid tally name
+    bool is_valid(std::string s);
```
gonuke, line 55: Perhaps `is_supported` is a better name...  a tally could be theoretically valid and not supported, and a user could use a supported tally, but define it in an invalid way...

gonuke (changes requested):
Thanks for adding these @Shimwell !  Next we’ll have to add He production.

In addition to the specific requests here, I think there is a lot of unnecessary (or possibly incorrect) whitespace/indentation change.

src/tally.cpp
```diff
@@ -423,37 +440,59 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
-	output << "FM" << tally_index << "4 " << normalization << std::endl;
+  output << "FM" << tally_index << "4 " << normalization << std::endl;
     } else if (tally_type.find("Current") != std::string::npos) {
-      // makes no sense in mcnp
+      // volume current tally makes no sense in mcnp
+
+    } else if (tally_type.find("TBR") != std::string::npos && particle_token == "n" ) {
+        output << "F" << tally_index << "4:" << particle_token << " " << entity_id << std::endl;
+        if(mat_num == "?" ) { std::cout << "mat_number unset for a need fm card" << std::endl;}
+          output << "FM" << tally_index << "4 " << " 1 " << mat_num << " 205 " << std::endl;
```
bam241, line 486: indentation seems off

src/tally.cpp
```diff
@@ -423,37 +440,59 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
+
+    } else if (tally_type.find("DPA") != std::string::npos ) {
+        output << "F" << tally_index << "4:" << particle_token << " " << entity_id << std::endl;
+        if(mat_num == "?" ) { std::cout << "mat_number unset for a need fm card" << std::endl;}
+          output << "FM" << tally_index << "4 " << " 1 " << mat_num << " 444 " << std::endl;
```
bam241, line 491: same here (indentation)

bam241, line 491: maybe there is a way to combine those sections, as the only difference between them is a `444` swapped for a `205`

src/tally.cpp
```diff
@@ -423,37 +440,59 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
   if (normalization > 1.0)
     output << std::scientific;
 
+  // TODO this could be cleaner
+  std::string mat_num = "?";
+  if ( material ) {
+    int mat_id = material->metadata["mat_number"].asInt();
```
bam241, line 446: You probably also need to check if the `mat_number` is a valid metadata....

tests/test_tally.py
```diff
@@ -42,6 +42,25 @@ def write_arb_p(file_name,path):
...
+
+# writes the neutron tally to arb file and path
+def write_arb_n_heat(file_name,path):
+    tally = Tally("HEAT","Neutron",12,"Volume","Volume 12","Neutron HEAT in Cell 12",35.0)
+    tally.write_hdf5(file_name,path)   
+
+# writes the photon tally to arb file and path
+def write_arb_p_heat(file_name,path):
+    tally = Tally("HEAT","Neutron",12,"Volume","Volume 12","Neutron HEAT in Cell 12",35.0)
+    tally.write_hdf5(file_name,path)
```
bam241, line 63: this adds method to compute and write the tallies, but I don't see them used in the test.
Are those tests missing ?

bam241 (changes requested):
@Shimwell  thx for putting this in.

There is some indentation to fix as well.

shimwell:
I'm not quite sure how to bring the PR back to life as the original source branch appears to have got lost. The PR is reporting an unknown source and I don't appear to have any Branches on my fork of Pyne.

PR #1072, opened 2018-11-19
add cell_fracs tag for the default r2s

zxkjack123:
Adding the `cell_number` and `cell_fracs` tag for default R2S is helpful for people to check the cells in a voxel and to do better visualization. Therefore, the `cell_number` and `cell_fracs` tags are added for default R2S in this PR.
Users can use `mesh.cell_number[:]` and `mesh.cell_fracs[:]` to see their contents after they are added in the r2s `step1` and `step2`.

### DIFF
--- a/pyne/r2s.py
+++ b/pyne/r2s.py
@@ -99,9 +99,8 @@ def irradiation_setup(flux_mesh, cell_mats, alara_params, tally_num=4,
 
     if m.structured:
         cell_fracs = discretize_geom(m, num_rays=num_rays, grid=grid)
-        # tag cell fracs
-        if sub_voxel:
-            m.tag_cell_fracs(cell_fracs)
+        # tag cell fracs for both default and subvoxel r2s modes
+        m.tag_cell_fracs(cell_fracs)
     else:
         cell_fracs = discretize_geom(m)
 
--- a/news/add_cell_fracs_tag_default_r2s.r2st
+++ b/news/add_cell_fracs_tag_default_r2s.r2st
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* Add cell_fracs and cell_number tags for both default and subvoxel r2s modes
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None


pyne/r2s.py
```diff
@@ -99,9 +99,8 @@ def irradiation_setup(flux_mesh, cell_mats, alara_params, tally_num=4,
 
     if m.structured:
         cell_fracs = discretize_geom(m, num_rays=num_rays, grid=grid)
-        # tag cell fracs
-        if sub_voxel:
-            m.tag_cell_fracs(cell_fracs)
+        # tag cell fracs for both default and subvoxel r2s modes
+        m.tag_cell_fracs(cell_fracs)
```
moatazharb, line 103: tag_cell_fracs doesn't order cells by number across all mesh voxels' tags. let us assume we have a cell tag of length 3 and we have three cells; 21; 24; 40. Two voxels could have their tags as: [24,40,-1] and [40, -1, -1], where -1 corresponds to void. This is fine but for visualization I guess it'd be better to have cell numbers ordered across all voxels' tags. From the example above; like [-1,24,40] and [-1, -1,40].
This way when using expand_tags.py we would get a number of tags corresponding to the length of the cell_fracs tag and each would be corresponding to only one cell. From the example above; cell_fracs_002 would correspond to cell 40, and so on.
To do this it requires an edit to tag_cell_fracs fuction in pyne.mesh.

gonuke: I think this requires a different solution @moatazharb.  The shape of `cell_fracs` Is `(num_mesh_elements, max_num_cells_per_voxel)`.  Therefore, in a problem in which there are C cells, but each mesh element has no more than m << C cells, this question of order would be irrelevant.

gonuke: We’ll need special purpose tools to unpack this for visualization.

gonuke: However, `tag_cell_fracs` automatically tags the cell with the largest fraction and its fraction...

gonuke:
Thanks @zxkjack123 - this little change will helps a number of people (including @cadangelo already)

PR #1074, opened 2018-11-29
add more guidance to the required format of the MCNP meshtal file

gonuke:
The meshtal reader only supports a particular format of mesh tally output - this PR makes that clear in the help string of this script.

### DIFF
--- a/scripts/meshtal_to_mesh
+++ b/scripts/meshtal_to_mesh
@@ -8,7 +8,8 @@ def main():
              'Reads an MCNP meshtal file and creates an h5m mesh file '
              'for each meshtally within the file. The output mesh files are '
              'named <filename>_tally_<tally_num>.h5m. Note that this script '
-             'only works for Cartesian meshes.'))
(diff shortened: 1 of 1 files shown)


gonuke:
This seems to be failing on a timeout with accessing data - anyone know how to fix this?  It's a small documentation change, so I'm confused about why this is happening?

pshriwise:
@gonuke I've seen that too on a few other PRs. I've been restarting them when I see this happening. It seems the links aren't super reliable and this probably isn't related to the change here.

zxkjack123:
@gonuke There are too many files to download during the tests. It will report an error even if there is only one timeout failure. So, the timeout failures happen a lot.
I found that several links are to be blamed for most cases.  I think adding some mirror backup links in pyne/data for these files with high failure probability would reduce the timeout failure.

gonuke:
Looks like this finally passed tests...

zxkjack123:
@gonuke LGTM.

PR #1077, opened 2018-12-06
Total photon source intensities combine

zxkjack123:
In voxel R2S, there isn't a `cell_fracs` tag, which exists in sub-voxel R2S. So the code for calculating the total photon source intensities are divided into two parts.
Now, a local `cell_fracs` tag for voxel R2S is created, then we can combine the two loops for calculating the total photon source intensities.


gonuke (approved):
I definitely like the way this simplifies things

gonuke:
Thanks @zxkjack123

PR #1080, opened 2019-01-03
Add description ebounds source sampling

zxkjack123:
An "e_bounds" file is required when the `source_sampling` are applied. This PR add a short description to remind the user to provide the "e_bounds" file.


gonuke:
Thanks for this documentation update @zxkjack123 - not enough people make these important improvements

PR #1084, opened 2019-01-08
Fix formulas rendering

bam241:
@gonuke 

the `imgmat` extension prevent the proper rendering of the formulas... (see the http://pyne.io/theorymanual/variance_reduction.html?highlight=variance)

I tested this locally, I got the formulas back...


gonuke:
Thanks @bam241

PR #1087, opened 2019-01-09
remove requirement that cmake has SSL support

kkiesling:
This quick fix allows Cmake to be used to download the data that does not have SSL support. Prior to this, I was not able to successfully download the decay and cram data due to lack of SSL support.


gonuke:
Thanks @kkiesling

PR #1088, opened 2019-01-09
Add a dummy file to test endl loading method

bam241:
This add a dummy file to test `endl` `loadfile()` in order to avoid depending on IAEA website for the CI testing (from which we got a lot a timeout....)


bam241:
cc @gonuke

PR #1089, opened 2019-01-15
Consistency fix between Python API and C++ API

bam241:
This changes the default value of the `datapath` name when reading and writing hdf5 mat library using C++ to match the the default value of the Python API.

### DIFF
--- a/src/material.h
+++ b/src/material.h
@@ -139,7 +139,7 @@ namespace pyne
     /// \param datapath Path to the the material in the file.
     /// \param row The index to read out, may be negative.
     /// \param protocol Flag for layout of material on disk.
-    void from_hdf5(std::string filename, std::string datapath="/material",
+    void from_hdf5(std::string filename, std::string datapath="/materials",
                                                           int row=-1, int protocol=1);
 
     /// Writes this material out to an HDF5 file.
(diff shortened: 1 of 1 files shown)


gonuke:
It appears that this **IS** already consistent with the python API - this change makes it inconsistent.   But there is a twist....

The current behavior in C++ is consistent with the `Material.write_hdf5()` method and `Material.from_hdf5()` has no default but instead always requires an argument.

The `MaterialLibrary.write_hdf5()` and `MaterialLibrary.from_hdf5()` methods uses the plural, consistent with the proposal here.  

To be clear, there is currently an inconsistency within the python API between `Material.write_hdf5()` and `MaterialLibrary.write_hdf5()`

I agree that all of them should be made consistent.

bam241:
I stand corrected! :)

but @gonuke, is it inconstant ?!
`write_hdf5()` method in both the C++  and the Python API are supposed to write a single marterial in a `hdf5` file...
where `write_hdf5()` method aims to write multiple material in a `hdf5` file.

So maybe the default as "/material" for the `material.write_df5()` (in both C++ and Python API)  and "/materials" for the `MaterialLibrary.write_hdf5()` make sense ?!

gonuke:
Sure - it kind of makes sense semantically, but it also makes sense to have all materials in one place.

That said...  Given the way it writes a master list of nuclides and doesn't appear to have a good way to update that list on subsequent writes, it may be necessary to do this differently between single materials and materilal libraries.

gonuke:
Did we decide to withdraw this?  The data structure for a single material is different than for a material library... and you are working on updating the MaterialLibrary code anyway?

bam241:
not relevant anymore...

PR #1090, opened 2019-01-15
Script to Merge multiple material lib files into a single one

bam241:
This script allows to merge multiple PyNE HDF5 materials libraries into a single one.

cc @pshriwise

### DIFF
--- a/scripts/merge_material_libraries.py
+++ b/scripts/merge_material_libraries.py
@@ -1,0 +1,48 @@
+from pyne.material import MaterialLibrary
+
+"""
+
+This script allows to merge multiple PyNE material hdf5 file into a single one.
+"""
+
+def merge_material_library(merged_libname, matlibs, datapaths=[], nucpaths=[]):
+    """ Merge the different hdf5 PyNE material libraries into a single one and
+    write the merged library in hdf5 format
+
+    Parameters:
+    merged_libname (str): name of the new library
+    matlibs ([str]): list of the material library name to be merged
+    datapaths ([str] -- optional ): list of the datapath for each library to
+    merge, using \"/materials\" as default
+    nucpaths ([str] -- optional ): list of the nucpath for each library to
+    merge, using \"/nucid\" as default
+
+    """
+
+
+    if len(datapaths) == 0:
+        print("No datapaths provided, using \"/Materials\" as default.")
+        for i in range(0, len(matlibs)):
+            datapaths.append("/Materials")
+    elif len(datapaths) != len(matlibs):
+        print("!Error! Number of matlibs does not match number of datapaths")
+
+    if len(nucpaths) == 0:
+        print("No nucpaths provided, using \"/nucid\" as default.")
+        for i in range(0, len(matlibs)):
+            nucpaths.append("/nucpaths")
+    elif len(nucpaths) != len(matlibs):
+        print("!Error! Number of matlibs does not match number of nucpaths")
+
+    merged_mat_library = MaterialLibrary()
+
+    for index, (filename, datapath, nucpath) in enumerate(zip(matlibs, datapaths, nucpaths)):
+        mat_lib = MaterialLibrary()
+        mat_lib.from_hdf5(filename, datapath, nucpath)
+        for item in mat_lib.items():
+            merged_mat_library.__setitem__(item[0], item[1])
+
+    merged_mat_library.write_hdf5(
+        merged_libname, datapath="/materials", nucpath="/nucid")
+
+


scripts/merge_material_libraries.py
```diff
@@ -0,0 +1,48 @@
...
+    Parameters:
+    merged_libname (str): name of the new library
+    matlibs ([str]): list of the material library name to be merged
+    datapaths ([str] -- optional ): list of the datapath for each library to
+    merge, using \"/materials\" as default
+    nucpaths ([str] -- optional ): list of the nucpath for each library to
+    merge, using \"/nucid\" as default
+
+    """
+
```
pshriwise, line 21: We can probably remove an empty line or two here

scripts/merge_material_libraries.py
```diff
@@ -0,0 +1,48 @@
...
+
+    if len(datapaths) == 0:
+        print("No datapaths provided, using \"/Materials\" as default.")
+        for i in range(0, len(matlibs)):
+            datapaths.append("/Materials")
+    elif len(datapaths) != len(matlibs):
+        print("!Error! Number of matlibs does not match number of datapaths")
```
pshriwise, line 28: If this is truly an error, the function should probably throw an exception here with the message in the print statement.

scripts/merge_material_libraries.py
```diff
@@ -0,0 +1,48 @@
...
+
+    if len(nucpaths) == 0:
+        print("No nucpaths provided, using \"/nucid\" as default.")
+        for i in range(0, len(matlibs)):
+            nucpaths.append("/nucpaths")
+    elif len(nucpaths) != len(matlibs):
+        print("!Error! Number of matlibs does not match number of nucpaths")
```
pshriwise, line 35: Same as the error message above.

scripts/merge_material_libraries.py
```diff
@@ -0,0 +1,48 @@
...
+
+    merged_mat_library = MaterialLibrary()
```
pshriwise, line 37: Nice. This section containing the main functionality is really clean, like we hoped!

scripts/merge_material_libraries.py
```diff
@@ -0,0 +1,48 @@
+from pyne.material import MaterialLibrary
+
+"""
+
```
pshriwise, line 4: Do we need this blank line?

scripts/merge_material_libraries.py
```diff
@@ -0,0 +1,48 @@
...
+This script allows to merge multiple PyNE material hdf5 file into a single one.
```
pshriwise, line 5: I wouldn’t technically call this a script. I like the capability of the `merge_material_library` function a lot - so much so I think it should become part of PyNE :) But here I would expect to see a 

```python
if __name__ == “__main__”:
```

somewhere in here with options for `datapaths` and `nucpaths` along with an option for the output filename etc.

scripts/merge_material_libraries.py
pshriwise: In fact, I’m wondering if it might be worth adding a function to the `pyne.Material` module for merging libraries. Better yet, it might be cool to be able to hijack the `+` operator and add libraries together (`mat_lib1` + `mat_lib2`) if you agree that might be appropriate.

pshriwise (changes requested):
A couple thoughts here. After writing up the part about making this a `MaterialLibrary` operation, I’m more convinced that we should do that.

scripts/merge_material_libraries.py
```diff
@@ -0,0 +1,48 @@
...
+            nucpaths.append("/nucpaths")
+    elif len(nucpaths) != len(matlibs):
+        print("!Error! Number of matlibs does not match number of nucpaths")
+
+    merged_mat_library = MaterialLibrary()
+
+    for index, (filename, datapath, nucpath) in enumerate(zip(matlibs, datapaths, nucpaths)):
+        mat_lib = MaterialLibrary()
+        mat_lib.from_hdf5(filename, datapath, nucpath)
+        for item in mat_lib.items():
```
gonuke, line 42: Is there a reason for not using 
```
for mat_name, material in mat_lib.items():
    merged_mat_libarary[mat_name] = material
```
?

gonuke (commented):
What is the expected/desired behavior when two libraries both contain the same entry?  Will the last one overwrite any previous ones?  Do we care?  Should there be a warning?

scripts/merge_material_libraries.py
pshriwise: Not that I can see.

bam241:
this might not be required anymore and should ultimately be covered by #1102

gonuke:
Now replaced by #1104

bam241:
yes ! closing...

PR #1091, opened 2019-01-15
Backport changes from DAGMC

ljacobson64:
[DAGMC](https://github.com/svalinn/DAGMC) contains an amalgamated version of pyne. The last time it was updated to match pyne/develop was about 2.5 years ago. Since then, we have made a few changes and bug fixes that never made it back upstream. This PR contains those changes.

Here is a list of what is changed:

1. Updated the copyright year to 2019
2. Updated the URLs referencing MOAB
3. Add an include guard in `particle.cpp` to avoid including `particle.h` when using amalgamated pyne
4. Changed the MCNP particle identifiers to be lower case in `particle.cpp`
5. Replaced some extraneous commas with semicolons in `particle.cpp`
6. Added a newline in `state_map.cpp`


gonuke:
Thanks @ljacobson64. 

This is failing tests because they expect the lower case particle identifier.  Consider either reverting that change or updating the tests.

ljacobson64:
Thanks for taking a look at the CI... I chose to update the tests.

gonuke:
Thanks @ljacobson64

PR #1092, opened 2019-01-20
Atomic mass update issue #1057

makoziol0:
Replaced the old mass file with the newer version from #1057 and redid some tests to match.


makoziol0:
Doesn't seem like the automated test is getting the new mass file data loaded in. Although when I do it manually on my computer it works fine. Advice? @katyhuff

cmake/ReplicatePythonSourceTree.cmake
```diff
@@ -12,5 +12,6 @@ file(COPY setup.py setup_sub.py pyne DESTINATION "${CMAKE_ARGV3}"
                     PATTERN "*.pyi"
                     PATTERN "*.pxd"
                     PATTERN "*.mas12"
```
gonuke: If we delete the 2012 file, we can probably remove this line.

gonuke (commented):
Just one little comment.

As for failing tests - I wonder if the continuous integration tests are not incorporating this new data and are using old data in the docker image - maybe @bam241 or @kkiesling knows and/or can confirm?

makeclean:
one should also run atomc_data_gen.py and commit the new cpp files

makoziol0:
> one should also run atomc_data_gen.py and commit the new cpp files

@makeclean 
The atomic_data.cpp and .h files from atomicgen.py are in the gitignore file so I think they are excluded on purpose.

cmake/ReplicatePythonSourceTree.cmake
makoziol0: Removed.

gonuke:
I agree with @makoziol0 on this - the CircleCI reports that new `atomic_data.h` and `atomic_data.cpp` files were generated by `atomicgen.py` during build.

I poked around a little further in the build/test process.
1. we do not invoke `nuc_data_make` during CI - we use a pre-existing nuclear data file
2. the pre-existing nuclear data file has the old atomic mass data

I think we need to use a special procedure to merge this and also update the nuclear data file in the CI image.

katyhuff:
Ok -- so ... I think we should open an issue describing the problem that is blocking this PR merge (continuous integration tests are not incorporating the new data introduced by the PR and are instead using old data in the docker image ). The student who submitted this PR is no longer working with me. Perhaps @Mongwell @aruzzo could take a look at this and:

- [x] check to see if any existing PyNE issue would address these CI failures (continuous integration tests are not incorporating the new data introduced by the PR and are instead using old data in the docker image )
- [x] create an issue describing how CI should be updated (e.g. should we run nuc_data_make in CI? or, perhaps only do so if a particular PR flag is present... ?)
- [x] take some steps to try solving this issue... so that then this pr can pass the tests and be merged...

gonuke:
Should we have a meeting to talk through this?  It seems like something that would benefit from having a lot of heads together to untangle some of the dependencies?

katyhuff:
Let me see if I can grasp it all by meeting with @aruzzo  and @Mongwell here at Illinois, then maybe we can get back to this once it's more coherent.

gonuke:
Sure thing... it’s seems that they have a pretty good handle on it, but others (read: me) need to be brought up to speed to help merge it all

gonuke:
replaced by #1281

PR #1094, opened 2019-01-23
disable check

zxkjack123:
The 'cell_fracs' and 'cell_number' tags are kept for both voxel and subvoxel mode of R2S after the PR #1072 . Therefore, this PR disables the code to check the cell_fracs tag under voxel mode.


gonuke:
Thanks @zxkjack123

PR #1098, opened 2019-02-18
Add include guards around measure.h for amalgamated pyne

ljacobson64:
`measure` cannot be added to amalgamated pyne without adding include guards.

### DIFF
--- a/src/measure.cpp
+++ b/src/measure.cpp
@@ -25,7 +25,9 @@
 
 #include <math.h>
 
+#ifndef PYNE_IS_AMALGAMATED
 #include "measure.h"
+#endif
 
 class CartVect {
   private:
--- a/src/source_sampling.h
+++ b/src/source_sampling.h
@@ -37,7 +37,9 @@
 
 #include "moab/Range.hpp"
 #include "moab/Core.hpp"
+#ifndef PYNE_IS_AMALGAMATED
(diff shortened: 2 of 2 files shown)


gonuke:
This looks OK to me, but can you provide more information about why the include guards inside `measure.h` are not enough?  It's probably something simple, but I'm not seeing it...

ljacobson64:
In amalgamated pyne, there is no file called `measure.h` (or anything else `.h` either); there is only `pyne.h`. So all the `.cpp` files that we add to `pyne.cpp` need to have these include guards to avoid trying to include files that don't exist.

gonuke:
test now failing?  Is this post pymoab merge?

ljacobson64:
Yeah CI appears to be broken...

bam241:
@ljacobson64  CI should be fixed now, could you redo a dummy commit ?

(explanation: for whatever reason, ci version 2.1 now needs to be enabled form circle-ci website configuration [here](https://circleci.com/gh/pyne/pyne/edit#advanced-settings), I enable version 2.1 for the PyNE repo so it should be good now...)

ljacobson64:
@bam241 thanks for fixing this!

gonuke:
Thanks @ljacobson64

PR #1100, opened 2019-02-19
Add some diagnostic information when source cell not found.

gonuke:
In rare circumstances, an MCNP job will fail when sampling a mesh-based source with the message "Nonsense source cell stopping".  It currently offers no additional information.

This PR adds some output information to help understand when and why this is happening, including the particle history number and x,y,z coordinates of the point in question

### DIFF
--- a/share/source.F90
+++ b/share/source.F90
@@ -37,7 +37,9 @@ function find_cell() result(icl_tmp)
     ! icl is now set
 
     if(icl_tmp .le. 0) then
-      write(*,*) 'Nonsense cell number stopping'
+       write(*,*) 'Nonsense cell number: ', icl_tmp
+       write(*,*) 'Looking for cell at position: ', xxx, yyy, zzz
+       write(*,*) 'Stopping'
       stop
     endif
     ! icl now set to be valid cell
--- a/share/source.F90
+++ b/share/source.F90
@@ -38,7 +38,7 @@ function find_cell() result(icl_tmp)
 
     if(icl_tmp .le. 0) then
        write(*,*) 'Nonsense cell number: ', icl_tmp
-       write(*,*) 'Looking for cell at position: ', xxx, yyy, zzz
+       write(*,*) 'Looking for cell at position: ', xxx, yyy, zzz, ' for history ', nps
        write(*,*) 'Stopping'
       stop
     endif


ljacobson64:
This LGTM but we should figure out what our plan will be so these changes are reflected in DAGMC

gonuke:
To get these changes in DAGMC, don’t we just pull a new copy of this file?

ljacobson64:
Yeah, we can do that. It is slightly complicated by the fact that MCNP5 and MCNP6 need to have different source.F90 files though.

gonuke:
Does PyNE have 2 versions?

ljacobson64:
No, the MCNP6 version only exists in DAGMC right now (it was part of the PR you approved earlier today)

bohmt:
Is there any issue with this feature?  This would really help me diagnose some issues with production calculations of SDR.

share/source.F90
```diff
@@ -37,7 +37,9 @@ function find_cell() result(icl_tmp)
     ! icl is now set
 
     if(icl_tmp .le. 0) then
-      write(*,*) 'Nonsense cell number stopping'
+       write(*,*) 'Nonsense cell number: ', icl_tmp
```
zxkjack123, line 40: As `(icl_tmp .le. 0)` is true, then `icl_tmp` is `-1`, which indicates that `icl_tmp` is not changed. We don't need to print `icl_tmp` here.

zxkjack123:
I think we shouldn't put this diagnostic information in pyne/pyne/develop. It should work as a special branch for developers to debug the problem.
This `Nonsense source cell` error happens when `find_cell` can't find out the `icl` of the source partilce with specific `xxx, yyy, zzz`, `chkcel` says this particle doesn't in any cell. This error has the follwing characteristics:
- It hasspens with low possiblility
- It is geom, and photon source dependent. For some geometry, there isn't this problem, for some geometry, it happens with a low possiblity.

I assume this error has little influence to reulsts, the best option for user is to skip this `bad source` with a warning message, but not stop at it. As for the developers, we should use content in this PR as a individual branch, repeat  the error and fix it.

However, I can't repeat this problem with some geometries I used. Someone who can repeat this error should share the geom.h5m, input file and the source.h5m to help to debug it.

share/source.F90
gonuke: Agreed

gonuke:
@zxkjack123 - could you PR a change that doesn’t cause this to stop, but instead resamples?

zxkjack123:
@gonuke Here #1130 is the re-sample version of this PR.

gonuke:
withdrawn in favor of #1130

PR #1102, opened 2019-02-27
[WIP] Add MaterialLibrary to the C++ API

bam241:
This is an attempt to add MaterialLibrary to the C++ API.

For the moment, the python API is not bind to it.

I also made some assumptions about how should behave the `write_hdf5` method. For the moment, I believe the behavior is very similar to the Python one: 
- if the file exists, try to write the materials in it regardless of what is in the file, 
- create the file otherwise.

For the `from_hdf5`, I made in in a way to in append the new material in the existing MaterialLibrary, allowing to merge multiple MaterialLibrary (form the file) into a single one.

To be done:
- [ ] Unit test
- [ ] Debug
- [ ] Improve the behavior when attempting to write in a file already containing some materials

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -14,6 +14,7 @@ set(PYNE_SRCS
   "jsoncpp.cpp"
   "jsoncustomwriter.cpp"
   "material.cpp"
+  "material_library.cpp"
   "nucname.cpp"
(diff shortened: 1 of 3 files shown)


bam241:
closing what to try the draft PR thing !

PR #1106, opened 2019-02-28
It is about time !

bam241:
Enforce C++11 as a standard for compilation... (I think it is time...)


gonuke (approved):
thanks @bam241

pshriwise (approved):
Nice :)

Issue #1105, opened 2019-02-28
FileNotFound returning invalid char pointer


pshriwise:
The `FileNotFound` error message is returned via `const char*` which is created by a string that goes out of scope in its error message function `what()`. When that string goes out of scope, the pointer is no longer valid and has been reported to cause garbled text by @zxkjack123. This string should be kept in memory by adding an attribute to the class so that string exists for the lifetime of the exception object.

zxkjack123:
Well this error is actually found by @ljacobson64 in #1099 , here:
https://github.com/pyne/pyne/pull/1099#issuecomment-466673262

zxkjack123:
There isn't a `FileNotFound` error in python [build-in exceptions](https://docs.python.org/2/library/exceptions.html#module-exceptions), but I find that there is an [IOError](https://docs.python.org/2/library/exceptions.html#exceptions.IOError).

Throw an `IOError` is at least a better way than throw a `runtime_error` when the `file not found` happens.

pshriwise:
This has been addressed in #1107. Thanks @gonuke!

PR #1107, opened 2019-02-28
Fix fnf

gonuke:
The `FileNotFound` exception relied on a local scope `std::string` that went out of scope before returning its `(const char*)` contents.

This string has been moved to be a member variable of the exception.

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -165,16 +165,16 @@ namespace pyne {
     /// Creates a helpful error message.
     virtual const char* what() const throw()
     {
-      FNFstr = "File not found: ";
+      FNF_message = "File not found: ";
       if (!filename.empty())
-        FNFstr += filename;
+        FNF_message += filename;
 
-      return (const char *) FNFstr.c_str();
+      return (const char *) FNF_message.c_str();
     };
 
   private:
     std::string filename; ///< unfindable filename.
-    std::string FNFstr; /// Message for exception
+    std::string FNF_message; /// Message for exception
   };
 
   /// Exception representing value errors of all kinds
--- a/src/utils.h
+++ b/src/utils.h
@@ -165,7 +165,7 @@ namespace pyne {
     /// Creates a helpful error message.
     virtual const char* what() const throw()
     {
-      std::string FNFstr ("File not found: ");
+      FNFstr = "File not found: ";
       if (!filename.empty())
         FNFstr += filename;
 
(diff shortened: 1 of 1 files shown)


src/utils.h
```diff
@@ -157,23 +157,23 @@ namespace pyne {
...
       filename = fname;
     };
 
     /// Creates a helpful error message.
     virtual const char* what() const throw()
     {
-      std::string FNFstr ("File not found: ");
       if (!filename.empty())
-        FNFstr += filename;
+        FNF_message += filename;
```
pshriwise: I couldn't get this to compile. I think because the method is `const`? 

I'd say we could remove `const` but the `std::exception` method it inherits from is also `const`. As we probably want to match that interface as closely as possible, I'd recommend adding `mutable` in front of the FNF_message declaration.

src/utils.h
```diff
@@ -157,23 +157,23 @@ namespace pyne {
...
 
-      return (const char *) FNFstr.c_str();
+      return (const char *) FNF_message.c_str();
```
pshriwise: `c_str` returns a `const char *` by default, so the c-style cast here can go

src/utils.h
```diff
@@ -157,23 +157,23 @@ namespace pyne {
     ~FileNotFound () throw () {};
 
     /// constructor with the filename \a fname.
-    FileNotFound(std::string fname)
+    FileNotFound(std::string fname) : FNF_message("File not found: ")
```
pshriwise: Tiny thing, but way as well initialize `filename` and `FNF_message` in the same way.

pshriwise (changes requested):
Thanks for the prompt fix to #1105 to @gonuke! Just a couple small comments.

src/utils.h
pshriwise: Or maybe I'm missing something here. LMK!

gonuke: Yeah - I'm using CI for testing :blush:

gonuke: Fixed this by setting the full message in the constructor

src/utils.h
gonuke: Yes - now do it in the flow (not with initializer) in order to capture to fully set the message using the filename

src/utils.h
gonuke: cleaned this up

pshriwise (approved):

PR #1103, opened 2019-02-28
[WIP] PHITS material

bam241:
Adding the capability to form PHITS input material section.

Documentation about PHITS material sections can find [[here](https://phits.jaea.go.jp/manual/manualE-phits.pdf)] 

Remaining work:
- [ ] Add unit test
- [ ] debug

PR #1108, opened 2019-02-28
PHITS material

bam241:
reopening #1103 as a draft :)
Adding the capability to form PHITS input material section.

Documentation about PHITS material sections can find [[here](https://phits.jaea.go.jp/manual/manualE-phits.pdf)] 

Remaining work:
- [x] Add unit test
- [x] debug


bam241:
@pshriwise @gonuke this should be good to review !

pshriwise:
It'll be a day or two for me to get to it, but it's on my list!

src/material.cpp
```diff
@@ -670,14 +670,136 @@ std::string pyne::Material::mcnp(std::string frac_type) {
...
+
+  // 'name'
+  if (metadata.isMember("name")) {
+    oss << "C name: " << metadata["name"].asString() << std::endl;
+  }
+  // Metadata comments
+  if (metadata.isMember("comments")) {
+    std::string comment_string = "comments: " + metadata["comments"].asString();
+    // Include as is if short enough
+    if (comment_string.length() <= 77) {
```
gonuke: remove magic number '77'

gonuke: make a function that will do this line wrapping?

src/material.cpp
```diff
@@ -670,14 +670,136 @@ std::string pyne::Material::mcnp(std::string frac_type) {
...
+  if (metadata.isMember("mat_number")) {
+    int mat_num = metadata["mat_number"].asInt();
+    oss << mat_num;
+  } else {
+    oss << "?";
+  }
+  oss << " ]" << std::endl;
+
+  // check for metadata
+  if (metadata.isMember("GAS")){
```
gonuke: do these (at least the integers) with a loop over a set of keywords

src/material.cpp
```diff
@@ -670,14 +670,136 @@ std::string pyne::Material::mcnp(std::string frac_type) {
...
+
+  // Set up atom or mass frac map
+  std::map<int, double> fracs;
+  std::string frac_sign;
+
+  if ("atom" == frac_type) {
+    if (density != -1.0) {
+      fracs = to_atom_dens();
+      for (comp_iter ci = fracs.begin(); ci != fracs.end(); ci++){
+        ci->second *= pyne::cm2_per_barn; // unit requirememt is [10^24 atoms/cm3]
```
gonuke:
```suggestion
        ci->second *= pyne::cm2_per_barn; // unit requirememt is [10^24 atoms/cm3] = [atoms/b.cm]
```

src/material.cpp
```diff
@@ -670,14 +670,136 @@ std::string pyne::Material::mcnp(std::string frac_type) {
...
+  // COND should be "<" or "=" or ">" if present
+  if (metadata.isMember("COND")){
+    oss << "     COND" << metadata["COND"].asString() << "0" << std::endl;
+  }
+
+  // Set up atom or mass frac map
+  std::map<int, double> fracs;
+  std::string frac_sign;
+
+  if ("atom" == frac_type) {
```
gonuke: If this is the same as MCNP, perhaps write a function that does this part and re-use

gonuke (changes requested):
Probably a few places to make common MCNP/PHITS methods and re-use

src/material.cpp
```diff
@@ -670,14 +670,136 @@ std::string pyne::Material::mcnp(std::string frac_type) {
       // Spaces are important for tests
       table_item = metadata["table_ids"][nucmcnp].asString();
       if (!table_item.empty()) {
-	oss << "     " << mcnp_id << "." << table_item << " ";
+  oss << "     " << mcnp_id << "." << table_item << " ";
```
gonuke: This whitespace/indentation looks wrong

gonuke (changes requested):
Double check some of your indentation...

bam241:
I had to change some of the mcnp tests, as when the density is present in the material the fraction of the nuclide are the fraction density.
It does not change anything for the mcnp calculation (as it is renormalized), but in the input all the fraction are multiplied by the density...

bam241:
@gonuke this should be good for an other round of review :)

src/material.cpp
```diff
@@ -15,7 +15,7 @@
 
 // h5wrap template
 template double h5wrap::get_array_index(hid_t, int, hid_t);
-
+const int mcnp_line_lenght = 77;
```
gonuke:
```suggestion
const int mcnp_line_length = 77;
```

and everywhere that its used

src/utils.cpp
```diff
@@ -72,15 +66,11 @@ std::string pyne::to_str(bool t) {
 }
 
 
-int pyne::to_int(std::string s) {
-  return atoi( s.c_str() );
-}
+int pyne::to_int(std::string s) { return atoi(s.c_str()); }
```
gonuke: I think our style guide prefers this to be over multiple lines

src/material.cpp
```diff
@@ -641,14 +627,90 @@ std::string pyne::Material::mcnp(std::string frac_type) {
...
+  std::string frac_sign = "";
 
-  if ("atom" == frac_type) {
-    fracs = to_atom_frac();
-    frac_sign = "";
+  if ("atom" != frac_type) {
+    frac_sign = "-";
+  }
+
+  // iterate through frac map
```
gonuke: These blocks are probably common between MCNP & PHITS and can be modularized

gonuke (changes requested):

src/material.cpp
```diff
@@ -641,27 +627,81 @@ std::string pyne::Material::mcnp(std::string frac_type) {
...
 
-  if ("atom" == frac_type) {
-    fracs = to_atom_frac();
-    frac_sign = "";
+  if ("atom" != frac_type) {
+    frac_sign = "-";
+  }
+
+  // write the frac map
+  oss << mcnp_frac(fracs, frac_sign);
```
gonuke: Why not pass in `frac_type` and determine the sign in `mcnp_frac`?

src/utils.cpp
```diff
@@ -90,198 +82,200 @@ double pyne::endftod_cpp(char * s) {
...
-void pyne::use_fast_endftod() {
-  pyne::endftod = &pyne::endftod_cpp;
-}
+void pyne::use_fast_endftod() { pyne::endftod = &pyne::endftod_cpp; }
 
 std::string pyne::to_upper(std::string s) {
   // change each element of the string to upper case.
-  for(unsigned int i = 0; i < s.length(); i++)
-    s[i] = toupper(s[i]);
+  for (unsigned int i = 0; i < s.length(); i++) s[i] = toupper(s[i]);
```
gonuke: There are a bunch of style changes here.  Does the style guide have an opinion, or are they arbitrary.  I'd rather not mix these changes in to the PR.

src/utils.cpp
```diff
@@ -90,198 +82,200 @@ double pyne::endftod_cpp(char * s) {
...
   s[0] = toupper(s[0]);
   // change each subsequent element of the string to lower case
-  for(unsigned int i = 1; i < slen; i++)
-    s[i] = tolower(s[i]);
+  for (unsigned int i = 1; i < slen; i++) s[i] = tolower(s[i]);
   return s;
 }
 
+std::ostringstream pyne::line_wrapping(std::string line,
+                                       int line_lenght) {
```
gonuke:
```suggestion
                                       int line_length) {
```
and matching changes throughout

src/utils.cpp
```diff
@@ -90,198 +82,200 @@ double pyne::endftod_cpp(char * s) {
...
   // uppercase the first character
   s[0] = toupper(s[0]);
   // change each subsequent element of the string to lower case
-  for(unsigned int i = 1; i < slen; i++)
-    s[i] = tolower(s[i]);
+  for (unsigned int i = 1; i < slen; i++) s[i] = tolower(s[i]);
   return s;
 }
 
+std::ostringstream pyne::line_wrapping(std::string line,
```
gonuke: Since this only works for comments: 
```suggestion
std::ostringstream pyne::comment_line_wrapping(std::string line,
```
Maybe the user can provide a comment prefix, with the default being `"C "`

gonuke (changes requested):
Nothing major, but a few small requests.

src/utils.cpp
bam241: old habits sorry....

gonuke:
I'm nearly ready to merge this, but it needs a news file

bam241:
thx for the reviews 
done @gonuke

bam241:
this is weird @gonuke merging your PR did not trigger the tests...

src/utils.cpp
```diff
@@ -163,6 +163,25 @@ std::string pyne::to_lower(std::string s) {
   return s;
 }
 
+std::ostringstream pyne::comment_line_wrapping(std::string line,
+                                               std::string comment_prefix,
+                                               int line_length) {
+  std::ostringstream oss;
+
+  line_length -= comment_prefix.length();
```
bam241, line 171: @gonuke this line breaks the tests.

It does make sense tho. but were where not accounting for the characters size before...
Should I keep it, and fix the test or shall we ignore it for the moment, make an issue an fix it in an other PR ?

gonuke: It was accounted for in the choice of the line length = 77.  You could possibly change the line length to 79 and it will be fixed

bam241: I fixed it, I also reflected the limit of 79 characters to the default

gonuke (approved):

gonuke:
Thanks @bam241

Issue #1109, opened 2019-03-07
pass back list of cells from particle_birth


gonuke:
The standard usage of the source sampling routines requires a search for the source cell after determining the x,y,z location.  This is currently a linear search over all the cells in a problem.

However, we already know which source voxel this is, and therefore we know what possible cells could exist at this point, typically many fewer than the number of cells in a problem.  If the particle birth method passed back the list of possible cells, we could narrow the linear search to that set of cells.

``` fortran
! only allocate this on the first particle
if (nps == 1)
    call get_max_num_cells(max_num_cells)
    allocate(max_num_cells) cell_list
end if
call particle_birth(rands, xxx, yyy, zzz, erg, wgt, cell_list)
icl_tmp = find_cell(cell_list)
```

gonuke:
This can be done as part of `sampling_setup`

zxkjack123:
Close it as the #1118 get merged.

PR #1097, opened 2019-02-15
Update PyNE to use PyMOAB instead of PyTAPS with newest MOAB

gonuke:
This is a large PR developed collectively by @CNERG to remove the dependency on the deprecated PyTAPS project and replace it with the native python support in MOAB, known as pymoab.

This is the culmination of many months of effort, primarily by @pshriwise, the author of pymoab, and completed by @cnerg over the last month.  

All changes have been internally reviewed, discussed and revised within the @cnerg team.

Details can be found in the news file included in this PR.  Most notably,
* some syntax for mesh and tags has been changed, generally to a more streamlined format
* some dependencies have been more narrowly confined
* new dockerfiles have been added to make it easier to update the CI docker images

Full support for Python 3 should now be possible but is not yet included in this PR.

Any partial review is appreciated and the @cnerg team is, of course, happy to respond to any questions.

This will probably demand a rev of the PyNE version string and a release.

PR #1110, opened 2019-03-09
Improve source sampling

zxkjack123:
As addressed in #1109, the efficiency of photon source sampling could be improved if we pass cell list back to fortran side. The imporvement  works for both voxel/sub-voxel R2S. This PR is trying to implement this improvement. 

Things to do:
- [x]: Unify the sampler, There are two types of sampler exist and used in source_sampling: (old) sampler for only voxel case, and (new) sampler for both voxel and sub-voxel cases. Old sampler is currently used for voxel R2S, and the new sampler is used for sub-voxel mode. Change to use new sampler for both voxel/sub-voxel modes.
- [ ]: Add test files (input and output files for test problem.
- [ ]: Modify the setup and sampler functions to pass cell_list to fortran.
- [ ]: Allow user to turn off void rejection in source sampling routine, #1111 

NOTICE: This PR is based on #1097, should be rebased when #1097 get merged.

### DIFF
--- a/src/source_sampling.cpp
+++ b/src/source_sampling.cpp
@@ -21,16 +21,7 @@ void pyne::sampling_setup_(int* mode) {
           "cell_number"));
     tag_names.insert(std::pair<std::string, std::string> ("cell_fracs_tag_name",
           "cell_fracs"));
-    if (*mode == 0) {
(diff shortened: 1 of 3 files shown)


zxkjack123:
Split the content to smaller PRs. Close this one now.

PR #1112, opened 2019-03-15
DAGMC with the new pyne

shimwell:
This PR Allows DAGMC to built into pyne with MOABv5 or newer

modified file:   CMakeLists.txt now calls the finddagmc function
modified file:   src/CMakeLists.txt adds libraries for moab and dagmc if needed
modified file:   src/dagmc_bridge.cpp makes an get_dagmc_instance 
new file: FindDAGMC.cmake which is from OpenMC and finds dagmc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,6 +89,14 @@ if(${MOAB_FOUND})
   endif(APPLE)
 endif(${MOAB_FOUND})
 
+# find DAGMC                                                                                                                        
+find_package(DAGMC)
+if(DAGMC_FOUND)
+  include_directories(${DAGMC_INCLUDE_DIRS})
+  set(LIBS ${LIBS} ${DAGMC_LIBRARIES})
+  link_directories(${DAGMC_LIBRARY_DIRS})
+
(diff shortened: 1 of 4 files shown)


pshriwise:
Hi @Shimwell! Thanks for the update! 

#1097 also works with MOABv5.0 plus and contains an alternate solution. I'm honestly not sure which of the two solutions is better in the long run, but your comments on the solution in that PR or suggested changes there are very welcome!

This PR could also be merged in the short term and #1097 could be updated to support the changes here. I'll leave that decision to @gonuke and the CNERG team as they are pushing that work toward the finish line.

makeclean:
Ah I see the confusion, that commit message perhaps isn't the most accurace, my fault - this specifically fixes dagmc inside of pyne with DagMC, is that also in that other PR? If so I'm must get round to reviewing it!

pshriwise:
@makeclean yup yup finding DAGMC is also included in the other PR.

makeclean:
Well I wasted @Shimwell 's time on this one, sorry Jon! Looks like we just need to merge the other change

PR #1113, opened 2019-03-19
use new sampler for both default and subvoxel r2s

zxkjack123:
There are two types of sampler exist and used in source_sampling:
- Old: sampler for only voxel case, and 
- New: sampler for both voxel and sub-voxel cases.

Old sampler is currently used only for voxel R2S, and the new sampler is used for sub-voxel mode. As the correctness of sub-voxel mode has been proved using tests and benchmark, it's better to unify the sampler, use the new sampler for both voxel and sub-voxel mode.

This PR is part of #1110. I split it into small PRs to make review process easier.

### DIFF
--- a/src/source_sampling.cpp
+++ b/src/source_sampling.cpp
@@ -64,7 +64,6 @@ pyne::Sampler::Sampler(std::string filename,
                  bool uniform)
   : filename(filename), src_tag_name(src_tag_name), e_bounds(e_bounds) {
   bias_mode = (uniform) ? UNIFORM : ANALOG;
-  sub_mode = DEFAULT;
   setup();
 }
 
@@ -77,7 +76,6 @@ pyne::Sampler::Sampler(std::string filename,
     e_bounds(e_bounds), 
     bias_tag_name(bias_tag_name) {
   bias_mode = USER;
-  sub_mode = DEFAULT;
   setup();
 }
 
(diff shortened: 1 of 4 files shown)


src/source_sampling.cpp
```diff
@@ -211,6 +187,25 @@ void pyne::Sampler::setup() {
   }
   else throw std::invalid_argument("Mesh file must contain only tets or hexes.");
 
+  if (ve_type == moab::MBHEX){
+      // cell_number_tag
+      if (tag_names.find("cell_number_tag_name") == tag_names.end()) {
```
gonuke, line 184: Can this be done with a `try` block that throws an error when the value isn’t found in the dictionary?  Or does the new mesh infrastructure just create this if it’s not found?

tests/test_source_sampling.py
```diff
@@ -41,10 +41,10 @@ def try_rm_file(filename):
 
 
 @with_setup(None, try_rm_file('tet.h5m'))
-def test_single_tet_tag_names_map():
```
gonuke, line 44: Is this for hex or for tet?

tests/test_source_sampling.py
```diff
@@ -950,40 +1035,41 @@ def test_template_examples():
...
-                for num_bias_groups in (1, num_e_groups):
-                    bias_tag = [[1.0]*num_bias_groups, [1.0]*num_bias_groups]
-                    _source_sampling_test_template(
-                        mode, cell_fracs_list, src_tag, bias_tag)
-            else:
-                _source_sampling_test_template(mode, cell_fracs_list, src_tag)
-
-    # SUBVOXEL
-    for mode in (SUBVOXEL_ANALOG, SUBVOXEL_UNIFORM, SUBVOXEL_USER):
+#    for mode in (DEFAULT_ANALOG, DEFAULT_UNIFORM, DEFAULT_USER):
```
gonuke: Remove commented lines

gonuke (changes requested):
Thanks for streamlining this.  Is there any reason to even distinguish between SUBVOXEL and DEFAULT anymore, in the sampler?

src/source_sampling.cpp
zxkjack123: For Cartesian mesh, there is a `cell_number` and `cell_fracs` tags. However, for unstructured mesh, there isn't. Therefore, this block of code should in `if` but not `try` block.

tests/test_source_sampling.py
zxkjack123: The old name of this function indicates that it is for `tet`. However, the mesh used inside the function is `hex`. Therefore, I changed the name of this function from `tet` to `hex`.

gonuke: Maybe we should change the file name as well from `tet.h5m` to `hex.h5m`

src/source_sampling.cpp
gonuke: I think the top level if block stays, but then try within that

gonuke: Oops - this is not python... leave it how it is...

gonuke (approved):

gonuke:
Thanks @zxkjack123

PR #1114, opened 2019-03-20
R2s script add tests

zxkjack123:
This PR adds some R2S examples for voxel, sub-voxel and unstructured mesh mode. These example files include both input files and expected output files.
Test functions for r2s step1 and step2 of these three modes are added as well.


pyne/r2s.py
```diff
@@ -101,6 +102,8 @@ def irradiation_setup(flux_mesh, cell_mats, cell_fracs, alara_params, tally_num=
         The if False, geometry discretization will be done with randomly fired
         rays. If true, a grid of sqrt(num_rays) x sqrt(num_rays) rays is used
         for each mesh row.
+    cell_fracs : record array
```
gonuke: This appears above in the comments

gonuke (changes requested):
Thanks @zxkjack123 - this will be useful!

I don’t think we need any of the nuclear data files used by ALARA to be included in this PR.

zxkjack123:
@gonuke That's true. Nuclear data are for ALARA, I have removed that.

gonuke (approved):

gonuke:
Thanks @zxkjack123

PR #1117, opened 2019-03-21
Fix expand element

makeclean:
Fix a bug where if the nuc_data.h5 were not present, the expand_elements() method would introduce nuclides that did not exist, due entirely to a badly placed loop iterator. The added tests and fix ensure that this can not longer happen.

The tests have to happen in two files due to that fact that setting and unsetting the pyne.NUC_PATH variable cannot be changed inside a single script, since the loading of the data is actually done in the C++ layer.

### DIFF
--- a/src/material.cpp
+++ b/src/material.cpp
@@ -1314,6 +1314,8 @@ pyne::Material pyne::Material::expand_elements(std::set<int> exception_ids) {
       }
       while(zabund <= znuc) {
         nabund = (*abund_itr).first;
+        zabund = nucname::znum(nabund);
+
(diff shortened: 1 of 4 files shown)


gonuke:
How did this ever work before under any circumstances?

makeclean:
Most people run nuc_data_make so you would never spot it, it only happens when you have no nuc_data.h5 file present.

gonuke:
It looks like it should miss the first nuclide for every element though? Maybe I'm not looking at it carefully enough.

It also could be simpler with slightly different data structures!

makeclean:
I had a whole other one written a long time ago, but when I found this bug, I realised a 1 line move would fix it. Im all for a refactor :)

gonuke:
Thanks @makeclean

PR #1118, opened 2019-03-23
Speed up source sampling

zxkjack123:
This PR fixed the error in voxel R2S, and speed up the source sampling.
When performing source sampling, the following procedure was performed: 
1. A source particle is firstly sampled by `particle_birth`, we get the `xxx`, `yyy`, `zzz`, `wgt`, `erg` and the expected `cell_num`. 
2. Loop over the entire cells (1 to mxa, the number of cells in the problem) to find out which cell the particle is really located: `ncl(icl_tmp)`. 
3. Compare the expected `cell_num` and the really cell number `ncl(icl_tmp)`. If they are the same, we got a useful particle, otherwise, sampled the `xxx`, `yyy`, `zzz` again repeat step2 and step3.
In step2, we need to check 0.5*mxa times to find out the really cell number `ncl(icl_tmp)`, it is quite time consuming.

@gonuke suggested (in #1109) to pass back the list of cells to speed up step2, by loop over the list of cells rather than the entire cells. The size of `cell_list` is 1 in sub-voxel mode (pointed out by @pshriwise) and `max_num_cells` in voxel mode, much smaller than `mxa`. By applying this method, we can reduce the check time from mxa/2 to (max_num_cells+1)/2.

As the changes in this PR related to share/source.F90, the correctness and the performance cannot be observed from CI, results of a test could see this spreadsheet: https://docs.google.com/spreadsheets/d/1xh4nT9st8fNQzze501srMFqVuOrM8Yf7Hctkvmv6Ujk/edit?usp=sharing
From the results comparison, we can see that:
- The previous voxel R2S was wrong. Fixed in this PR. As it was wrong, it's meaningless to compare their CPU time. But at least this PR doesn't slow down the speed.
- The CPU time of photon transport of sub-voxel R2S decreased about 60% to 70%.

Major changes:
- Introduce new variable 'p_src_num_cells` to fix error in voxel R2S
- Pass `cell_list` back to Fortran to speed up source sampling
- Update test functions
- Add `cell_fracs` backwards compatibility.


gonuke:
We know which cell we want for subvoxel sampling, but not for default sampling, right?

gonuke:
Also, there is probably a way to convert to the right cell index without having to build your own data structure.  

Generally, we refer to the number assigned by Trelis (or the user in native MCNP) as the `ID` and the ordinal number that counts cells from 1 to `mxa` as the `index`.  It is generally easier to convert from `index` to `ID`, but I think MCNP provides both.

pshriwise:
Impressive speedup! Very thorough demonstration and testing too. Nice!

Maybe for both default sampling and subvoxel sampling a list can be passed back to limit the cell search, but for subvoxel this will always be a list with size == 1? It's been a while since I've looked at this code, so maybe that isn't as easy as I make it sound :)

zxkjack123:
@gonuke No, we don't know the cell we want for default sampling. In default sampling, we sample the position first and then find out which it is located.
For default sampling, the two options:
1. Pass back the `cell_list`, and loop over the `cell_list` to find out which cell the source particle really located. Expected check times: 0.5\*(max_num_cells+1).
2. Do not pass back the `cell_list`, still use `cell_num`. This `cell_num` is one of the cells in the `cell_list`.   Assuming that the cell fraction of the sampled cell_num is `a`. Then the expected check time is: a\*1 + (1-a)\*a\*2 + (1-a)\*(1-a)\*a\*3 + ... = 1/a, which is greater than 0.5\*(max_num_cells+1) on average.

For sub-voxel sampling, we know the cell we want, things go slight different
1. Pass back the `cell_list`, then the expected check time is: (max_num_cells+1)/2 \* 1/a;
2. Use `cell_num`, then the expected check time is still 1/a.

Therefore, for default sampling, it's faster to pass back the `cell_list`, for sub-voxel R2S, it's faster to use `cell_num` directly.

zxkjack123:
@pshriwise Good idea! Regarding the `cell_num` as a list of size==1, we can use the same code for both default and sub-voxel sampling.

gonuke:
(As I just mentioned in slack)
`cell_idx = namchg(1,cell_id)` will find the cell index for a given cell ID
and
`cell_id = ncl(cell_idx)` will find the cell ID for a given cell index

zxkjack123:
I made several syntax errors in passing arrays between C++ and Fortran, and in passing parameters between Fortran functions. After fixing several bugs, the sub-voxel mode with cell_list passed back works now. However, there is still at least a bug with the voxel mode, fixing it now...

zxkjack123:
Found the reason of the voxel mode error. Because I used a very old file that doesn't have `cell_number` and `cell_fracs` tag.
The code works well with voxel R2S.
Performance tests ongoing ...

zxkjack123:
## Error found in voxel R2S.

[This spreadsheet](https://docs.google.com/spreadsheets/d/1xh4nT9st8fNQzze501srMFqVuOrM8Yf7Hctkvmv6Ujk/edit?usp=sharing),  contains the two test cases:
1. Without this PR
2. With this PR
`tally_after/tally_before` shows that there is a huge difference between the results of these two cases.
The voxel R2S without this PR#1118 is wrong. The error was introduced [here](https://github.com/pyne/pyne/commit/af3a8e86ca33c8bbc53bd3769eaa944f5c6a5448#diff-bf00019a7df8611d284284dff3938adbL278), in PR #1113. 

**Note**: **This error does not influence sub-voxel mode. And this error is fixed in this PR.**

## Reason of this error:
 - The variable `max_num_cells` is mainly used for sub-voxel R2S before PR #1113. In voxel mode, `max_num_cells` is set to be 1 and will not be changed. So the size of `pdf` is `num_ves * num_e_groups * max_num_cells`.
- Before #1113, the length of `pdf` for voxel R2S is correct. After #1113, the `max_num_cells` is changed by [this line](https://github.com/pyne/pyne/commit/af3a8e86ca33c8bbc53bd3769eaa944f5c6a5448#diff-bf00019a7df8611d284284dff3938adbR279):
```
max_num_cells = num_groups(cell_fracs_tag);
```
- Therefore, the size of `pdf` for voxel R2S is `num_ves * num_e_groups * max_num_cells`, but not `num_ves * num_e_groups * 1`. 
- Then, the elements of `pdf` will multiply by the cell fractions, which is also changed according to the real `cell_fracs`. The consequence is that the `pdf` is distorted. Leading to wrong `pdf` -> wrong source distribution -> wrong dose rate.

## Why the wrong code in #1113 passed the test?
Because in all of the test cases designed for voxel R2S, their cell_fracs is [1.0], or [0.5, 0.5]. The cell fractions for different cells are the same, the `pdf` is not distorted by coincidence.

## Fix
- A new variable `p_src_num_cells` is used to indicate how many cells in each photon source tag (1 for voxel mode, `max_num_cells` for sub-voxel mode).
- Do not change the `cell_fracs` when it's voxel mode to make sure the `pdf` is correct.

pshriwise:
Glad you caught this @zxkjack123! Is this PR ready for another round of review then?

zxkjack123:
For this [spreadsheet](https://docs.google.com/spreadsheets/d/1xh4nT9st8fNQzze501srMFqVuOrM8Yf7Hctkvmv6Ujk/edit?usp=sharing), we can see that by applying this PR, we don't get a speed up, what's worse, it slows down. CPU time increased to about 150% to 300% of the origin code.

Analysis of the reason. This massive slows down may be caused by extra loops/iterations. Possible reason:
- namchg: Every time we call `namchg`, it iterate from 1 to mxa to find the cell idx. Expected iteration time for each valid source particle: max/2 \* (1+p_src_num_cells)/2 \* 1/a = (1+p_src_num_cells)\*max/(4a). Tested with two cases: case 1, sub-voxel mode with namchg, speed up about 58% ~ 70%. Case 2, sub-voxel mode with cell_idx, speed up about 63% ~ 70%. Shows that `namchg` doesn't contribute a lot to the cpu time.
- tiny cell (with cell fractions a < 0.01) not in cell_list but sampled: equivalent to sample source particle number a\*idum(2), which is about 1. Therefore, on average, the tiny cell is not a big deal.

zxkjack123:
@pshriwise Thanks. But I think it's better to review when I have finish the new test for sub-voxel mode. Also, I need to test the influence of the `namchg`/`cell_idx` to see which is faster. I will mention you when I finish the new tests.

pshriwise:
Sounds good! Let me know :)

zxkjack123:
Update:
Finished the test. 

Major changes in this PR:
- Pass `cell_list` back to fortran, and use this `cell_list` to narrow down the iteration arrangement in `find_cell`
- Use `namchg` to find out the cell index according to the cell number

Short conclusion of the speed up effect of this PR:
- For voxel R2S, this PR will **slow down** the photon transport to the extent of about 150% ~ 300%.
- For sub-voxel R2S, this PR **speed up** the photon transport by about 58% ~ 70%.

Currently, I don't know the reason why it slows down when it's voxel mode. But it's ready for another round of review @gonuke @pshriwise.

pshriwise:
To make sure I'm reading this right, the ` voxel` mode got about 1.5-3 times slower than the sampler before this PR and the `subvoxel` mode got 2-3 times faster?

Will review soon. Thanks for all of the hard work @zxkjack123!

pyne/source_sampling.pyx
```diff
@@ -482,6 +482,7 @@ cdef class Sampler:
         cdef int irands
         cdef int rands_size
         cdef double * rands_data
+#        cdef cpp_vector[int] cell_list_proxy
```
pshriwise: This can go, presumably?

pyne/source_sampling.pyx
```diff
@@ -499,8 +500,14 @@ cdef class Sampler:
         cdef cpp_source_sampling.SourceParticle c_src = \
                 (<cpp_source_sampling.Sampler *> self._inst)\
                 .particle_birth(rands_proxy)
+#        cell_list = c_src.get_cell_list()
```
pshriwise: These can be removed as well.

share/source.F90
```diff
@@ -12,35 +12,41 @@
 ! Full instructions on compiling and using MCNP5 with this subroutine are found
 ! in the PyNE user manual.
 
-function find_cell() result(icl_tmp)
-! This function to determines the current MCNP cell index location and exits if
-! no valid cell is found. This only works if there are no repeated geometries or
-! universes present in the model.
+function find_cell(cell_list, cell_list_size) result(icl_tmp)
```
pshriwise, line 15: There should probably be some documentation about the input parameters to this function.

share/source.F90
```diff
@@ -12,35 +12,41 @@
...
+! This only works if there are no repeated geometries or universes present in
+! the model.
 
     use mcnp_global
     use mcnp_debug
     ! xxx,yyy,zzz are global variables
     ! mxa is global
     integer :: i ! iterator variable
-    integer :: j ! tempory cell test
+    integer :: j ! temporary cell test
```
pshriwise, line 51: good fix :)

zxkjack123:
> To make sure I'm reading this right, the ` voxel` mode got about 1.5-3 times slower than the sampler before this PR and the `subvoxel` mode got 2-3 times faster?
> 
> Will review soon. Thanks for all of the hard work @zxkjack123!

Yes, that's right.

share/source.F90
```diff
@@ -12,35 +12,41 @@
...
-         icl_tmp = i
-         exit
-      endif
-    enddo
-    ! icl is now set
-
-    if(icl_tmp .le. 0) then
-      write(*,*) 'Nonsense cell number stopping'
-      stop
+    if (cell_list_size .eq. 1) then
```
pshriwise: Do we need a special case for `cell_list_size .eq. 1`? Can't the loop below be used regardless of size?

Maybe there's a lot of overhead to that which I don't know about, but it seems like it would make the code a little cleaner.

share/source.F90
```diff
@@ -12,35 +12,41 @@
...
+       call chkcel(cid, 0, j)
+       if (j .eq. 0) then
+           icl_tmp = cid
+       endif
+    else
+       do i = 1, cell_list_size
+          cid = namchg(1, cell_list(i))
+          call chkcel(cid, 0, j)
+          if (j .eq. 0) then
+             ! valid cel set
```
pshriwise: cel -> cell

share/source.F90
```diff
@@ -72,24 +81,36 @@ subroutine source
...
+       icl_tmp = find_cell(cell_list, cell_list_size)
    endif
-   if (mat(icl_tmp).eq.0 .and. tries < idum(2)) then
-       tries = tries + 1
-       goto 200
-   end if
 
-   if(tries.eq.idum(2)) then
+   ! check whether this is a valid cell
+   if (icl_tmp .le. 0) then
```
pshriwise, line 143: Before these were `icl_tmp .eq. 0` and now they are `icl_tmp .le. 0`. Why the change?

share/source.F90
```diff
@@ -72,24 +81,36 @@ subroutine source
...
-   call particle_birth(rands, xxx, yyy, zzz, erg, wgt, cell_num)
-   icl_tmp = find_cell()
-   if (idum(1) > 2) then
-       if (cell_num .ne. ncl(icl_tmp) .and. tries < idum(2)) then
-           tries = tries + 1
-           goto 200
-       endif
+   call particle_birth(rands, xxx, yyy, zzz, erg, wgt, cell_list)
+   ! In tet mesh, cell_list_size = 0, loop over entire cells to find icl_tmp
+   if (cell_list_size == 0) then
```
pshriwise: I'm feeling like this check should go in `find_cell`. You've now modified `find_cell` to take a short list of cells to check for - a guess if you will. If that list isn't provided (or is length zero), then `find_cell` should perhaps default to searching over all cells. This would make it a more flexible function. To do this, the `cell_list` can be an optional argument (if the Fortran standard in use allows for it) and `cell_list_size` can be set to 0 to indicate a global search over all cells should be performed.

src/source_sampling.cpp
```diff
@@ -31,15 +32,19 @@ void pyne::particle_birth_(double* rands,
...
     pyne::SourceParticle src = sampler->particle_birth(rands2);
     *x = src.get_x();
     *y = src.get_y();
     *z = src.get_z();
     *e = src.get_e();
     *w = src.get_w();
-    *c = src.get_c();
+    std::vector<int> c_list = src.get_cell_list();
+    int cell_list_size = sampler->get_cell_list_size();
+    for (int i=0; i<cell_list_size; i++){
```
pshriwise: Can use `std::copy` here

pshriwise:
```cpp
std::copy(c_list.begin(), c_list.end(), cell_list);
```

src/source_sampling.cpp
```diff
@@ -85,7 +90,8 @@ pyne::Sampler::Sampler(std::string filename,
                  int mode)
   : filename(filename),
     tag_names(tag_names),
-    e_bounds(e_bounds) {
+    e_bounds(e_bounds),
+    mode(mode) {
```
pshriwise, line 94: Is this a fix?

src/source_sampling.cpp
```diff
@@ -139,14 +146,22 @@ pyne::SourceParticle pyne::Sampler::particle_birth(std::vector<double> rands) {
   xyz_rands.push_back(rands[3]);
   xyz_rands.push_back(rands[4]);
   moab::CartVect pos = sample_xyz(ve_idx, xyz_rands);
-  // cell_number
-  if (ve_type == moab::MBHEX) {
-  cell_id = cell_number[ve_idx*max_num_cells + c_idx];
+  if (ve_type == moab::MBTET) {
+     cell_list.resize(0);
```
pshriwise: No need to resize here, the cell list should be size zero as it was just created.

src/source_sampling.cpp
```diff
@@ -139,14 +146,22 @@ pyne::SourceParticle pyne::Sampler::particle_birth(std::vector<double> rands) {
...
   } else {
-     cell_id = -1;
+     if (mode == 3 or mode == 4 or mode == 5) { // sub-voxel
+        cell_list.resize(1);
```
pshriwise: Can use

```cpp
  cell_list.push_back(cell_number[ve_idx*max_num_cells + c_idx]);
```

and since we're C++11 now IIRC, this would be even better

```cpp
  cell_list.emplace_back(cell_number[ve_idx*max_num_cells + c_idx]);
```

src/source_sampling.cpp
```diff
@@ -139,14 +146,22 @@ pyne::SourceParticle pyne::Sampler::particle_birth(std::vector<double> rands) {
...
+        cell_list[0] = cell_number[ve_idx*max_num_cells + c_idx];
+     }
+     else { // mode in 0, 1, 2, voxel
+        cell_list.resize(max_num_cells);
```
pshriwise: Same comment as above. No need for resize here, just push/emplace onto the vector.

src/source_sampling.cpp
```diff
@@ -139,14 +146,22 @@ pyne::SourceParticle pyne::Sampler::particle_birth(std::vector<double> rands) {
...
+        for (int c=0; c<max_num_cells; c++) {
+           cell_list[c] = cell_number[ve_idx*max_num_cells + c];
+        }
+     }
   }
   pyne::SourceParticle src = SourceParticle(pos[0], pos[1], pos[2],
-      sample_e(e_idx, rands[5]), sample_w(pdf_idx), cell_id);
+      sample_e(e_idx, rands[5]), sample_w(pdf_idx), cell_list);
```
pshriwise, line 159: Nice to see this change :)

src/source_sampling.cpp
```diff
@@ -254,11 +269,12 @@ void pyne::Sampler::mesh_tag_data(moab::Range ves,
...
-  // r2s can use the same form of pdf size description
+  // Set the default value of max_num_cells to 1, so that the structured mesh
+  // and unstructured mesh r2s can use the same form of pdf size description.
   max_num_cells = 1;
-  // set the default value of cell_fracs to 1.0
-  cell_fracs.resize(num_ves*max_num_cells);
+  p_src_num_cells = 1;
+  // set the default value of cell_fracs to 1.0 for unstructured mesh
+  cell_fracs.resize(num_ves);
   for(int i=0; i<cell_fracs.size(); i++){
```
pshriwise: This was already here... but this resize/loop can be done in one line.

```cpp
cell_fracs.resize(num_ves, 1.0);
```

You may want to read the documentation on this, though.  This will put 1's at the end of a vector if it is making it larger, but when resizing to a smaller vector it will not change any existing values. For an unstructured mesh run, these are presumably always 1's, so that should be ok.

Alternatively, you can use this:

```cpp
cell_fracs  = std::vector<double>(num_ves, 1.0);
```

every time, but I don't think it will be as efficient.

src/source_sampling.cpp
```diff
@@ -269,24 +285,27 @@ void pyne::Sampler::mesh_tag_data(moab::Range ves,
       rval = mesh->tag_get_handle(cell_fracs_tag_name.c_str(),
                                   cell_fracs_tag);
       max_num_cells = num_groups(cell_fracs_tag);
-      num_e_groups /= max_num_cells;
-      cell_fracs.resize(num_ves*max_num_cells);
-      rval = mesh->tag_get_data(cell_fracs_tag, ves, &cell_fracs[0]);
+      if (mode == 3 or mode == 4 or mode == 5) {
```
pshriwise: I'll need to think/read about this - While using `or` is more human readable I think that it's unfortunately less stable (some environments may declare `or`/`and` as macros etc.). I'm not sure where people have landed on this but I'll try to look into it and get back to you.

src/source_sampling.cpp
```diff
@@ -494,6 +513,21 @@ int pyne::Sampler::num_groups(moab::Tag tag) {
...
 
+int pyne::Sampler::get_cell_list_size() {
+   // cell_list_size should be:
+   // 0: for unstructured mesh
+   // 1: for sub-voxel R2S
+   // max_num_cells: for voxel R2S
+   if (ve_type == moab::MBTET) {
+      return 0;
+   } else {
+      if (mode == 0 or mode == 1 or mode == 2) {
```
pshriwise: `mode < 3` is equivalent here.

pshriwise: Probably true for my comment above too

src/source_sampling.cpp
```diff
@@ -554,17 +588,16 @@ pyne::SourceParticle::SourceParticle() {
     z = -1.0;
     e = -1.0;
     w = -1.0;
-    c = -1;
 }
 
 pyne::SourceParticle::SourceParticle(double _x, double _y, double _z,
-                                     double _e, double _w, int _c) {
+                                     double _e, double _w, std::vector<int> _cell_list) {
```
pshriwise: Initializers are typically preferred I think
```suggestion
pyne::SourceParticle::SourceParticle(double _x, double _y, double _z,
                                     double _e, double _w, int _c) {
                                     double _e, double _w, std::vector<int> _cell_list) : 
                                     x(_x), y(_y), z(_z), 
                                     e(_e), w(_w), c(_c), 
                                     cell_list(_cell_list) { }
```

tests/test_source_sampling.py
```diff
@@ -1673,21 +1714,27 @@ def _source_sampling_test_template(mode, cell_fracs_list, src_tag,
...
+        rands = np.array([uniform(0, 1) for x in range(6)])
+        s = sampler.particle_birth(rands)
         # check w, and c for each particle
         # calculate the expected weight and cell_number
         exp_w, exp_c = _cal_exp_w_c(s, mode, cell_fracs, src_tag, bias_tag)
         assert_equal(s.w, exp_w)
         # when mode in (0, 1, 2), the set exp_c is (-1), otherwise it contains
         # several available cell number
-        assert(s.c in exp_c)
+        if mode in (0, 1, 2):
```
pshriwise, line 1734: I feel like we run into this mode check a lot, where we're really just trying to figure out if we're in sub-voxel or voxel mode. Didn't we discuss having two "mode" enumerators at some point? One for the sampling type (analog, uniform, user-specified) and another for the sampling regions (voxel, sub-voxel). My terminology may not be ideal here..

src/source_sampling.h
```diff
@@ -198,7 +206,14 @@ namespace pyne {
...
     int num_bias_groups; ///< Number of groups tag \a _bias_tag_name
     int max_num_cells; /// Max number of cells in voxels
+    // Equivalent cell number in photon source.
+    // For voxel R2S, p_src_num_cells = 1
+    // For sub-voxel R2S, p_src_num_cells = max_num_cells
+    // For unstructured R2S, p_src_num_cells = 1.
+    int p_src_num_cells;
+    int cell_list_size;
     BiasMode bias_mode; ///< Bias mode: ANALOG, UNIFORM, USER
+    int mode; ///< Sampler mode, currently support 0, 1, 2, 3, 4, 5
```
pshriwise, line 220: Here it is (see my comment below on bias vs. sampling mode). This mode should be limited to only specifying `VOXEL` or `SUBVOXEL` the `bias_mode` attribute can be used to look up the bias when needed. This will simplify a lot of checks in the code I think.

pshriwise (changes requested):
Looks really good @zxkjack123! One thing I'm a little confused on is where the `cell_list` is being limited when doing voxel sampling. I see that in `particle_birth_` a cell list is constructed. If we're in voxel mode, however, the cell list size from `get_cell_list_size` is always `max_num_cells`, right? 

Maybe this is why the code is taking longer in voxel mode? Let me know if I'm missing the place where this list is limited.

zxkjack123:
> Looks really good @zxkjack123! One thing I'm a little confused on is where the `cell_list` is being limited when doing voxel sampling. I see that in `particle_birth_` a cell list is constructed. If we're in voxel mode, however, the cell list size from `get_cell_list_size` is always `max_num_cells`, right?
> 
> Maybe this is why the code is taking longer in voxel mode? Let me know if I'm missing the place where this list is limited.

Yes, @pshriwise. The size of `cell_list`, or the value of `cell_list_size` is:
- 0: for unstructured mesh
- 1: for sub-voxel mode
- `max_num-cells`: for voxel mode

pshriwise:
Oh I thought there was some way to limit that list, even in voxel mode. My mistake. In that case, can we not hold a static list of cells to return. Right now I think that list is being created every time we sample.

tests/test_source_sampling.py
zxkjack123: Yes, we have discussed that `mode` before. And I added that at a PR, however, I removed that variable at a recent PR because I was trying to unify the voxel/sub-voxel source sampling.
It turns out that I need to add that enumerator back.

zxkjack123:
Updates according to the suggestion of @pshriwise:
- Add a variable `mesh_mode` to simplify the judgement statements
- Use initializers in a constructor of `SourceParticle`
- Some cleanup

share/source.F90
zxkjack123: When we use `cell_list` to find out the cell, there is a very small possibility to find that the `xxx, yyy, zzz` located in a place not listed in the `cell_list` (because ray tracing may not find that small cell in the voxel). In that case, the `icl_tmp` is set to `-1`. And then sample the particle again.

src/source_sampling.cpp
zxkjack123: No, actually, the variable `mode` was not in the attribute of the  `sampler`. I need to use it as a global variable of sampler, so add it here.

share/source.F90
pshriwise: Sounds good. Thank you for clarifying!

src/source_sampling.cpp
```diff
@@ -139,14 +144,18 @@ pyne::SourceParticle pyne::Sampler::particle_birth(std::vector<double> rands) {
...
-  // cell_number
   if (ve_type == moab::MBHEX) {
-  cell_id = cell_number[ve_idx*max_num_cells + c_idx];
-  } else {
-     cell_id = -1;
+     if (mesh_mode == SUBVOXEL) {
+         cell_list.emplace_back(cell_number[ve_idx*max_num_cells + c_idx]);
+     }
+     else { // Voxel
+        for (int c=0; c<max_num_cells; c++) {
```
pshriwise, line 153: For voxel mode it would probably be best to maintain a static cell_list rather than create this vector each time we sample a particle.

src/source_sampling.cpp
```diff
@@ -85,7 +88,8 @@ pyne::Sampler::Sampler(std::string filename,
                  int mode)
   : filename(filename),
     tag_names(tag_names),
-    e_bounds(e_bounds) {
+    e_bounds(e_bounds),
+    mode(mode) {
   // determine the bias_mode
   if (mode == 0){
```
pshriwise, line 96: On the C/C++ side it would probably be best to continue using the `MeshMode` enumerators where possible.

src/source_sampling.cpp
```diff
@@ -179,6 +188,17 @@ void pyne::Sampler::setup() {
   }
   else throw std::invalid_argument("Mesh file must contain only tets or hexes.");
 
+  // Assign MeshMode: VOXEL, SUBVOXEL, TET
+  if (ve_type == moab::MBHEX){
+     if (mode < 3){
```
pshriwise: Make `mode` a `MeshMode` enumerator value and you won't have to this extra check.

pshriwise (changes requested):
Looking pretty good! I think storing a static cell_list when in sampling with `VOXEL` could help with the performance issue you're seeing there a lot. Have you given that a try?

It's getting close I think!

src/source_sampling.cpp
zxkjack123: This mode is passed from Fortran. DAGMC input -> source.F90 -> source_sampling.cpp. The mode could be (0, 1, 2, 4, 5, 6). We have to keep it as an input parameter.
We can't assign `mesh_mode` from this mode, because the `TET` mesh and `VOXEL` share the modes (0, 1, 2). We need extra checks when we get both the `ve_type` and `mode`.

zxkjack123:
> Looking pretty good! I think storing a static cell_list when in sampling with `VOXEL` could help with the performance issue you're seeing there a lot. Have you given that a try?
> 
> It's getting close I think!

I have set the `cell_list` as a variable of `Sampler` in the latest commit according to the suggestion of @pshriwise.

pshriwise:
The latest commit is a step toward what I was thinking, but now I'm not sure I was thinking the right thing! Let's chat about about this in person sometime.

share/source.F90
```diff
@@ -12,35 +12,58 @@
...
+        do i = 1, mxa
+           call chkcel(i, 0, j)
+           if (j .eq. 0) then
+              ! valid cell found
+              icl_tmp = i
+              exit
+           endif
+        enddo
+    else
+        ! HEX mesh. VOXLE/SUBVOXEL
```
gonuke:
```suggestion
        ! HEX mesh. VOXEL/SUBVOXEL
```

pshriwise:
Any updates on why voxel mode is slower when applying this PR? Should we dig into this more?

zxkjack123:
> Any updates on why voxel mode is slower when applying this PR? Should we dig into this more?

I tested the case of returning a 0 sized `cell_list` to `source.F90`, it turns out that the CPU time if almost the same as the case of returning a `max_num_cells` sized.

zxkjack123:
And there is another thing I'd like to say, in voxel mode, if we find cell in the entire cells (mxa), the SDR is much higher than the case of finding cell in `cell_list`. For example, the mesh case mesh_8_6_8_def, first  cooling time, the results are:
- find cell by iterate the entire cells (mxa): tally is: 7.49e-11
- find cell by iterate in the `cell_list` (max_num_cells): tally is: 5.09e-11, 3.89e-11.

Tests:
- Find cell using two ways for the same xxx, yyy, zzz. Then compare the icl found. Tested with FNG mesh_8_6_8_def, nps 1e7. The icls found by two options are the same. No error found.

share/source.F90
```diff
@@ -12,35 +12,58 @@
...
+           if (j .eq. 0) then
+              ! valid cell found
+              icl_tmp = i
+              exit
+           endif
+        enddo
+    else
+        ! HEX mesh. VOXEL/SUBVOXEL
+        do i = 1, cell_list_size
+           if (cell_list(i) .le. 0) then
```
pshriwise, line 63: If we know the number of valid entries when we pass the `cell_list` to `find_cell` we should just pass that value as `cell_list_size`. This check can then be removed.

share/source.F90
```diff
@@ -12,35 +12,58 @@
...
+               ! not a valid cell number (-1)
+               exit
+           endif
+           cid = namchg(1, cell_list(i))
+           if (cid .eq. 0) then
```
pshriwise: Should this case ever be hit? Should we be halting the program here?

pshriwise (commented):
Review specific to `find_cell` - I didn't see anything

I can see a few places to check for inconsistent behavior against `develop`:

  - Make sure we aren't ever exiting `find_cell` because we can't look up a cell using `namchg`. I don't think that should happen, correct me if I'm wrong.
  - Double check that the `cell_list_size` going into find_cell is correct - no "off-by-one" error there.
  - After checking those things I would go into checking that the returned cell ids are consistent with `develop`.

share/source.F90
zxkjack123: This line is hit by a tiny chance in FNG model, it is firstly hit at about nps between 1e6~1e7.
This line is not hit if I use a very simple geometry (test case, rectangular cells, less than 10 cells).

I didn't halt the program at this line before, but I think it worth to do it.

share/source.F90
zxkjack123: We can get the valid cell list size by two options:
- A: pass back another variable
- B: find the first negative value in the cell list

But I both these two options just remove the check here, but add checks at other places. The differences are the speed of the different options. I think we should consider that after we checked the correctness of the PR.

pshriwise: I'm confused... can you clarify?

What do you mean by pass back another variable? 

What checks would be added in other places?

share/source.F90
pshriwise: My sense is that this should never happen. And if we do hit that case, there is something systematic about what's going wrong. We should always be able to look up the index of a valid cell in the problem right?

share/source.F90
zxkjack123: For the pass back another variable part. We have to pass back the `cell_list_size` because we need to allocate the `cell_list`. And now we want to know the number of valid cells, we can pass pack another variable named such as `valid_num_cells` or `valid_cell_list_size`.

zxkjack123: For checks at other places. If we don't pass back an extra variable of the valid cell list size, then we have to calculate it with the given `cell_list`. This calculation will performed with code blocks iteration and check of -1, or something like find index of first -1.

share/source.F90
zxkjack123: Yeah, I think I can repeat that problem with more tests, and then find out the reason.

share/source.F90
pshriwise: Ah yeah I see what you mean. We don't know how many valid cells there are when we're building the `cell_list` because the tag is fixed length and padded with -1 entries... and yeah that ties into a check somewhere else for that first -1 entry. Darn.

zxkjack123:
Update:
As the test with 1e7 nps using the following code block:
```
call particle_birth(rands, xxx, yyy, zzz, erg, wgt, cell_list)
   ! Loop over cell_list to find icl_tmp
   icl_tmp1 = find_cell(cell_list, cell_list_size)
   icl_tmp2 = find_cell(cell_list, 0)
   ! compare icl_tmp of two options
   if (icl_tmp1 > 0 .and. icl_tmp1 /= icl_tmp2) then
      write(*, *) "source particle: xxx, yyy, zzz = ",xxx,yyy,zzz
      write(*, *) "icl_tmp1 =",icl_tmp1, " icl_tmp2 =",icl_tmp2
   endif
   icl_tmp = icl_tmp2
```
the results show that they find the same icl_tmp. Therefore, the reason of got different results in differennt runs would probably a mistake somewhere I didn't notice.

zxkjack123:
@gonuke @pshriwise 
Update:
The CPU times (in secs) of several functions are recorded with test case using `CPU_TIME()` with FNG model mesh case mesh_8_6_8_def, nps:1e7.

Test1: cpu time of particle birth with/without cell_list calculation:
- particle_birth with calculating cell_list: 1.46e1
- particle_birth without calculating cell_list: 1.83e1

Test2: cpu time after particle birth (mainly find_cell):
- search in cell_list: 1.41e2
- search in mxa cells: 3.28e2

Test3: cpu time for generating 1e7 valid source particles:
- search in cell_list: 1.67e2
- search mxa cells: 3.51e2

Conclusion:
- Pass back cell_list and search in cell_list is faster than search in mxa cells
- The `find_cell` function takes most proportion in the process of source sampling.

PS: The contribution of source sampling of voxel R2S to total CPU is about 12.5% of total CPU time. The contribution of `particle_birth` is less than 1.5% (when this PR applied). It will not make any significant improvement using bounding box in `praticle_birth`.

zxkjack123:
Reason why I though this PR slows down the voxel source sampling:
Here are the data of previous tests:
https://docs.google.com/spreadsheets/d/1xh4nT9st8fNQzze501srMFqVuOrM8Yf7Hctkvmv6Ujk/edit#gid=1266272788
Parameters of the two cases in the table:
- before: FNG, mesh_8_6_8_def, with bugs/errors in sampler setup, wrong results, no cell_list passed to fortran
- after: FNG, mesh_8_6_8_def, bug fixed, correct results, with cell_list passed to fortran

The actual CPU time (see the column: `cpu_before` and `cpu_after` and `cpu_after/cpu_before`), here wasn't really a big difference. Passing back the `cell_list` do not slow down the source sampling.

However, I used scaled_cpu to do the comparison, and the `err_before` was coincidentally smaller than `err_after`, making the `scaled_cpu_before` smaller than `scaled_cpu_after`. But the `tally_before` and `err_before` are wrong results, it's meaningless to perform the scale. Therefore, passing `cell_list` to fortran code does not slow down the source sampling.

share/source.F90
zxkjack123: I repeated this problem and found that when the code hit this line, the cell number is 138. I checked the output file of both neutron transport and photon transport, found that this cell 138 exist in neutron transport but not exist in photon transport.
The material of this cell 138 is air.

It reminds me that the model for neutron transport and photon transport are slightly different:
- The lateral port plug exists in the model for neutron transport, but not in the model for photon transport
- Two detectors exist in the model for photon transport, but not neutron transport
I will double check the location of this cell 138 tomorrow.

Therefore, some cells (eg. cell A) exist in the neutron transport, and the discretize_geom will get the infomation of cell A and add it in the `cell_number` tag. However, cell A is removed in the photon transport, then the namchg could not find cell A and returns 0. That's why this line is hitted. And this line should be kept.

zxkjack123: The vol 138 is a thin layer of air around detectors. It is removed in photon transport, it won't influence the results.

zxkjack123:
@gonuke @pshriwise Any further comments and suggestions about this PR?
As this PR has been on the deck of software design and review meeting several times, to reduce confusing of the changes, here is a short summary of this PR:
- A bug in voxel R2S was fixed.
- This PR speeds up the sub-voxel R2S.
- And it does not slow down the voxel R2S.

zxkjack123:
Update: Add more description about cell not found problems, classify them into three types and treat according to their properties. Differentiate the allowed cell not found (type 1), cell not found caused by the inaccuracy of getting `cell_fracs` (type2) and possible bug (type 3).

share/source.F90
```diff
@@ -12,36 +12,98 @@
...
-      call chkcel(i, 0, j)
-      if (j .eq. 0) then
-         ! valid cel set
-         icl_tmp = i
-         ! icl now set to be valid cell
-         exit
+    ! If the icl_tmp is not find yet (for HEX mesh), type 2 or type 3 happends,
+    ! or the cell_list is not given (for TET mesh),
+    ! find it in the entire list of cells
+    if (icl_tmp == -1) .or. (cell_list_size .eq. 0) then
```
ljacobson64: This line gives me compiler errors. Adding parentheses around the entire if statement; i.e.
```
if ((icl_tmp .eq. -1) .or. (cell_list_size .eq. 0)) then
```
fixes the errors.

zxkjack123: My bad, I should have tested the compilation before commit it. This is fixed in 56d5ba4

zxkjack123:
Update: Tested with FNG model, no runtime error.

ljacobson64:
The compiler error is fixed, but I'm still getting the runtime error with the JET model. The error occurs in the `pyne::Sampler::mesh_tag_data()` function, specifically in the line

```max_num_cells = num_groups(cell_fracs_tag);```.

The 2 previous lines both set `rval` to `moab::MB_TAG_NOT_FOUND`. I'm not sure why these tags wouldn't be found, but this model and source file did work before the source_sampling updates, so I'm erring on the side of assuming it's not a problem with the source file.

**Comment by zxkjack123** [comment]
@ljacobson64 When was that source file (`source_x.h5m`) is generated? And what mode (voxel or sub-voxel) is it?
R2S has updated several times. At the very beginning, there isn't a `cell_fracs` tag for voxel R2S. Later on, the `cell_fracs` is also add to `source_x.h5m` but it is not used for voxel R2S. But the newest code, for exaple in this PR, there is `cell_fracs` tag and it is used to too.

Therefore, if the source files are generated several monthes ago, there is not a `cell_fracs` tag. Then the error you described will happen.

Could you try `h5dump source.h5m | grep "cell_fracs"` to check whether there is a `cell_fracs` tag?

**Comment by ljacobson64** [comment]
That command didn't return anything, so I guess that means it doesn't have the `cell_fracs` tag. I don't know when it was made, but I would assume it was made more than a few months ago. (Tim sent it to me; I don't know where he got it from.)

**Comment by zxkjack123** [comment]
@ljacobson64 That make sense. The reason of the error is the lack of `cell_fracs` tag in the source file. Run the R2S step1 and step2 again would fix the problem.

**Comment by ljacobson64** [comment]
So how do we fix it so Tim can run his problem again?

**Comment by zxkjack123** [comment]
@ljacobson64 Update the PyNE, and run step1 and step2 using the new `config.ini` and `alara_params.txt`. The ALARA run could be skipped.

**Comment by gonuke** [comment]
Probably better to handle the errors for a lack of `cell_fracs`.  This is a backwards compatibility issue.  We should gracefully ignore the missing `cell_fracs` since others may have old models that suddenly break

**Comment by zxkjack123** [comment]
@gonuke Agreed. Working on it.

**Comment by zxkjack123** [comment]
Update: Add `cell_fracs` tag backwards compatibility to support older version source file.

**Comment by ljacobson64** [comment]
Success! The JET model worked! Thanks @zxkjack123 !

File: share/source.F90 [hunk 28]
```diff
@@ -12,36 +12,98 @@
...
+! Parameters:
+!     cell_list: array of integers. Contains cell numbers that the source
+!                particle possiblely located.
+!     cell_list_size: integer. Size of the cell_list.
+!
+! There are 3 types of not found icl_tmp (icl_tmp == -1):
+! - Type 1: Sorce particle is located in a cell that exist in neutron transport
+!             but removed in photon transport. Therefore, the cell number does
+!             not exist in the ncl list.
+!           This is not an error, happens with samll frequency.
```
pshriwise: samll -> small

share/source.F90
```diff
@@ -12,36 +12,98 @@
...
+!           Skip it and resample next particle without warning message.
+! - Type 2: Source particle is not located in the given cell_list for HEX
+!             mesh cases (both voxel and sub-voxel). Because cell_list from
+!             discretize_geom() missed some cells with very small volume
+!             fractions.
+!           This is caused by random error, happens with samll frequency.
```
pshriwise: samll -> small

share/source.F90
```diff
@@ -12,36 +12,98 @@
...
+!           Skip it andd resample next particle with warning message.
+! - Type 3: Source partilce with a specific coordinates can't be find in any
```
pshriwise: find -> found

share/source.F90
```diff
@@ -12,36 +12,98 @@
...
+!             cell from ncl(1) to ncl(mxa).
+!           This is an error. When this error happens, it means that there is
+!             something wrong in this file (source.F90) or DAGMC geometry
```
pshriwise: "... there is something wrong in either the source.F90 file or the DAGMC geometry representation."

share/source.F90
```diff
@@ -12,36 +12,98 @@
...
+    endif
 
-    do i = 1, mxa
-      call chkcel(i, 0, j)
-      if (j .eq. 0) then
-         ! valid cel set
-         icl_tmp = i
-         ! icl now set to be valid cell
-         exit
+    ! If the icl_tmp is not find yet (for HEX mesh), type 2 or type 3 happends,
```
pshriwise: find -> found

happends -> happens

src/source_sampling.cpp
```diff
@@ -119,18 +123,17 @@ pyne::Sampler::Sampler(std::string filename,
...
   }
-
   setup();
 }
 
 pyne::SourceParticle pyne::Sampler::particle_birth(std::vector<double> rands) {
   // select mesh volume and energy group
-  // In DEFAULT mode, max_num_cells = 1
+  // For Unstructured mesh, p_src_num_cells and max_num_cells are set to 1
+  // For Cartisian mesh, max_num_cells is obtained by read cell_fracs tag
```
pshriwise: ".. is obtained via the cell_fracs tag"

pshriwise (commented):
Just some comments on the comments. 

Thanks for documenting the different cases of a failed `find_cell` so well @zxkjack123. Once a few minor typos are corrected I'm good with this PR.

zxkjack123:
@pshriwise Thanks for the review. These typos remind me of adding some kinds of grammar-checking to vim.

ljacobson64 (approved):

pshriwise (approved):
This looks good to me! Will merge at the end of the day if there are no other comments.

Thank you for all the hard work on this @zxkjack123!

share/source.F90
```diff
@@ -12,36 +12,96 @@
 ! Full instructions on compiling and using MCNP5 with this subroutine are found
 ! in the PyNE user manual.
 
-function find_cell() result(icl_tmp)
-! This function to determines the current MCNP cell index location and exits if
-! no valid cell is found. This only works if there are no repeated geometries or
-! universes present in the model.
+function find_cell(cell_list, cell_list_size) result(icl_tmp)
+! This function to determines the current MCNP cell index location.
```
gonuke:
```suggestion
! This function determines the current MCNP cell index location.
```

src/source_sampling.cpp
```diff
@@ -179,6 +187,17 @@ void pyne::Sampler::setup() {
   }
   else throw std::invalid_argument("Mesh file must contain only tets or hexes.");
 
+  // Assign MeshMode: VOXEL, SUBVOXEL, TET
+  if (ve_type == moab::MBHEX){
+     if (mode < 3){
```
gonuke: Can we define a constant here instead of '3'?

src/source_sampling.cpp
```diff
@@ -254,39 +273,51 @@ void pyne::Sampler::mesh_tag_data(moab::Range ves,
...
                                   cell_number_tag);
       rval = mesh->tag_get_handle(cell_fracs_tag_name.c_str(),
                                   cell_fracs_tag);
-      max_num_cells = num_groups(cell_fracs_tag);
-      num_e_groups /= max_num_cells;
-      cell_fracs.resize(num_ves*max_num_cells);
-      rval = mesh->tag_get_data(cell_fracs_tag, ves, &cell_fracs[0]);
-      cell_number.resize(num_ves*max_num_cells);
-      rval = mesh->tag_get_data(cell_number_tag, ves, &cell_number[0]);
+      // has_cell_fracs is assinged in get_max_num_cells
```
gonuke:
```suggestion
      // has_cell_fracs is assigned in get_max_num_cells
```

gonuke: I'm not sure I like the idea of `has_cell_fracs` being assigned without being anywhere in the function interface.  This is something that will easily be forgotten at some point.

Is there some step that is more like initialization/file reading when we can set this flag?

news/speed_up_source_sampling.rst
```diff
@@ -0,0 +1,16 @@
+**Added:** None
+
+**Changed:**
+
+* Pass cell_list back to Fortran, to speed up source sampling.
+* Update the test functions acoording to the cell_list change.
```
bam241:
```suggestion
* Update the test functions according to the cell_list change.
```

src/source_sampling.cpp
zxkjack123: I defined it as a const int `SUBVOXEL_START` in the code.

zxkjack123:
Updates according to the suggestion of @gonuke :
- Add a function to assign `has_cell_fracs`
- Add a const int `SUBVOXEL_START` of the mode

share/source.F90
```diff
@@ -12,36 +12,96 @@
...
+    ! If the cell_list is given (for HEX mesh),
+    ! use it to find cell first
+    if (cell_list_size > 0) then
+        ! HEX mesh. VOXEL/SUBVOXEL
+        do i = 1, cell_list_size
+           if (cell_list(i) .le. 0) then
+               ! not a valid cell number (-1)
+               exit
+           endif
+           cid = namchg(1, cell_list(i))
```
gonuke: `cid` is actually a cell index and not a cell ID... maybe we should rename this to `cidx`

gonuke (commented):
I recommend one change in variable name.  Otherwise this looks great!

gonuke:
@zxkjack123 - I think a conflict was created when I merged #1133 - please resolve and then I'll merge this

zxkjack123:
Update: Resolved conflicts by apply the #1133 into this PR and fix the file with conflicts.

zxkjack123:
@gonuke I think this PR is ready to merge now.

gonuke:
Thanks to everyone who put a lot of energy into this important update - and especially @zxkjack123 for all of his great analysis.

PR #1121, opened 2019-04-10
Speed up r2s step2 1

zxkjack123:
The changes in this PR intend to speed up the r2s step2.
Using the "line_profile" tool, I found that 87.1% CPU time of `photon_source_hdf5_to_mesh` is used in the code block to get the `phtn_src_dc` (from line number 236 to 240). It iterates in the ranges of `number of cooling times` \* `number of nulicdes`  \* `number of voxels` to get the list of cooling times. This is the hotspot of the function.

The changes of this PR reduce the iteration time. By applying this PR, the CPU time of the code block to get the `phtn_src_dc` takes about 0.2% of the function `photon_source_hdf5_to_mesh`.

Major changes:
- Change the method of getting `phtn_src_dc`, speed up the r2s step2.

### DIFF
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -233,8 +233,12 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
     # creat a list of decay times (strings) in the source file
     phtn_src_dc = []
     with tb.open_file(filename) as h5f:
+#        import pdb; pdb.set_trace()
         for row in h5f.root.data:
-            phtn_src_dc.append(row[2])
+            if row[2] not in phtn_src_dc:
+                phtn_src_dc.append(row[2])
+            else:
+                break
     phtn_src_dc = list(set(phtn_src_dc))
 
     # iterate through each requested nuclide/dectay time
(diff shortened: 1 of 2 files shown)


gonuke:
Thanks @zxkjack123

pyne/alara.py
```diff
@@ -234,8 +234,10 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
     phtn_src_dc = []
     with tb.open_file(filename) as h5f:
         for row in h5f.root.data:
-            phtn_src_dc.append(row[2])
-    phtn_src_dc = list(set(phtn_src_dc))
+            if row[2] not in phtn_src_dc:
```
gonuke, line 237: Please add a comment that explains why this is correct, particularly what the known/assumed order of the rows in the table are.

zxkjack123: The content in the `h5f.root.data` is a table with data type of 
```
dt = np.dtype([
        ('idx', np.int64),
        ('nuc', 'S6'),
        ('time', 'S20'),
        ('phtn_src', np.float64, G),
    ])
```
The `time` changes fastest. And for every (`idx`, `nuc`) pair, it has the exactly the same decay times. The size of the table if `number of voxels` \* `average number of nuclides in each voxel` \* `number of decay times`.

If we want to get the decay times, we don't need to iterate the entire table, what we need is to iterate the first (`idx`, `nuc`) pair and then we get decay times.

This PR reduces the iterate time form `number of voxels` \* `average number of nuclides in each voxel` \* `number of decay times` to `number of decay times` + 1.

PR #1104, opened 2019-02-28
Add MaterialLibrary to the C++ API 

bam241:
This is an attempt to add MaterialLibrary to the C++ API.

For the moment, the python API is not bind to it.

I also made some assumptions about how should behave the `write_hdf5` method. For the moment, I believe the behavior is very similar to the Python one: 
- if the file exists, try to write the materials in it regardless of what is in the file, 
- create the file otherwise.

For the `from_hdf5`, I made in in a way to in append the new material in the existing MaterialLibrary, allowing to merge multiple MaterialLibrary (form the file) into a single one.

To be done:
- [x] Unit test
- [x] Debug
- [x] Improve the behavior when attempting to write in a file already containing some materials

A [summary of some changes](https://docs.google.com/spreadsheets/d/16mfZiYD6XpzGfhVLre-RBiQgGX1WQMJtZVXTCSF839U/edit#gid=0) in `MaterialLibrary` and `Material` I/O.

PR #1122, opened 2019-04-18
GDML material input

bam241:
This adds the capability to write material input deck.
Documentation about GDML material input can be found [[here](http://lcgapp.cern.ch/project/simu/framework/GDML/doc/GDMLmanual.pdf)], p12-13.

Material input in GDML should have 3 sections:
`<isotope> </isotope>`
`<element></element>`
`<material></material>`
to avoid redundancy within the isotopes declaration I did't include the part writing the isotopes in the material.
I was thinking about adding the capability in `nucname.cpp` and/or in MaterialLibrary (see #1104 )

This also adds standard chemical element names from [wikipedia](https://en.wikipedia.org/wiki/List_of_chemical_elements), allowing to name the different element present int the material.

As the element seems to be required to build a material in GDML format, this loop over the material composition, group isotopes by elements, and build the elements with the proper abundances.
It then build a GDML material using the different elements previously build.

To build the elements input, the standard `nucname::name` are used for the isotopes (i.e.: `Am241`).
To avoid conflict if other material have the same elements (with or without the same abundances) the elements are name as: `MATNAME_ELEMENTNAME`.

To be done:
- [x] python binding
- [x] nosetests
- [ ] Isotopes GDML input deck from ??


pyne/material.pyx
```diff
@@ -345,6 +345,19 @@ cdef class _Material:
         cdef std_string card
         card = self.mat_pointer.mcnp(frac_type.encode())
         return card.decode()
+    
+    def gdml(self):
```
zxkjack123, line 375: Docstring of this function is not clear. 
- Parameters are usually before Return.
- There isn't an input paremeter, what's the `int 0`?

bam241: sorry, I forgot the clean the copy paste :)

thx for the catch!

bam241:
GDML needs the isotopes definition in the input deck, but to avoid redundancy I am not sure it should occur at the material level...

bam241:
@gonuke this should be ready for a round of reviews.

src/material.cpp
```diff
@@ -961,6 +961,79 @@ std::string pyne::Material::mcnp_frac(std::map<int, double> fracs, std::string f
...
+    mat_name = metadata["name"].asString();
+  }
+  if (metadata.isMember("mat_number")) {
+    mat_name_by_num += std::to_string(metadata["mat_number"].asInt());
+  }
+
+  // 'Isotope list'
+
+  // 'Element list'
+  std::map<int, double> fracs = comp;
```
gonuke, line 980: I don't think this is used.

src/material.cpp
```diff
@@ -961,6 +961,79 @@ std::string pyne::Material::mcnp_frac(std::map<int, double> fracs, std::string f
...
+  // loop over the composition and group isotopes per element
+  for (auto it : comp) {
+    int z = nucname::znum(it.first);
+    std::string element_name = mat_name_by_num + "_" + nucname::name_elt[nucname::zz_name[z]];
+    if (elements_list.find(z) == elements_list.end()) {
+      std::map<int, double> list;
+      list[it.first] = it.second;
+      elements_list.insert(std::pair<int, std::map<int, double>>(z, list));
+
+      // add fraction in the material composition
```
gonuke, line 992: why use `element_name` here instead of `z`?

src/material.cpp
```diff
@@ -961,6 +961,79 @@ std::string pyne::Material::mcnp_frac(std::map<int, double> fracs, std::string f
...
+      element_comp.insert( std::make_pair(element_name, it.second));
+    } else {
+      elements_list[z][it.first] = it.second;
+      // add fraction in the material composition
+      element_comp[element_name] += it.second;
+    }
+  }
+
+  // loop over the different chemical elements
+  for (auto it : elements_list) {
```
gonuke, line 1002: would it be more readable to use `nuclide` instead of `it`?

src/nucname.cpp
```diff
@@ -6,8 +6,148 @@
...
+  lname["Be"] = "Beryllium";
+  lname["B"] = "Boron";
+  lname["C"] = "Carbon";
+  lname["N"] = "Nitrogen";
+  lname["O"] = "Oxygen";
+  lname["F"] = "Fluorine";
+  lname["Ne"] = "Neon";
+  lname["Na"] = "Sodium";
+  lname["Mg"] = "Magnesium";
+  lname["Al"] = "Aluminium";
```
gonuke, line 28: What is the basis for these spellings.  Are they based on something from Geant?  They may not be universal....

gonuke (changes requested):
A few little changes, but seems mostly OK

PR #1123, opened 2019-04-18
Activation responses

zxkjack123:
Suggested by @bohmt, it is useful to do some 3D mesh map enhancement for better visualization of the data. This PR adds a script to extract activation responses, such as decay heat, and write to h5 file for visualization, using a procedure similar to r2s:
- setup: write the `config.ini` and `alara_params.txt`
- step1: write `alara_inp`, `alara_matlib` and `alara_fluxin`
- step2: read responses from `output.txt` and write to h5 file

Setting the response as `photon_source`, then it is r2s workflow (but sub-voxel mode not supported). 

To do:
- [x] : add code to get `decay_heat`
- [x] : add nosetests for `decay_heat`
- [x] : add other responses and tests
    - [x] : specific activity
    - [x] : alpha heat
    - [x] : beta heat
    - [x] : gamma heat
    - [x] : wdr
    - [x] : photon_srouce (total)
- [x] : add the capability to treat multiple responses at once


pyne/alara.py
```diff
@@ -178,6 +178,111 @@ def photon_source_to_hdf5(filename, chunkshape=(10000,)):
...
+    chunksize = chunkshape[0]
+    rows = np.empty(chunksize, dtype=dt)
+    idx = 0
+    count = 1
+    decay_times = []
+    zone_start = False
+    response_start = False
+    for i, line in enumerate(f, 1):
+        # terminate condition
+        if 'Totals for all zones' in line and response_start:
```
moatazharb: Wouldn't it be better to check the existence of "Totals for all zones" in a line by checking the existence of "Totals for all zones".lower().split() in line.lower().split(). just thinking of how to make it more generic in case any format changed that led to a difference in spacing or capitalization.

moatazharb: are you missing a "." here?. shouldn't it be 'Totals for all zones.'

pyne/alara.py
```diff
@@ -178,6 +178,111 @@ def photon_source_to_hdf5(filename, chunkshape=(10000,)):
...
+            break
+        # get response string
+        if zone_start and is_response_string(line, response):
+            response_start = True
+            continue
+        # get decay times
+        if 'isotope\t shutdown' in line:
```
moatazharb: why is there a \t ?
You don't need it if you dopt the previous suggestion of using 'isotope shutdown'.lower().split() in line.lower().split()

pyne/alara.py
```diff
@@ -923,3 +1093,213 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    Returns
+    -------
+    True : if this line contains results data
+    False : if this line doesn't contain results data
+    """
+    # check the list from the second value, if they are float, then return True
+    ls = line.strip().split()
+    if len(ls) < 2:
+        return False
+    for i in range(1, len(ls)):
```
moatazharb: you could try here something like:
`try:
    np.array(ls[1:]).astype(float)
except:
    return False`
instead of the for loop

pyne/alara.py
```diff
@@ -178,6 +178,111 @@ def photon_source_to_hdf5(filename, chunkshape=(10000,)):
...
+        if not _is_data(line):
+            continue
+    
+        ls = line.strip().split()
+        # put data into table
+        # format of each row: idx, nuc, time, decay_heat
+        for k, dc in enumerate(decay_times):
+            j = (count-1) % chunksize
+            nuc = ls[0].strip()
+            if nuc in ('total', 'Total'):
```
moatazharb: nuc.lower() == 'total'

pyne/alara.py
```diff
@@ -923,3 +1093,213 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    line : string
+        A line from ALARA output.txt
+
+    Returns
+    -------
+    decay_times : array of string
+        Array of decay times.
+    """
+    ls = line.strip().split()
+    decay_times = ['shutdown']
```
moatazharb, line 1223: is not "shutdown" already in the line?

pyne/alara.py
```diff
@@ -278,6 +376,77 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
...
+    for tag_name in tags.values():
+
+        mesh.tag(tag_name, np.zeros(1, dtype=float), 'nat_mesh',
+                 size=1, dtype=float)
+        tag_handles[tag_name] = mesh.get_tag(tag_name)
+
+    phtn_src_dc = _read_phtn_src_dc(filename)
+
+    # iterate through each requested nuclide/dectay time
+    for cond in tags.keys():
```
moatazharb, line 440: what does "cond" stand for?

pyne/alara.py
```diff
@@ -278,6 +376,77 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
...
+                nuc = "TOTAL"
+
+            # time match, convert string mathch to float mathch
+            dc = _find_phsrc_dc(cond[1], phtn_src_dc)
+            # create of array of rows that match the nuclide/decay criteria
+            matched_data = h5f.root.data.read_where(
+                "(nuc == '{0}') & (time == '{1}')".format(nuc, dc))
+
+        idx = 0
+        for i, _, ve in mesh:
```
moatazharb, line 458: Could you include a snippet of what mesh looks like (first line may be in the iterator) so that it would be more obvious that "i" is the id and "ve" is for volume id (maybe) and it would be clear why there's a "_"? 
I know this format is found in other places in PyNE, but maybe better to use 
for m in mesh:
then you'd just replace i with m[0] and ve with m[2]

pyne/alara.py
```diff
@@ -911,7 +1081,7 @@ def _find_phsrc_dc(idc, phtn_src_dc):
         # Loop over decay times in phtn_src_dc list and compare to idc_s.
         for dc in phtn_src_dc:
             # Skip "shutdown" string in list.
-            if dc == u'shutdown':
+            if str_to_unicode(dc) == u'shutdown':
```
moatazharb: Is there a reason why this is in unicode? could that be generalized in other parts of the script or it is just a special case here?

scripts/activation_responses.py
```diff
@@ -0,0 +1,189 @@
...
+# with their counterparts in the phtn_src file produced in step1. No spaces
+# should appear in this line except the space between the time and the time unit
+# for each entry.
+decay_times:1E3 s,12 h,3.0 d
+"""
+
+alara_params =\
+    """material_lib alara_matlib
+element_lib nuclib
+data_library alaralib fendl3bin
```
moatazharb, line 69: Does ALARA now work with fendl3?

scripts/activation_responses.py
```diff
@@ -0,0 +1,189 @@
...
+
+#     flux name    fluxin file   norm   shift   unused
+flux  my_flux     alara_fluxin  1e10     0      default
+
+# Specify the irradiation schedule below.
+# Syntax is found in the ALARA user manual
+# This example is for a single 3.5 h pulse
+schedule    my_schedule
+    3.5 d my_flux my_pulse_history 0  s
```
moatazharb, line 79: This is a typo that has always been there in alara in r2s scripts!. It says it is a 3.5 h pulse yet in the line below it is 3.5 d !!

scripts/activation_responses.py
```diff
@@ -0,0 +1,189 @@
...
+"""
+
+def setup():
+    with open(config_filename, 'w') as f:
+        f.write(config)
+    with open(alara_params_filename, 'w') as f:
+        f.write(alara_params)
+    print('File "{}" has been written'.format(config_filename))
+    print('File "{}" has been written'.format(alara_params_filename))
+    print('Fill out the fields in these filse then run ">> activation_response.py step1"')
```
moatazharb: typo in "filse"

scripts/activation_responses.py
```diff
@@ -0,0 +1,189 @@
...
+                      tally_num, num_rays=num_rays, grid=grid, reverse=reverse,
+                      output_mesh="activation_responses_step1.h5m",
+                      flux_tag=flux_tag, decay_times=decay_times,
+                      sub_voxel=False, responses=responses, wdr_file=wdr_file)
+
+    # create a blank mesh for step 2:
+    ves = list(flux_mesh.iter_ve())
+    for tag in flux_mesh.get_all_tags():
+        tag.delete()
+    flux_mesh.write_hdf5('blank_mesh.h5m')
```
moatazharb, line 139: maybe rename the file to be more specific not to be confused with r2s blank mesh file

scripts/activation_responses.py
```diff
@@ -0,0 +1,189 @@
...
+    print('Decay_heat step1 complete, run ALARA with the command:')
+    print('>> alara alara_inp > output.txt')
+
+
+def step2():
+    config = ConfigParser.ConfigParser()
+    config.read(config_filename)
+    structured = config.getboolean('general', 'structured')
+    decay_times = config.get('step2', 'decay_times').split(',')
+    #output = config.get('step2', 'output')
```
moatazharb: It is a good idea to have the user specify the name of the alara output file in case the the user named the file during alara run. You could have it set to "output.txt" by default

scripts/activation_responses.py
```diff
@@ -0,0 +1,189 @@
...
+    step2_help = 'Creates mesh-based activation response data for visulization.'
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers(help=activation_response_help, dest='command')
+
+    setup_parser = subparsers.add_parser('setup', help=setup_help)
+    step1_parser = subparsers.add_parser('step1', help=step1_help)
+    step2_parser = subparsers.add_parser('step2', help=step2_help)
+
+    args, other = parser.parse_known_args()
+    if args.command == 'setup':
```
moatazharb, line 188: maybe:

`if args.command == "setup":

    setup()

else:

    config = ConfigParser.ConfigParser()
    config.read(config_filename)
    if  args.command == "step1":
        step1(config)
    elif args.command == "step2":
        step2(config)`
and then pass config as input

tests/files_test_activation_responses/activation_responses_examples/alara_params.txt
```diff
@@ -0,0 +1,21 @@
+material_lib alara_matlib
+element_lib data/nuclib
+data_library alaralib data/fendl2.0bin
+
+#     flux name    fluxin file   norm   shift   unused
+flux  my_flux     alara_fluxin  1e10     0      default
```
moatazharb, line 6: Is it more safe to have the normalization set to 1.0?. I guess it is better to have the user to actively change it to the desired normalization and the default to be providing a normalized tally as input.

tests/files_test_activation_responses/activation_responses_examples/alara_params.txt
```diff
@@ -0,0 +1,21 @@
...
+
+# Specify the irradiation schedule below.
+# Syntax is found in the ALARA user manual
+# This example is for a single 3.5 h pulse
```
moatazharb: either h or d in the both the description line and example

pyne/alara.py
zxkjack123: Changes as your suggestion in https://github.com/pyne/pyne/pull/1123/commits/52d7393dc393e0b2a5a39d4246a31f102b4ced49.

pyne/alara.py
zxkjack123: Changed as your suggestion in https://github.com/pyne/pyne/pull/1123/commits/8b00e875a84baa0ffb2679c76fc69e7bb16fc21d.

The `\t` here is the Tab in the output file. ALARA use Tab at many places.

pyne/alara.py
zxkjack123: Well, yes, a `.` is missing here. But it doesn't matter.

pyne/alara.py
zxkjack123: Changed in https://github.com/pyne/pyne/pull/1123/commits/1d234997aff5ff3c77e3c6defa8a4107ef8b4cc4.

pyne/alara.py
zxkjack123: Changed in https://github.com/pyne/pyne/pull/1123/commits/54e568fdf6ca62ab96a563d8dc0ae48b907c4ac6.

scripts/activation_responses.py
zxkjack123: Actually, this `blank_mesh.thm` works exactly the same as the `blank_mesh.h5m` in R2S. Their content is almost the same. I think kept its name is OK as it's not the final output.

scripts/activation_responses.py
zxkjack123: Changed in https://github.com/pyne/pyne/pull/1123/commits/9a61e614ffc8d76a9af224556d7e47ae8040f1f8.

tests/files_test_activation_responses/activation_responses_examples/alara_params.txt
zxkjack123: I don't think `1.0` is a safer number here. This `1e10` here is easier for users to notice that this is a number need to be replaced! If it is `1.0`, users may ignore it and not change it.

tests/files_test_activation_responses/activation_responses_examples/alara_params.txt
zxkjack123: Changed in https://github.com/pyne/pyne/pull/1123/commits/3978c22844d0e8fda4de1b099e401ca3b0b51224.

scripts/activation_responses.py
zxkjack123: Where is the `config_filename` defined?

pyne/alara.py
zxkjack123: The `i, _, ve` here means mesh element index, material, volume element. And there isn't a iterator to work as 
```python
for m in mesh
```.

Comment (zxkjack123) [reply pyne/alara.py:458]: I add a comment line here in https://github.com/pyne/pyne/pull/1123/commits/f28cf3d0b7f9fa427fbbcfe081c4c32227c5f64d.

File: pyne/alara.py (hunk 5 above)
Comment (zxkjack123) [reply pyne/alara.py:1223]: Yes, the `shutdown` is in the line. But `shutdown` does not have a unit, the format is different.

File: pyne/alara.py (hunk 6 above)
Comment (zxkjack123) [reply pyne/alara.py:440]: Each element in `tags` is a `(nuc, cooling_time)` pair. I think the `cond` here stands for the condition.

File: scripts/activation_responses.py (hunk 11 above)
Comment (zxkjack123) [reply scripts/activation_responses.py]: Thanks, fixed in https://github.com/pyne/pyne/pull/1123/commits/eafb5aff5ce47e971d0b84c9692ad2dfb301b06d.

File: pyne/alara.py (hunk 8 above)
Comment (zxkjack123) [reply pyne/alara.py]: The strings with prefix `u` is usually for the compatibility of python2/python3, especially these strings read from a file.

Python2 and Python3 has a different default `str` type. And strings read from files sometimes have strnage coding format. Explictlly define a string as a specific type will solve the problem.

File: scripts/activation_responses.py (hunk 9 above)
Comment (zxkjack123) [reply scripts/activation_responses.py:69]: As far as I know, no.

File: scripts/activation_responses.py (hunk 14 above)
Comment (moatazharb) [reply scripts/activation_responses.py:188]: It is defined as a global variable below the imports

File: tests/files_test_activation_responses/activation_responses_examples/alara_params.txt (hunk 15 above)
Comment (moatazharb) [reply tests/files_test_activation_responses/activation_responses_examples/alara_params.txt:6]: Sure. But by  default (I assume) a user of MCNP might produce a normalized tally and provide it to ALARA without noticing the normalization that has to change.

I know that it depends, but most probably a mesh tally would be produced with a provided normalization MCNP. Plus, and error of not normalizing inside ALARA parameters file will result in a termination of the run because of truncation while a normalization applied twice might go for longer!.

File: scripts/activation_responses.py (hunk 14 above)
Comment (zxkjack123) [reply scripts/activation_responses.py:188]: My bad. I didn't noticed that. But I don't think this change makes any improvement in the performance. The `step1` and `step2` always run separately.

File: tests/files_test_activation_responses/activation_responses_examples/alara_params.txt (hunk 15 above)
Comment (zxkjack123) [reply tests/files_test_activation_responses/activation_responses_examples/alara_params.txt:6]: It's OK to do as what you mentioned, if there is only one multiplier (i.e., the power of the plant is a constant).

But how about the case that the power of plant changes at different irradiation time? Eg. `P1` MW in the first year, and `P2` MW in the second year. Users must be noticed to change the normalization when they use ALARA.

File: scripts/activation_responses.py (hunk 14 above)
Comment (moatazharb) [reply scripts/activation_responses.py:188]: Yes I don't think it has an impact on performance. I just thought to point out a different style for calling setup and steps 1 & 2.

File: tests/files_test_activation_responses/activation_responses_examples/alara_params.txt (hunk 15 above)
Comment (moatazharb) [reply tests/files_test_activation_responses/activation_responses_examples/alara_params.txt:6]: Sure. I didn't think about that!. Maybe add a comment on the config file where the meshtal file is to be specified that the user has to change normalization in the alara_parameters file.

Comment (zxkjack123) [reply tests/files_test_activation_responses/activation_responses_examples/alara_params.txt:6]: Added a comment in https://github.com/pyne/pyne/pull/1123/commits/c57c66eb4fbe1425f0f19b15cb48bb78f30d6790.

File: pyne/r2s.py [hunk 17]
```diff
@@ -75,7 +75,8 @@ def irradiation_setup(flux_mesh, cell_mats, cell_fracs, alara_params,
                       fluxin="alara_fluxin", reverse=False,
                       alara_inp="alara_inp", alara_matlib="alara_matlib",
                       output_mesh="r2s_step1.h5m", output_material=False,
-                      decay_times=None, sub_voxel=False):
+                      decay_times=None, sub_voxel=False, responses=None,
+                      wdr_file=None):
```
katyhuff, line 79: should this have a default value? Is the wdr file typically called something specific, or placed somewhere during pyne install? Should it be?

pyne/r2s.py
```diff
@@ -126,6 +127,10 @@ def irradiation_setup(flux_mesh, cell_mats, cell_fracs, alara_params,
         List of the decay times. If no decay times given, use '1 s'.
     sub_voxel : bool, optional
         If true, sub-voxel r2s work flow  will be used.
+    responses : list of str, optional
+        The list of requested responses.
+    wdr_file : str
+        File name of the standard to calculate wdr.
```
katyhuff: just the filename, or also the path?

pyne/utils.py
```diff
@@ -293,6 +294,32 @@ def line_almost_same(l1, l2, rel_tol=1e-9):
...
+    -----------
+    s : str
+        String to be checked.
+    line : str
+        The given line
+
+    Returns:
+    --------
+    True : bool
+        If all the elements of the string are in the line.
```
katyhuff:
```suggestion
        True if all the elements of the string are in the line.
```

pyne/utils.py
```diff
@@ -293,6 +294,32 @@ def line_almost_same(l1, l2, rel_tol=1e-9):
...
+    False : bool
+        If not.
```
katyhuff:
```suggestion
        False if one or more of the elements of the string is not found in the line.
```

scripts/activation_responses.py
```diff
@@ -0,0 +1,195 @@
...
+            mesh = Mesh(structured=structured, mesh='blank_mesh.h5m')
+            tags = {('TOTAL', dc): tag_name}
+            response_hdf5_to_mesh(mesh, ''.join([response, '.h5']), tags, response)
+            mesh.write_hdf5('{0}_{1}.h5m'.format(response, i+1))
+
+    print('Activation_response step2 complete.')
+
+
+def main():
+
```
katyhuff: It is customary to put a docstring on this line ( in this case, describing the activities of the main function)
Also, PEP8 discourages unnecessary blank lines.

tests/test_activation_responses.py
```diff
@@ -0,0 +1,277 @@
...
+else:
+    from itertools import izip
+
+warnings.simplefilter("ignore", QAWarning)
+
+thisdir = os.path.dirname(__file__)
+
+responses = ['decay_heat', 'specific_activity', 'alpha_heat', 'beta_heat',
+             'gamma_heat', 'wdr', 'photon_source']
+#responses = ['specific_activity']
```
katyhuff:
```suggestion
```

tests/test_activation_responses.py
```diff
@@ -0,0 +1,277 @@
...
+
+        # generate expected h5 file
+        exp_h5_filename = os.path.join(thisdir, "files_test_activation_responses", ''.join(['exp_', response, '.h5']))
+        _generate_exp_h5(filename, response, exp_h5_filename)
+
+        # compare two h5 files
+        command = ''.join(['h5diff ', h5_filename, ' ', exp_h5_filename])
+        diff_flag = os.system(command)
+        # return value 0 if no difference, 1 if differences found, 2 if error
+        assert_equal(diff_flag, 0)
```
katyhuff, line 133: If this test fails, it's not obvious to me that the user will learn what is different between the two h5 files in the test output. Will the h5diff output be printed? I think the failure case will simply say something like "diff_flag=2, expected 0."

tests/test_activation_responses.py
```diff
@@ -0,0 +1,277 @@
...
+    os.remove(h5m_filename)
+    os.remove(dst)
+
+    # return value 0 if no difference, 1 if differences found, 2 if error
+    assert_equal(diff_flag4, 0)
+    assert_equal(diff_flag5, 0)
+
+
+def test_activation_responses_script():
+
```
katyhuff:
```suggestion
```

tests/test_alara.py
```diff
@@ -647,3 +649,79 @@ def test_phtn_src_energy_bounds():
...
+
+
+def test_alara_responses_output_zone():
+    # test decay_heat
+    response = ['decay_heat', 'specific_activity']
+    exp_response_code = \
+"""output zone
+      total_heat
+      specific_activity
+end"""
```
katyhuff, line 664: Can the all-the-way-to-the-right indentation be avoided? It will probably fail an autopep8 check.

tests/test_alara.py
```diff
@@ -647,3 +649,79 @@ def test_phtn_src_energy_bounds():
...
+    response_code = responses_output_zone(response)
+    assert_equal(response_code, exp_response_code)
+
+
+def test_is_data():
+    """
+    This function tests alara.is_data, which check whether a line in the output
```
katyhuff:
```suggestion
    This function tests alara.is_data, which checks whether a line in the output
```

katyhuff (commented):
I don't have much background to review the ALARA components of this, so I've left a non-committal set of comments regarding code style and clarity.

tests/test_alara.py
gonuke: Do you mean all-the-way-to-the-left?  I'm not sure there is a better way to generate output with appropriate indentation?

pyne/r2s.py
zxkjack123: The `wdr_file` here should not have a default value. This is a user provided file. This file is not part of pyne, it is for ALARA. Users have to define this file explicitly.

pyne/r2s.py
zxkjack123: I should be the path. I will update the doc here.

tests/test_activation_responses.py
zxkjack123: Well, the comparison of the `h5` files is actually put here for developers, but not for users. Users who install the pyne will not encounter a failure of the test.
This `h5diff` output is not printed.

Once the error happens, it usually means the data in the `h5` file are different, the developer should check the data (with `h5dum`) and find out the reason.

pyne/utils.py
```diff
@@ -293,6 +294,32 @@ def line_almost_same(l1, l2, rel_tol=1e-9):
         return True
 
 
+def str_elements_in_line(s, line):
```
gonuke: I'm concerned that this method doesn't seem to care about the order of the string elements that are found in the line.  It may work for the provided strings, but it seems fragile.

Would regular expression (`import regex`) matching work better?

pyne/alara.py
```diff
@@ -198,6 +199,112 @@ def photon_source_to_hdf5(filename, nucs='all', chunkshape=(10000,)):
...
+    tab = h5f.create_table('/', 'data', dt, chunkshape=chunkshape)
+
+    chunksize = chunkshape[0]
+    rows = np.empty(chunksize, dtype=dt)
+    idx = 0
+    count = 1
+    decay_times = []
+    zone_start = False
+    response_start = False
+    for i, line in enumerate(f, 1):
```
gonuke: Do you use this variable `i` or could you instead have:
```suggestion
    for line in f:
```

pyne/alara.py
```diff
@@ -943,3 +1116,214 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    """
+    lines = alara_params.split('\n')
+    for line in lines:
+        if "data_library" in line:
+            alara_lib = line.strip().split()[-1]
+            return alara_lib
+    raise ValueError("alara_lib not found!")
+
+
+def is_response_string(line, response):
```
gonuke: Not sure we need a function for this.

The dictionary can just be declared at a global scope and the `if ... in line` test can just be used directly.

pyne/alara.py
```diff
@@ -943,3 +1116,214 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    if 'wdr' in responses:
+        code_block= ''.join(["       wdr ", wdr_file, "\n"])
+    # define code block for photon_source
+    if 'photon_source' in responses:
+        alara_lib = get_alara_lib(alara_params)
+        code_block= ''.join([code_block, "       photon_source ", alara_lib,
+                             " phtn_src 1 2e7\n"])
+    return ''.join([start_str, code_block, end_str])
+
+def _is_data(line):
```
gonuke, line 1176: This function could perhaps be more robust in the following ways:
* check that the first entry in `ls` is in fact a valid `nucname` or `lower(TOTAL)`
* pass in the number of shutdown times and makes sure the list of floats is long enough

pyne/alara.py
```diff
@@ -198,6 +199,112 @@ def photon_source_to_hdf5(filename, nucs='all', chunkshape=(10000,)):
...
+                zone_start = True
+            continue
+        # skip the lines does not contain specific response info
+        if not response_start:
+            continue
+        # skip the lines don't contain wanted data
+        if not _is_data(line):
+            continue
+    
+        ls = line.strip().split()
```
gonuke: Since the first thing that `_is_data()` does is split this string, why not do this line first, and pass `ls` into `_is_data()`?

gonuke: A better name than `ls` might be `tokens` or `words`

pyne/alara.py
```diff
@@ -198,6 +199,112 @@ def photon_source_to_hdf5(filename, nucs='all', chunkshape=(10000,)):
...
+        # put data into table
+        # format of each row: idx, nuc, time, decay_heat
+        for k, dc in enumerate(decay_times):
```
gonuke:
```suggestion
        for dc, response_value in zip(decay_times,ls[1:]):
```

see below for related changes

pyne/alara.py
```diff
@@ -198,6 +199,112 @@ def photon_source_to_hdf5(filename, nucs='all', chunkshape=(10000,)):
...
+            j = (count-1) % chunksize
+            nuc = ls[0].strip()
```
gonuke: this line should be outside/before the `for` loop

pyne/alara.py
```diff
@@ -198,6 +199,112 @@ def photon_source_to_hdf5(filename, nucs='all', chunkshape=(10000,)):
...
+            continue
+        # skip the lines don't contain wanted data
+        if not _is_data(line):
+            continue
+    
+        ls = line.strip().split()
+        # put data into table
+        # format of each row: idx, nuc, time, decay_heat
+        for k, dc in enumerate(decay_times):
+            j = (count-1) % chunksize
```
gonuke: This line is only used in one place, so it might be good to move it to just before that place.

Some might argue that this could just be done inline, without introducing this variable `j` at all, but it might help readability a little bit.  It might be better if it was called `chunk_row` or something?

pyne/alara.py
```diff
@@ -198,6 +199,112 @@ def photon_source_to_hdf5(filename, nucs='all', chunkshape=(10000,)):
...
+            nuc = ls[0].strip()
+            if nuc.lower() == 'total':
+                nuc = nuc.upper()
+            rows[j] = (idx, nuc, dc, float(ls[k+1]))
```
gonuke: to match previous change
```suggestion
            rows[j] = (idx, nuc, dc, response_value)
```

pyne/alara.py
```diff
@@ -198,6 +199,112 @@ def photon_source_to_hdf5(filename, nucs='all', chunkshape=(10000,)):
...
+            response_start = True
+            continue
+        # get decay times
+        if str_elements_in_line('isotope shutdown', line):
+            if len(decay_times) == 0:
+                decay_times = read_decay_times(line)
+            continue
+        # get zone idx
+        if 'Zone #' in line:
+            idx = _get_zone_idx(line)
```
gonuke: This is a mesh index, right?  maybe `mesh_idx` would be helpful?

pyne/alara.py
```diff
@@ -198,6 +199,112 @@ def photon_source_to_hdf5(filename, nucs='all', chunkshape=(10000,)):
...
+            if idx == 0:
+                ## new blocks, disable first response string
+                #response_start = False
```
gonuke: is this meant to be an intentional comment to describe the situation, or a line that you have commented out?

pyne/alara.py
```diff
@@ -10,7 +10,8 @@
 import os
 import collections
 from warnings import warn
-from pyne.utils import QAWarning, to_sec
+from pyne.utils import QAWarning, to_sec, is_float, str_to_unicode, \
```
gonuke: `is_float` isn't used

pyne/alara.py
```diff
@@ -298,6 +398,78 @@ def photon_source_hdf5_to_mesh(mesh, filename, tags, sub_voxel=False,
...
+
+    # create a dict of tag handles for all keys of the tags dict
+    tag_handles = {}
+    for tag_name in tags.values():
+
+        mesh.tag(tag_name, np.zeros(1, dtype=float), 'nat_mesh',
+                 size=1, dtype=float)
+        tag_handles[tag_name] = mesh.get_tag(tag_name)
+
+    phtn_src_dc = _read_phtn_src_dc(filename)
```
gonuke: This can be used for responses other than photon source, right?  Is this variable name confusing?

gonuke: It seems that there are more places below where the name based on photon source is used

pyne/alara.py
```diff
@@ -943,3 +1116,214 @@ def _find_phsrc_dc(idc, phtn_src_dc):
         # if idc doesn't match any string in phtn_src_dc list, raise an error.
         raise ValueError(
             'Decay time {0} not found in phtn_src file'.format(idc))
+
+
+def responses_output_zone(responses=None, wdr_file=None, alara_params=None):
+    """
+    This function returns a string representing the output zone of alara input
```
gonuke: Slightly better wording:

This function returns a string that is an output block for an alara input file, configured for zone resolution.

pyne/alara.py
```diff
@@ -943,3 +1116,214 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+
+    # set default value for functions do not need response
+    if responses == None:
+        return ''
+    # input check
+    for response in responses:
+        if response not in ('decay_heat', 'photon_source', 'specific_activity',
+                        'alpha_heat', 'beta_heat', 'gamma_heat', 'wdr'):
+            raise ValueError('response {0} not supported.'.format(response))
+
```
gonuke, line 1158: I think a more streamlined version of this would be:
* create a dictionary of response -> string for output block
* start a list of strings with "output zone"
* for each response, add its string from the dictionary
* add "end" to the list of strings
* `'\n'.join()` the list

gonuke: you'll need to treat WDR and photon source specially

pyne/alara.py
```diff
@@ -943,3 +1116,214 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    try:
+        np.array(ls[1:]).astype(float)
+        return True
+    except:
+        return False
+    
+
+def read_decay_times(line):
+    """
+    This function reads a line contian decay times information from alara
```
gonuke, line 1209:
```suggestion
    This function reads a line containing decay times information from alara
```

pyne/alara.py
```diff
@@ -943,3 +1116,214 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+
+    Returns
+    -------
+    decay_times : array of string
+        Array of decay times.
+    """
+    ls = line.strip().split()
+    decay_times = ['shutdown']
+    for i in range(2, len(ls), 2):
+        decay_times.append(''.join([ls[i], ' ', ls[i+1]]))
```
gonuke: can't you just add strings?
```suggestion
        decay_times.append(ls[i] + ' ' + ls[i+1])
```

pyne/alara.py
```diff
@@ -943,3 +1116,214 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+
+    Parameters
+    ----------
+    line : string
+        A line from ALARA output.txt
+
+    Returns
+    -------
+    int, zone index
+    """
```
gonuke, line 1241: extra blank line
```suggestion
    """
    
```

pyne/alara.py
```diff
@@ -943,3 +1116,214 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    ls = line.strip().split()
```
gonuke:
```suggestion
    last_word = line.strip().split()[-1]
```

pyne/alara.py
```diff
@@ -943,3 +1116,214 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+    return int(ls[-1].split('_')[-1])
```
gonuke:
```suggestion
    return int(last_word.split('_')[-1])
```

pyne/alara.py
```diff
@@ -943,3 +1116,214 @@ def _find_phsrc_dc(idc, phtn_src_dc):
...
+                        'beta_heat': 'Beta Decay Heat',
+                        'gamma_heat': 'Gamma Decay Heat',
+                        'wdr': 'WDR/Clearance index',
+                        'photon_source': 'Photon Source Distribution'}
+    if response_strings[response] in line:
+        return True
+    else:
+        return False
+
+def _read_phtn_src_dc(filename):
```
gonuke: As noted above, this is no longer for just photon sources, right?

scripts/activation_responses.py
gonuke: should probably change this default then:
```suggestion
data_library alaralib fendl2bin
```

scripts/activation_responses.py
```diff
@@ -0,0 +1,197 @@
+#!/usr/bin/env python
```
gonuke, line 1: Is there a way to merge this with `r2s.py` or pull out some common elements?  It seems that the two files are very similar

gonuke (changes requested):
This is a great addition!  I have a lot of changes to request, mostly to improve long term readability.

I didn't review the tests yet.

scripts/activation_responses.py
gonuke: Maybe that could be left for another PR??

pyne/utils.py
zxkjack123: I will try "regex" later in new updates.

pyne/alara.py
zxkjack123: We don't need it anymore, simplified in https://github.com/pyne/pyne/pull/1123/commits/e1a3dc87bcd667a8950fac97323ae478ee918115.

pyne/alara.py
zxkjack123: Updated the check in https://github.com/pyne/pyne/pull/1123/commits/8112a4f3c77232dac68f05e98bcafbc3cb990348.

pyne/alara.py
zxkjack123: Updated this part in https://github.com/pyne/pyne/pull/1123/commits/e6069dce639cc742d7e8b30e4f65420e4c0d848b. Now `j` is defined just before it is used.

pyne/alara.py
zxkjack123: Actually, it should be `zone_idx`. For the activation responses part, the `zone_idx` is equivalent to `mesh_idx`. But we do have sub-voxel R2S part, where `zone_idx` does not equal to `mesh_idx`.
For better understanding, I changed it to `zone_idx` in https://github.com/pyne/pyne/pull/1123/commits/440ab3560bfd3fe0e37351dd7e8115a10824c696.

pyne/alara.py
zxkjack123: This function can be used for both photon source and the responses. For better consistency, I removed `phtn_src` and rename the `dc` to `decay_times` in https://github.com/pyne/pyne/pull/1123/commits/9a61ffa550cfa61a87cefb8c580d2a20817d8696.

pyne/alara.py
zxkjack123: Thanks. Suggestion adopted in https://github.com/pyne/pyne/pull/1123/commits/6997cf65c8bf97eb09781ff4645e750168820daa.

scripts/activation_responses.py
zxkjack123: Yes, this file and the `r2s.py` shares almost the same work flow. We can combine them together. However, I don't think it's a good idea to merge them at the current stage.
Because some variables work differently in the two files. For example:
- the `sub-voxel` option could not used in `activation_responses.py` now;
- the `photon_source` meaning different thing in `r2s.py` (multiple group` and `activation_responses.py` (total`)

I think it's better to merge these two files when every thing is done with this PR, and we need also wait for the changes related to DagOpenMC R2S.

pyne/alara.py
```diff
@@ -887,59 +1063,230 @@ def _get_subvoxel_array(mesh, cell_mats):
...
+
+    # set default value for functions do not need response
+    if responses == None:
+        return ''
+    # input check
+    for response in responses:
+        if response not in response_strings.keys():
+            raise ValueError('response {0} not supported.'.format(response))
+
+    output_strings = {"decay_heat": "      total_heat\n",
```
gonuke: Define this dict outside of loop

gonuke: Oh - maybe not since you change it for WDR each time

pyne/alara.py
```diff
@@ -887,59 +1063,230 @@ def _get_subvoxel_array(mesh, cell_mats):
...
+    -------
+    True : if this line contains results data
+    False : if this line doesn't contain results data
+    """
+    # check the list from the second value, if they are float, then return True
+    ls = line.strip().split()
+    if len(ls) < 2:
+        return False
+    # first block should be a valid nucname or 'total'
+    if not (nucname.isnuclide(ls[0]) or ls[0] == 'TOTAL'.lower()):
```
gonuke: Doe the `not` apply to both conditions?

gonuke: Yep - I just counted the () more carefully

pyne/alara.py
zxkjack123: Use a line as input parameter to `_is_data()` because it has better readability, and it's easier to test and maintenance.

The `ls` is replaced by `tokens` in https://github.com/pyne/pyne/pull/1123/commits/29b3b6c00175c557617a23d08b959f7361cada92.

pyne/alara.py
```diff
@@ -887,59 +1063,231 @@ def _get_subvoxel_array(mesh, cell_mats):
...
+        output_strings['wdr'] = ''.join(["       wdr ", wdr_file, "\n"])
+    if 'photon_source' in responses:
+        alara_lib = get_alara_lib(alara_params)
+        output_strings["photon_source"] =  ''.join(["      photon_source ",
+            alara_lib, " phtn_src 1 2e7\n"])
+    output_zone = ["output zone\n"]
+    for response in responses:
+        output_zone.append(output_strings[response])
+    output_zone.append("end")
+    return ''.join(output_zone)
```
gonuke: Maybe remove the "\n" from each string and use `"\n".join()`

zxkjack123: Thanks, I applied it in https://github.com/pyne/pyne/pull/1123/commits/bc19b1d3625988e768697adf61d4b7bf2ab19da4.

pyne/utils.py
```diff
@@ -279,20 +280,46 @@ def line_almost_same(l1, l2, rel_tol=1e-9):
...
+                if str_almost_same(tokens1[i], tokens2[i], rel_tol):
                     pass
                 else:
                     return False
         return True
 
 
+def str_elements_in_line(s, line):
+    """
+    This function is used to check whether all the elements of a string is in
```
gonuke:
```suggestion
    This function is used to check whether all the words of a string is in
```

pyne/utils.py
```diff
@@ -279,20 +280,46 @@ def line_almost_same(l1, l2, rel_tol=1e-9):
...
+    a given line (another string).
```
gonuke:
```suggestion
    a given line (another string), without consideration for the order of those words.
```

scripts/activation_responses.py
gonuke: Either fix the comment or fix the template.

pyne/utils.py
zxkjack123: I remove the function `str_elements_in_line()` (in https://github.com/pyne/pyne/pull/1123/commits/c11ab7711f1a0e019738284386e53e76c7f2b066) because this function is only used for matching strings when read ALARA output. It is fragile and may cause errors if someone uses it to other places.

For matching strings of specific lines for ALARA, a simple `some_string in line` is enough.

scripts/activation_responses.py
zxkjack123: Good catch. Fixed in https://github.com/pyne/pyne/pull/1123/commits/6ec025758f16bde841055196cfde03479baa7d5e.

gonuke (approved):

pyne/alara.py
```diff
@@ -198,6 +205,110 @@ def photon_source_to_hdf5(filename, nucs='all', chunkshape=(10000,)):
...
+    response : str
+        Key word of the response
+    chunkshape : tuple of int
+        A 1D tuple of the HDF5 chunkshape.
+    """
+    f = open(filename, 'r')
+    f.seek(0)
+
+    dt = np.dtype([
+        ('idx', np.int64),
```
gonuke: maybe this should also be `mesh_idx`

zxkjack123: @gonuke About this part, I think we need to keep it as `idx`. The `idx` can be used in both voxel (`idx == mesh_idx`) and sub-voxel R2S (`idx != mesh_idx`), and also the activation responses (`idx == mesh_idx`).

As we are going to merge them together, it's better to keep this `idx`.

pyne/alara.py
```diff
@@ -142,7 +149,7 @@ def photon_source_to_hdf5(filename, nucs='all', chunkshape=(10000,)):
     f.seek(0)
     G = len(header) - 2
 
-    dt = np.dtype([
+    dtype = np.dtype([
```
gonuke: Do you think we should use a different name: `response_dtype` or something?  Just to avoid confusion with the built in name `dtype`

pyne/alara.py
```diff
@@ -198,6 +205,108 @@ def photon_source_to_hdf5(filename, nucs='all', chunkshape=(10000,)):
...
+        The path to the file output.txt
+    response : str
+        Key word of the response
+    chunkshape : tuple of int
+        A 1D tuple of the HDF5 chunkshape.
+    """
+    f = open(filename, 'r')
+    f.seek(0)
+
+    dtype = np.dtype([
```
gonuke: Maybe this should be defined once at global scope to make sure it doesn’t get out of sync..

zxkjack123: No, we can't. There is a variable `response` in the keys.

gonuke: I just PR'ed a solution to this

gonuke:
Thanks @zxkjack123 - this is a great addition!

Issue #1124, opened 2019-04-25
Add 'hdf5-tools` in prerequisites for better h5 file comparison in tests


zxkjack123:
`h5diff` is a convenient tool to compare two h5 files.
In some of the nose test functions, h5 files are written, however, we compared the contents not by directly comparing two h5 files, but via other ways such as reading the data out and comparing the data.
As addressed in #1115 , it's useful to compare two h5 files. There is already a tool named `h5diff` to do this task. The `h5diff` is in `hdf5-tools` (not yet in the dependencies).
Therefore, I suggest to add `hdf5-tools` as a prerequisite of pyne.

kkiesling:
I am in favor of using this tool to test files, but it does add an additional dependency only for testing. Curious to know what others in the PyNE community think about adding this as a dependency?

bam241:
maybe we can make it an optional dependency, and having the test checking if `h5diff` are present.

so we can rely on it for the `CI` without forcing users to have it ?

zxkjack123:
Maybe we can just install `h5diff` in CI, but not add it as a dependency for pyne. Therefore:
- For CI, install `h5diff`, test with `h5diff`
- For user test pyne, if `h5diff` not exist, skip the test
- For user test pyne, if `h5diff` exist, do the test

zxkjack123:
Solved in #1128 . Close this issue.

PR #1127, opened 2019-05-04
add hdf5-tools in docker images for circleci

zxkjack123:
As addressed in issue #1124, this PR adds the `hdf5-tools` (contains `h5diff`) in the docker images for CircleCI. Therefore, we can test codes in CircleCI with `h5diff` without request users to install `hdf5-tools`.


zxkjack123:
Updated in PR #1128, close this one.

PR #1128, opened 2019-05-06
Update docker for circleci

zxkjack123:
Update the docker file for circleci:
- add `hdf5-tools` for better nose test
- add 'future' for python2/3 compatibility

### DIFF
--- a/news/update_docker_for_circleci.rst
+++ b/news/update_docker_for_circleci.rst
@@ -1,6 +1,7 @@
 **Added:**
 
-* Add hdf5-tools and future as dependency for docker images used in CircleCI.
+* Add hdf5-tools as dependency for docker images used in CircleCI, for better nose test comparing h5 files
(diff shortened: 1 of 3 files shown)


news/update_docker_for_circleci.rst
```diff
@@ -0,0 +1,13 @@
+**Added:**
+
+* Add hdf5-tools and future as dependency for docker images used in CircleCI.
```
bam241: Could you just briefly explain why you need to add `future` as a python dependency ? (basically what you said on slack...)

bam241:
LGTM ! (with a small comment on future dependency addition...)

news/update_docker_for_circleci.rst
zxkjack123: I updated the reason of adding these two dependencies in news file.

PR #1130, opened 2019-05-06
sikp nonsense cell in source sampling

zxkjack123:
This PR is a variation of PR #1100 , it changes the method of treating `nonsense cell` in source sampling. Different from #1100 , this PR prints out warning messages and skip the bad particle. This warning message could be used to help understand when and why this is happening, including the particle history number and x,y,z coordinates of the point in question.

### DIFF
--- a/share/source.F90
+++ b/share/source.F90
@@ -31,16 +31,16 @@ function find_cell() result(icl_tmp)
       if (j .eq. 0) then
          ! valid cel set
          icl_tmp = i
+         ! icl now set to be valid cell
          exit
       endif
     enddo
-    ! icl is now set
 
+    ! icl now is -1
     if(icl_tmp .le. 0) then
-      write(*,*) 'Nonsense cell number stopping'
-      stop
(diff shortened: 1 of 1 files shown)


share/source.F90
```diff
@@ -31,16 +31,16 @@ function find_cell() result(icl_tmp)
...
          exit
       endif
     enddo
-    ! icl is now set
 
+    ! icl now is -1
     if(icl_tmp .le. 0) then
-      write(*,*) 'Nonsense cell number stopping'
-      stop
+      write(*,*) 'history ', nps, 'at position ', xxx, yyy, zzz, ' no in any cell'
```
ljacobson64: "no" should be "not"

ljacobson64:
This looks good to me... any comments @gonuke @makeclean ? Also not sure the best way we could test this. @bohmt needs this fix for some JET SDDR work

gonuke:
LGTM - Thanks @zxkjack123

PR #1133, opened 2019-05-14
fix error in mesh create from h5m file

zxkjack123:
This PR fixes a problem in unstructured R2S. One of the PRs breaks the unstructured R2S step2.
The reason is that there is the `/materials` in the `blank_mesh.h5m` generated in r2s step1.
In the old version of unstructured R2S (about Jan/2019), there isn't a `/materials` in `blank_mesh.h5m`. (h5ls blank_mesh.h5m).
However, after some PR, there is a `/materials` in `blank_mesh.h5m`. I don't know which PR brings this change. And this change breaks the unstructured R2S not because it adds the `/materials` to `blank_mesh.h5m`.

The deeper reason is, the default `/materials` information of the `blank_mesh.h5m`, the material composition is something like this: (-1, -1, -1, {}), representing (mass, density, atoms_per_molecule, fractions of nuclides). While, after the process of wring into `h5m` files and then read it back, the default value become: (-1, -1, -1), the empty dict is gone away, leads to an error when access to `row[3]`.

Major changes:
- [x] : Fix the problem in creating mesh from h5m file.
- [x] : Add test functions that use the `blank_mesh.h5m` from step1
- [x] : Change the example files of unstructured R2S to simpler one (less volume elements) to speed up test and reduce the test file size.


gonuke:
@bam241 is this the same problem you are seeing?

zxkjack123:
@gonuke No, they are actually different problems. The problem of @bam241 's PR about material API is still there.

gonuke:
FWIW - I think this all emerged because of the code around [line 824 of mesh.py](https://github.com/pyne/pyne/blob/33c7a04df0195ad4b725f83020bae4bc65d442a2/pyne/mesh.py#L824-L831)

gonuke:
This PR doesn't solve the cause, but it does solve a symptom that make occur for other reasons in the future.

Thanks @zxkjack123

Issue #1138, opened 2019-05-24
source.F90 should use "implicit none"


ljacobson64:
source.F90 currently uses `implicit real`, which automatically assigns the `real` type to variables based on the first letter of the variable name. This is not good modern code practice, and can lead to a lot of issues.

For example, recently DAGMC was updated to have the newer source sampling routines present in pyne, in which some function signatures changed, but the source.F90 was not updated. The compiler was perfectly happy to compile DAGMC despite the incompatibility between the source sampling routines and source.F90. The error was not noticed until trying to run a problem using the source sampling routines, at which point a runtime error occurred.

`implicit none` means that the type of every variable must be declared explicitly. If `implicit none` were used, the aforementioned problem with DAGMC would have triggered a compiler error, and it would have been fixed before being merged into DAGMC.

I will submit a PR that fixes this soon.

gonuke:
Fixed in #1139

PR #1139, opened 2019-05-24
Use implicit none in source.F90 and add MCNP6 version

ljacobson64:
This PR:

* changes source.F90 to use `implicit none` instead of `implicit real` for the reasons described in #1138 
* tidies the formatting of source.F90 (consistent indentation, removal of trailing spaces, etc.)
* adds an MCNP6 version, called source_mcnp6.F90

This PR should not change the functionality of source.F90 at all. I tested this version with the JET shutdown photon source and it works the same as before.

### DIFF
--- a/share/source.F90
+++ b/share/source.F90
@@ -9,7 +9,7 @@
 ! (e.g. most Cartesean meshes). This version of find_cell does not work for
 ! repeated geometries or universes.
 !
-! Full instructions on compiling and using MCNP6 with this subroutine are found
+! Full instructions on compiling and using MCNP5 with this subroutine are found
 ! in the PyNE user manual.
 
 function find_cell(cell_list, cell_list_size) result(icl_tmp)
--- a/share/source.F90
+++ b/share/source.F90
@@ -1,15 +1,15 @@
 ! This is Fortran90 code that can be compiled directly into MCNP5 in order
 ! to use the mesh-based sampling capabilities provided by the Sampler class
-! within source_sampling.cpp. The subroutine "source" calls the MCNP5 interface 
+! within source_sampling.cpp. The subroutine "source" calls the MCNP5 interface
 ! C++ functions within source_sampling.cpp. The function "find_cell"
 ! determines what geomety cell a sampled x, y, z are in, which allows for
-! void rejection within the source subroutine. Void rejection ensures that 
+! void rejection within the source subroutine. Void rejection ensures that
 ! particles are never born in void (which is non-physical) which may arise in
 ! the case of source density meshes that are non-conformal to the geometry
 ! (e.g. most Cartesean meshes). This version of find_cell does not work for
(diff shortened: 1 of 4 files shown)


share/source.F90
```diff
@@ -1,15 +1,15 @@
...
-! void rejection within the source subroutine. Void rejection ensures that 
+! void rejection within the source subroutine. Void rejection ensures that
 ! particles are never born in void (which is non-physical) which may arise in
 ! the case of source density meshes that are non-conformal to the geometry
 ! (e.g. most Cartesean meshes). This version of find_cell does not work for
-! repeated geometries or universes. 
+! repeated geometries or universes.
 !
-! Full instructions on compiling and using MCNP5 with this subroutine are found
+! Full instructions on compiling and using MCNP6 with this subroutine are found
```
zxkjack123: Isn't it the source.F90 file for MCNP5?

share/source.F90
```diff
@@ -42,128 +42,146 @@ function find_cell(cell_list, cell_list_size) result(icl_tmp)
 !             complex geometry. The results is suspicious under this condition.
 !           Skip it and resample next particle with error message.
 
+  use fixcom, only: mxa
+  use mcnp_global, only: ncl
+  use mcnp_params, only: dknd
+  use pblcom, only: xxx, yyy, zzz, erg
```
zxkjack123, line 48: The `xxx`, `yyy`, `zzz` and `erg` are also defined below. Are they different?

share/source_mcnp6.F90
```diff
@@ -0,0 +1,182 @@
+! This is Fortran90 code that can be compiled directly into MCNP5 in order
```
zxkjack123:
```suggestion
! This is Fortran90 code that can be compiled directly into MCNP6 in order
```

share/source_mcnp6.F90
```diff
@@ -0,0 +1,182 @@
...
+! to use the mesh-based sampling capabilities provided by the Sampler class
+! within source_sampling.cpp. The subroutine "source" calls the MCNP5 interface
```
zxkjack123:
```suggestion
! within source_sampling.cpp. The subroutine "source" calls the MCNP6 interface
```

share/source.F90
ljacobson64: Not sure what you mean. This line allows the `find_cell()` function to access those variables. The `source` subroutine doesn't automatically have access to the same variables as `find_cell()`, so they must be imported there too.

zxkjack123: Got it.

ljacobson64:
Any more comments on this? I want to get this merged so I can update DAGMC.

zxkjack123:
I think this PR is ready to merge. I will merge it if there isn't any more comments by about 10 pm today.

PR #1146, opened 2019-06-13
DagOpenMC R2S

zxkjack123:
This PR provides an interface between OpenMC and PyNE R2S, reads the neutron flux from OpenMC and write the photon source information to `h5m` files by adding the following functions:
- Move the `MeshTally` class from `pyne.mcnp` to `pyne.mesh`
- Reads the OpenMC state point file (containing mesh tally) and creates a `MeshTally` instance for it
- Setup R2S using data from OpenMC state point file

Photon source sampling changes for OpenMC is NOT included in this PR.


gonuke:
I think we should discuss the architecture/design of `MeshTally` for reading multiple formats.  There is a long-standing inconsistency in PyNE regarding where to put code-specific versions of things.

- option A: put all methods specific for code C in C.py, with different methods for reading each PyNE object type (e.g. material, tally, etc).  Adding support for a new code is primarily an issue of changing that file; adding support for a new object requires changing many files.

- option B: put all the methods specific for object O in O.py, with different methods for reading/writing each code. Adding support for a new code requires changing many files; adding support for a new object requires changing a single file.

I think the tendency is top option B.  Perhaps the constructor for `MeshTally` needs to be altered not read anything, but then there can be a `from_mcnp` method and a `from_openmc` method?

zxkjack123:
@gonuke I think option B is better than option A too. And I totally agree that the constructor in `MeshTally` need not read anything. What's more, I think we can do even better based on option B by combine some features of option A, then the `from_mcnp` and `from_openmc` do not need to read anything too. 

Put all the methods specific for an object O in O.py, but put methods for reading/writing each code in C.py. Taking reading tally data from OpenMC into `MeshTally` as an example. We want to read tally data from OpenMC state point file format. We need these data:
- tally id
- energ bins
- mesh boundaries
- flux data
- flux relative error
- ...

We add method `from_openmc` in `MeshTally` in `mesh.py`. But we do not put the code reading data in `mesh.py`, alternatively, we put it in `openmc.py`. Here is the structure of `MeshTally.from_openmc`:
- `tally_id = openmc.get_tally_id()`
- `e_bins = openmc.get_tally_e_bins()`
- `mesh_bounds = openmc.get_mesh_bounds()`
- `flux_data, flux_rel_err = openmc.get_flux_data_err()`
- ...
- `MeshTally.constructor()`

The tasks of the functions in the `C.py` are reading the specific data. And the method in `O.py` constructs the object using the data received from these functions in `C.py.` By this way:
- Adding support for new code requires changes in both `C.py` and `O.py` two files, but the changes in `C.py` and `O.py` are relatively independent.
- Adding support for new object, requires changes in `O.py`.

zxkjack123:
Update:
The relationship of object `MeshTally` and code `OpenMC`/`MCNP` is changed. The current structure:
- The constructor of `MeshTally` will only set several default values.
- The file reading part are put in `openmc.py` and `mcnp.py`.
- In `openmc.py` and `mcnp.py`, there are functions called `create_meshtally()`. This funtion will read data, initialize `MeshTally` object and assign vlues to variables.

pyne/openmc.py
```diff
@@ -156,3 +165,241 @@ def xml(self):
         ace_tables = "\n  ".join([a.xml() for a in self.ace_tables])
         s = template.format(filetype=self.filetype, ace_tables=ace_tables)
         return s
+
+def get_ebins_from_openmc_sp(filename, tally_num):
```
makeclean, line 169: Why duplicate the OpenMC python interface? Why not just use it?

zxkjack123: This PR relates to read and write hdf5 files. I do not use the OpenMC python interface because:
- HDF5 dependency issue:
    - HDF5 related packages: PyNE uses `pytables` to deal with h5 files, but OpenMC uses `h5py`. Use OpenMC python interface means we need add `h5py` as another dependency.
    - HDF5 file interface: Both PyNE and OpenMC have their own h5 interface. PyNE has `src/h5wrap`. OpenMC has `src/hdf5_interface`. They use different  interfaces.
 - Code dependency: PyNE provides interfaces for different codes by reading their output and writing their input files. But PyNE does not depend on them, i.e., we don't need to install OpenMC if we want to use PyNE OpenMC interface. what we need is the output files of OpenMC. We can install OpenMC and PyNE on different machines.

zxkjack123: @makeclean Now  I totally agree with you to use the OpenMC python API. I need more and more functions related to OpenMC, it's stupid to do it again. I opened an issue here: #1166 .

gonuke:
Will you close this PR then @zxkjack123? Or will you just continue to amend it to move towards the OpenMC API?

zxkjack123:
@gonuke Closing it is a better choice. Because the changes of OpenMC API have conflicts with current PR.

Issue #1145, opened 2019-06-13
Some metastable isotopes have wrong decay heats


aruzzo:
After looking more at #1054, there are more metastable isotopes that have a very large decay heat value when used in pyne.material.Material (28 in total that I found): [https://gist.github.com/aruzzo/4dbf3c283eeed8b712e16c150bd19a64](url)
I edited the material.cpp file to print what values the `decay_heat()` function uses. I tested Ag-108m, one of the original problem isotopes from the issue:
```
########OUTPUT FROM material.cpp########
decay constant: 5.77623e+08
q value: 1.636
decay heat: 8.44974e+11
########END FROM material.cpp########
{471080001: 844974163698.3318}
########OUTPUT FROM material.cpp########
decay constant: 5.01473e-11
q value: 0
decay heat: 0
########END FROM material.cpp########
{471080002: 0.0}
```
For id `471080001`, the function retrieves the right q value from the nuc_data.h5 file but the wrong half-life/decay constant. Using id `471080002` gets the right decay constant but a 0 q value (no entry exists in the decay/q_values tree so it defaults to 0). It seems to pull the half-life from the decay/level_list tree, which contains all the different energy levels of the isotopes. For the problematic isotopes, an energy level of 1 exists before the metastable level with an extremely short half-life; the metastable isotopes are the second energy level, which is why using `471080002` retrieves the correct half-life and decay constant.
Also, when parsing through the different ids in the decay/q_values tree, there was one id `162500000` which seems wrong - I don't think an isotope of sulfur exists with a mass of ~250 amu.

katyhuff:
What reference are you using to define "right" ?

aruzzo:
I don't know of another resource to check the q values of the reactions, I assumed they were right. I calculated a few of the decay constants using the half-lives available in the decay/decays tree, which are the same values as in decay/level_list but easier to search through since it has much fewer entries. They were the same values that were used by material.cpp.

katyhuff:
Sorry -- still confused.

You say:
> the function retrieves the right q value 

How do you know?  What is the right q value? Where did you get it from?

> but the wrong half-life/decay constant

I assume you're looking up the decay constant on nndc.bnl.gov and it's different there than is given by the decay_constant function? Please recall that $t_{1/2} = ln(2)/\lambda$

> I don't know of another resource to check the q values of the reactions, I assumed they were right.

What do you mean by "they" in this sentence?

aruzzo:
[...]

PR #1147, opened 2019-06-18
Fix decay_heat() in material.cpp

aruzzo:
This PR addresses #1145:

- `decay_heat()` in material.cpp now calls `metastable_id()` before calling `decay_const()`

`decay_const()` searches for the the half-life using state_id notation, but `decay_heat()` formerly called it without first using `metastable_id()` to convert from zas_id to state_id.
I updated the previous test file [found here ](https://gist.github.com/aruzzo/28cbe1e861436f677d574092c06ad116) with a new one [found here](https://gist.github.com/aruzzo/3e87e6381702b6572b13d31e43429c26) to show that `decay_heat()` still works for two previously normal isotopes and now works for two previously problematic isotopes.

### DIFF
--- a/src/material.cpp
+++ b/src/material.cpp
@@ -1221,8 +1221,8 @@ pyne::comp_map pyne::Material::decay_heat() {
   double masspermole = mass * pyne::N_A;
   for (pyne::comp_iter i = comp.begin(); i != comp.end(); ++i) {
     dh[i->first] = masspermole * (i->second) * \
-                   decay_const(i->first) * q_val(i->first) / \
-                   atomic_mass(i->first) / pyne::MeV_per_MJ;
+                   decay_const(metastable_id(i->first,i->first % 10000)) * \
+                   q_val(i->first) / atomic_mass(i->first) / pyne::MeV_per_MJ;
   }
   return dh;
 }
--- a/src/material.cpp
+++ b/src/material.cpp
@@ -1221,7 +1221,7 @@ pyne::comp_map pyne::Material::decay_heat() {
   double masspermole = mass * pyne::N_A;
   for (pyne::comp_iter i = comp.begin(); i != comp.end(); ++i) {
     dh[i->first] = masspermole * (i->second) * \
-                   decay_const(metastable_id(i->first,i->first % 10000)) * \
+                   decay_const(metastable_id(i->first,nucname::snum(i->first))) * \
                    q_val(i->first) / atomic_mass(i->first) / pyne::MeV_per_MJ;
   }
   return dh;
--- a/tests/test_material.py
+++ b/tests/test_material.py
@@ -222,10 +222,19 @@ class TestMaterialMethods(TestCase):
         assert_equal(set(obs.values()), set(exp.values()))
 
 
-    def test_decay_heat(self):
+    def test_decay_heat1(self):
         mat = Material({922350000: 0.05, 922380000: 0.95}, 15)
         obs = mat.decay_heat()
         exp = {922350000: 4.48963565256e-14, 922380000: 1.2123912039e-13}
+        assert_equal(set(obs), set(exp))
+        for key in exp:
+            assert_almost_equal(obs[key], exp[key])
+    
+
+    def test_decay_heat2(self):
+        mat = Material({471080001: 0.5, 551340001: 0.5}, 2)
+        obs = mat.decay_heat()
+        exp = {471080001: 7.33578506845e-8, 551340001: 6.241109861949e-3}
         assert_equal(set(obs), set(exp))
         for key in exp:
             assert_almost_equal(obs[key], exp[key])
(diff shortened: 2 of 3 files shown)


src/material.cpp
```diff
@@ -1221,8 +1221,8 @@ pyne::comp_map pyne::Material::decay_heat() {
   double masspermole = mass * pyne::N_A;
   for (pyne::comp_iter i = comp.begin(); i != comp.end(); ++i) {
     dh[i->first] = masspermole * (i->second) * \
-                   decay_const(i->first) * q_val(i->first) / \
-                   atomic_mass(i->first) / pyne::MeV_per_MJ;
+                   decay_const(metastable_id(i->first,i->first % 10000)) * \
```
bam241: maybe use `nucname::snum(int nuc)` to get the metastable_id ? it would avoid using magic number...

bam241 (changes requested):
Thx @aruzzo for fixing this.

Could you add a test in `tests/test_materials.py`:`test_decay_heat(self)` to ensure this is being tested and working properly in the future ?

tests/test_material.py
```diff
@@ -222,13 +222,22 @@ def test_activity(self):
...
+    def test_decay_heat1(self):
         mat = Material({922350000: 0.05, 922380000: 0.95}, 15)
         obs = mat.decay_heat()
         exp = {922350000: 4.48963565256e-14, 922380000: 1.2123912039e-13}
         assert_equal(set(obs), set(exp))
         for key in exp:
             assert_almost_equal(obs[key], exp[key])
+    
+
+    def test_decay_heat2(self):
```
bam241: `test_decay_heat_metastable` ?

tests/test_material.py
```diff
@@ -222,13 +222,22 @@ def test_activity(self):
         assert_equal(set(obs.values()), set(exp.values()))
 
 
-    def test_decay_heat(self):
+    def test_decay_heat1(self):
```
bam241: maybe `test_decay_heat_stable` ?

bam241 (commented):
one minor remaining change, LGTM otherwise !

bam241 (approved):

tests/test_material.py
```diff
@@ -222,13 +222,22 @@ def test_activity(self):
...
         exp = {922350000: 4.48963565256e-14, 922380000: 1.2123912039e-13}
         assert_equal(set(obs), set(exp))
         for key in exp:
             assert_almost_equal(obs[key], exp[key])
+    
+
+    def test_decay_heat_metastable(self):
+        mat = Material({471080001: 0.5, 551340001: 0.5}, 2)
+        obs = mat.decay_heat()
+        exp = {471080001: 7.33578506845e-8, 551340001: 6.241109861949e-3}
```
bam241, line 239: Just one quick though, how did you get those number ?
did you run a calculation using the fixed Pyne code ?
If so it might be good to have an other way to check it ? (not sure it is easy to find..)
in any case, I think you should put a comment to explain how those numbers where determined.

aruzzo: I copied the decay heat values from the jupyter notebook that I created to test some isotopes: [test_decay_heat_fix](https://gist.github.com/aruzzo/3e87e6381702b6572b13d31e43429c26). In short, I copied Avogadro's number and the value of MeV per MJ from pyne's data.cpp file and I rewrote the equation for calculating decay heat in python following the same steps in pyne's material.cpp file. For the atomic data, I got the info from outside sources (which should also be the same sources that pyne uses) in order to separate the calculation from pyne code. The half-lives were from [BNL](https://www.nndc.bnl.gov/nudat2/chartNuc.jsp), the q-values were from [ORNL](https://technicalreports.ornl.gov/cppr/y2001/rpt/97914.pdf), and the atomic masses were from [AMDC](https://www-nds.iaea.org/amdc/ame2016/mass16.txt).

bam241: great ! could you add a small comment to explain that ?
something like:
`// external calculation using Half-live: BNL, q-values: ORNL, Atomic mass: AMDC`

src/material.cpp
```diff
@@ -1221,8 +1221,8 @@ pyne::comp_map pyne::Material::decay_heat() {
   double masspermole = mass * pyne::N_A;
   for (pyne::comp_iter i = comp.begin(); i != comp.end(); ++i) {
     dh[i->first] = masspermole * (i->second) * \
-                   decay_const(i->first) * q_val(i->first) / \
-                   atomic_mass(i->first) / pyne::MeV_per_MJ;
+                   decay_const(metastable_id(i->first,nucname::snum(i->first))) * \
+                   q_val(i->first) / atomic_mass(i->first) / pyne::MeV_per_MJ;
```
gonuke, line 1225: Shouldn’t `q_val` also depend on the isomeric state?

gonuke (commented):
Thanks for this catch @aruzzo . One question about `q_val`

src/material.cpp
aruzzo: Yeah, `q_val` depends on the isomeric state but the table that holds the different q_values uses zas_id unlike the table for the decay constants which uses state_id

gonuke:
Thanks @aruzzo - sorry for the delays.

PR #1150, opened 2019-07-01
Cell fracs sort

zxkjack123:
This PR changes the default sequence of the `cell_fracs`. Sorting the `cell_fracs` according to the order of ['idx', 'vol_frac`]. Therefore, the cells with larger volume fractions are in front of the cells with smaller cell fractions. By applying this change, the speed of source sampling would be faster.

Take sampling the photon source in a voxel as an example, assuming the cell volume fractions is `[f1, f2, ..., fn]`, where `f1 + f2 + ... + fn = 1` adn `fi >= 0`.
When sampling a source particle, the code will iterate each cell and check whether the source particle located in the chosen cell. The expectation of the iterate/check times is: `1 * f1 + 2 * f2 + ... + n * fn`. It is the dot prodct of two series: `[1, 2, ..., n]` and `[f1, f2, ..., fn]`. According to the [rearrangement inequality](https://en.wikipedia.org/wiki/Rearrangement_inequality), the minimum value is when the `[f1, f2, ..., fn]` is a decreasing series. Therefore,  sort the `cell_fracs` by decreasing order of `vol_frac` could reduce the check time of source sampling.

Major changes:
- Sort the `cell_fracs` by the order of ascending `idx` and descending `vol_frac`.
- Add a test to show the sorting results.


gonuke:
Is there a test?  The changes in the tests don’t look sorted to me.  Do you want/need to reverse the sort order?

zxkjack123:
@gonuke The original sort order of `cell_fracs` is `[idx, cell_number]`, now the sort order is `[idx, vol_frac]`. The aim of this change is to reduce the iterate/check time in source sampling.

Take sampling the photon source in a voxel as an example, assuming the cell volume fractions is `[f1, f2, ..., fn]`, where `f1 + f2 + ... + fn = 1` adn `fi >= 0`.
When sampling a source particle, the code will iterate each cell and check whether the source particle located in the chosen cell. The expectation of the iterate/check times is: `1 * f1 + 2 * f2 + ... + n * fn`. It is the dot prodct of two series: `[1, 2, ..., n]` and `[f1, f2, ..., fn]`. According to the [rearrangement inequality](https://en.wikipedia.org/wiki/Rearrangement_inequality), the minimum value is when the `[f1, f2, ..., fn]` is a decreasing series. Therefore,  sort the `cell_fracs` by decreasing order of `vol_frac` could reduce the check time of source sampling.

zxkjack123:
Working on the test of sorting.

zxkjack123:
Added a test to demonstrate the sort.

pyne/mesh.py
```diff
@@ -1658,3 +1660,46 @@ def step(self, num_steps):
...
+    num_vol_elements: int
+        number of volume elements.
+
+    Returns
+    -------
+    cell_fracs : structured array
+        Sorted cell_fracs.
+    """
+    # sort ascending along idx and vol_frac
+    cell_fracs.sort(order=['idx', 'vol_frac'])
```
pshriwise, line 1695: Another option here might be to temporarily multiply the volume fractions by `-1.0` to simplify this function and still get an effective reversed sort on the volume fracs. This way the entire body of this function would be:

```
cell_fracs['vol_frac'] *= -1.0
cell_fracs.sort(order=['idx', 'vol_frac'])
cell_fracs['vol_frac'] *= -1.0
```

There also might be some specialized method to sort a single field in reverse, but this should work fine for this case I think.

pshriwise (changes requested):
Just one small comment on making the sorting function a little cleaner.

pyne/mesh.py
zxkjack123: Good idea. Suggestion adopted in d3975ab.

pyne/mesh.py
```diff
@@ -1395,8 +1395,10 @@ def tag_cell_fracs(self, cell_fracs):
             cell changing fastest.
 
         """
-
         num_vol_elements = len(self)
+        # sort cell_fracs
+        cell_fracs = _cell_fracs_sort_vol_frac_reverse(cell_fracs, num_vol_elements)
+        #cell_fracs.sort(order=['idx', 'vol_frac'])
```
pshriwise: This commented line can probably go, yeah?

pyne/mesh.py
pshriwise: A few comments about why the values are being modified and replaced here might be nice.

pyne/mesh.py
zxkjack123: Yes, removed this comment in 5abacf8.

gonuke:
Thanks @zxkjack123 - this looks like a good change

gonuke:
It seems like you addressed @pshriwise request - so I'll merge

PR #1152, opened 2019-07-18
redundancies in tally

bam241:
While looking at the `tally.cpp` they were a lot of copy pasted code...

this fixes it.


zxkjack123 (approved):
LGTM.

gonuke:
This looks like a good simplification.  I have a quibble with the original code, but won't hold this up.

I'm not sure why the normalization is being tested against 1.0???  I'll make an issue.

PR #1153, opened 2019-07-19
Mesh tally

bam241:
This PR aims to bring `MeshTally` management in the `PyNE::Tally` class.

It only does allow to write the tally in MCNP format for the moment.
I think we should try to add the hdf5 read/write.

Python binding are missing yet this is not exposed on the Python API yet.

this is not completely over. please feel free to comment if you have any suggestion.


src/tally.cpp
```diff
@@ -57,75 +61,98 @@ pyne::Tally::Tally(std::string type, std::string part_name,
...
-			    int row) { 
+void pyne::Tally::from_hdf5(std::string filename, std::string datapath,
+                            int row) {
   // line of data to acces
   int data_row = row;
 
   // check for file existence
-  if (!pyne::file_exists(filename))
-    throw pyne::FileNotFound(filename);
+  if (!pyne::file_exists(filename)) throw pyne::FileNotFound(filename);
```
gonuke: Does the style guide indicate a preference here?

gonuke (changes requested):
There are a lot of whitespace changes here.  In one case, it looks like your editor set the indent to be only a single space - seems too little - does the style guide indicate the write indentation?

If there will be whitespace changes (and there may well be), then please make a whitespace only PR for affected files (or all files?) and then resubmit this as an algorithmic change.

bam241:
I restored the previous formatting to what it was, or close... (I did not restored the tabs... and kept  the pointer stars at a place I like better :) )
@gonuke this should be good for a first review.

src/tally.cpp
```diff
@@ -440,14 +466,78 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
     } else if (tally_type.find("Current") != std::string::npos) {
       // makes no sense in mcnp
     }
+  
+  
+  } else if (entity_type.find("Mesh") != std::string::npos) {
+    output << "FMESH4:" << particle_token << "  ";
```
gonuke: We should use a tally index - like in the other tallies - to allow multiple mesh tallies with different numbers.

src/tally.cpp
```diff
@@ -59,17 +58,44 @@ pyne::Tally::Tally(std::string type, std::string part_name,
...
+  entity_geometry = ent_geom;
+  tally_name = tal_name;
+  entity_size = -1;
+  
+  for (int i=0; i<3; i++)  {
+    origin[i] = orgn[i];
+    vec[i] = vec_[i];
+    axl[i] = axl_[i];
+  }
+  i_meshs = mesh_i;
```
gonuke: why not store these in a vector - you end up using them that way eventually

src/tally.cpp
```diff
@@ -440,14 +466,78 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
+    output << "GEOM=";
+    std::stringstream sup_var;
+
+    if (entity_geometry.find("XYZ") != std::string::npos) {
+      output << "XYZ ";
+    } else if (entity_geometry.find("Cylinder") != std::string::npos) {
+      output << "CYL ";
+      if (axl[0] != 0 || axl[1] != 0 || axl[2] != 0) {
+        sup_var << "AXL=";
+        for (int i = 0; i < 3; i++) {
```
gonuke: Seems like it would be useful to write a method to write vectors as "a b c" triplets.

src/tally.cpp
```diff
@@ -440,14 +466,78 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
+    output << " ORIGIN= " << origin[0] << " " << origin[1] << " " << origin[2]
+        << "\n";
+    std::string dir_name[3] = {"I", "J", "K"};
+    std::vector<double> meshes[3] = {i_meshs, j_meshs, k_meshs};
+    std::vector<int> bins[3] = {i_bins, j_bins, k_bins};
+
+    for (int j = 0; j < 3; j++) {
+      output << "          ";
+      output << dir_name[j] << "MESH=";
+      for (int i = 0; i < meshes[j].size(); i++) {
```
gonuke: as above, a method to convert vectors of arbitrary length into a string "a b c d ..."

gonuke: maybe this is essentially the same as above...

pyne/tally.pyx
```diff
@@ -4,6 +4,7 @@
 # Do not modify!!!                             #
 #                                              #
 #                                              #
+# With some edits form BaM241                  #
```
gonuke:
```suggestion
# With some edits from BaM241                  #
```

pyne/tally.pxd
```diff
@@ -4,15 +4,19 @@
 # Do not modify!!!                             #
 #                                              #
 #                                              #
+# With some edits form BaM241                  #
```
gonuke:
```suggestion
# With some edits from BaM241                  #
```

pyne/cpp_tally.pxd
```diff
@@ -4,61 +4,77 @@
 # Do not modify!!!                             #
 #                                              #
 #                                              #
+# With some edits form BaM241                  #
```
gonuke:
```suggestion
# With some edits from BaM241                  #
```

gonuke (changes requested):
The c++ part looks good - not sure I have what it takes to review the Cython....

pyne/cpp_tally.pxd
```diff
@@ -4,61 +4,77 @@
 # Do not modify!!!                             #
```
zxkjack123, line 4: As this file has been modified, why not delete this `Do not modify` statement?

pyne/tally.pyx
```diff
@@ -21,6 +25,44 @@ from pyne.utils import QAWarning
...
 warn(__name__ + " is not yet QA compliant.", QAWarning)
 
+cdef vector[int] to_vector_int(value):
+    cdef vector[int] value_proxy
+    cdef int ivalue
+    cdef int value_size
+    cdef int * value_data
+    # value is a ('vector', 'int32', 0)
+    value_size = len(value)
+    if isinstance(value, np.ndarray) and (<np.ndarray> value).descr.type_num == np.NPY_INT32:
```
zxkjack123: I found that in python3, when a `numpy.int32` data add a default `int`, it becomes `numpy.int64`. I met this problem once when I worked on the python2/3 compatibility. Should we consider the case of `np.NPY_INT64` here incase of the type conversion?

pyne/tally.pyx
```diff
@@ -171,9 +196,95 @@ cdef class Tally:
         tal_name_bytes = tal_name.encode()
         self._inst = new cpp_tally.Tally(std_string(<char *> type_bytes), std_string(<char *> part_name_bytes), <int> ent, std_string(<char *> ent_type_bytes), std_string(<char *> ent_name_bytes), std_string(<char *> tal_name_bytes), <double> size, <double> norm)
 
+    
+    def _tally_tally_2(self, part_name, ent_geom, origin, i, j, k,
+            i_ints, j_ints, k_ints, e, e_ints, tal_name='', norm=1.0,
+            vec=[0,0,0], axl=[0,0,0]):
+        """Tally(self, type, part_name, ent, ent_type, ent_name, tal_name='', size=0.0, norm=1.0)
+         This method was overloaded in the C-based source. To overcome
+        this we ill put the relevant docstring for each version below.
```
zxkjack123:
```suggestion
        this we will put the relevant docstring for each version below.
```

pyne/tally.pyx
```diff
@@ -171,9 +196,95 @@ cdef class Tally:
...
+        Each version will begin with a line of # characters.
+
+        ################################################################
+
+        Parameters
+        ----------
+        entity_name : std::string
```
zxkjack123: The parameters listed here are incomplete and inconsistent with the function definition.

pyne/tally.pyx
```diff
@@ -171,9 +196,95 @@ cdef class Tally:
         tal_name_bytes = tal_name.encode()
         self._inst = new cpp_tally.Tally(std_string(<char *> type_bytes), std_string(<char *> part_name_bytes), <int> ent, std_string(<char *> ent_type_bytes), std_string(<char *> ent_name_bytes), std_string(<char *> tal_name_bytes), <double> size, <double> norm)
 
+    
+    def _tally_tally_2(self, part_name, ent_geom, origin, i, j, k,
```
zxkjack123: Are the `i, j, k` here representing the `imesh, jmesh, kmesh` (in MCNP)? If they are, it's better to use `imesh, jmesh, kmesh` instead, as `i, j, k` are often used as iterate arguments.

pyne/tally.pyx
```diff
@@ -171,9 +196,95 @@ cdef class Tally:
...
+        (5, list), (6, list),
+        (7, list), (8, list),
+        (9, list), (10, list),
+        (11, str), (12, float), (13, float),
+        (15, list), (15, list),
+        ("part_name", str), ("ent_geom", str), ("ent_name", str), 
+        ("tal_name", str), ("size", float), ("norm", float),
+        ("origin",list), 
+        ("i", list), ("i_ints", list), 
+        ("j", list), ("ji_ints", list),
```
zxkjack123:
```suggestion
        ("j", list), ("j_ints", list), 
```

src/tally.cpp
```diff
@@ -2,21 +2,21 @@
...
-enum entity_type_enum {VOLUME, SURFACE}; // Enumeration for entity types
-enum tally_type_enum  {FLUX, CURRENT};   // Enumeration for tally types
+enum entity_type_enum { VOLUME, SURFACE, MESH }; // Enumeration for entity types
+enum tally_type_enum { FLUX, CURRENT };  // Enumeration for tally types
 
 const std::string tally_type_enum2string[] = {"Flux", "Current"};
-const std::string entity_type_enum2string[] = {"Volume", "Surface"};
-
+const std::string entity_type_enum2string[] = {"Volume", "Surface", "Mesh"};
+const std::string geometry_type_enum2string[] = {"XYZ", "Cylinder"};
```
zxkjack123: In the MCNP manual, the counterpart of `Cylinder` (or `cylindrical`) is `Cartesian`. The `XYZ` and `RZT` are counterparts. The `REC` and `CYL` are counterparts. Suggest to choose a pair of counterparts here.

src/tally.h
```diff
@@ -105,11 +123,25 @@ namespace pyne
...
+    double origin[3];
+    double vec[3];
+    double axl[3];
+    std::vector<double> i_meshs;
+    std::vector<double> j_meshs;
+    std::vector<double> k_meshs;
+    std::vector<int> i_bins;
+    std::vector<int> j_bins;
+    std::vector<int> k_bins;
+    std::vector<double> energy;    ///< Energy Mesh
```
zxkjack123: When I see `energy`, I always think it is a float number at first. How about change it to `e_bounds` or `energy_bounds` to represent the meaning of energy boundaries?

zxkjack123 (commented):
I am thinking about the case sensitivity. As MCNP input file is not case sensitive, but python code is. There is several find and compare statements about the key words of MCNP. But I didn't see any the code dealing with the case sensitivity.

src/tally.cpp
bam241: I am not sure which one your talking, the `X_meshs` are already vectors, are you talking about `vec` and `axl`?

gonuke: I mean a vector of vectors.  Later on, you do that to loop through the different dimensions, so I thought it would make sense to do that now

src/tally.cpp
```diff
@@ -420,34 +448,103 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
-                                entity_id, entity_size, normalization);
+      tally_id = 4;
     } else if (tally_type.find("Current") != std::string::npos) {
       // makes no sense in mcnp
+      return "";
+    }
+    output << form_mcnp_tally(tally_index, tally_id, particle_token, entity_id,
+                              entity_size, normalization);
+  } else if (entity_type.find("Mesh") != std::string::npos) {
+    output << "FMESH" << tally_index << "4:" << particle_token << "  ";
```
gonuke: What if this was in a method called `form_mcnp_mesh_tally()` to be parallel with the normal tallies?

src/tally.cpp
```diff
@@ -420,34 +448,103 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
+template <typename T>
+bool pyne::Tally::is_zero(T vect) {
+  int size = sizeof(vect) / sizeof(vect[0]);
+  bool result = true;
+  for (int i = 0; i < size; i++) result &= (vect[i] == 0);
+  return result;
+}
+
+template<typename T>
+std::string pyne::Tally::to_string(std::vector<T> vect){
```
gonuke:
```suggestion
std::string pyne::Tally::mcnp_list_to_string(std::vector<T> vect){
```

src/tally.cpp
```diff
@@ -420,34 +448,103 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
+  std::stringstream out;
+  out << std::setiosflags(std::ios::fixed) << std::setprecision(6);
+  if (normalization > 1.0)
```
gonuke: Why this?

pyne/tally.pyx
```diff
@@ -122,58 +147,140 @@ cdef class Tally:
...
-        entity : int
-
         normalization : double
-
         particle_name : std::string
-
         type : std::string
-
         tally_name : std::string
+        i: vector
```
gonuke: This parameter list doesn't seem to match the function

src/tally.h
```diff
@@ -86,13 +104,17 @@ namespace pyne
     void write_hdf5(std::string filename, std::string datapath);
 
     // mcnp tally
-    std::string mcnp(int tally_index = 1, std::string mcnp_version = "mcnp5" );
-  
+    std::string mcnp(int tally_index = 1, std::string mcnp_version = "mcnp5",
```
gonuke, line 117: why isn't `out` part of the tally class member variables?

gonuke (changes requested):

src/tally.h
bam241: I was thinking user would provide it when writing the tally... so it provide more freedom
maybe it is'nt a good idea....

src/tally.cpp
bam241: I don't know, wanted to match what was previously done with the writing precision

src/tally.cpp
```diff
@@ -420,34 +448,103 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
+template <typename T>
+bool pyne::Tally::is_zero(T vect) {
+  int size = sizeof(vect) / sizeof(vect[0]);
+  bool result = true;
+  for (int i = 0; i < size; i++) result &= (vect[i] == 0);
+  return result;
+}
+
+template<typename T>
+std::string pyne::Tally::mcnp_list_to_string(std::vector<T> vect){
```
bam241: I am wondering is this specific to mcnp ? or could it is used in other cases ?

Shall it go to utils ?

bam241: cc @gonuke

src/tally.cpp
```diff
@@ -420,34 +448,103 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
+  std::stringstream out;
+  out << std::setiosflags(std::ios::fixed) << std::setprecision(6);
+  if (normalization > 1.0) 
+    out << std::scientific;
+  
+  for( auto elmt : vect)
```
makeclean: the auto keyword should be banned!

pshriwise: Like, from PyNE? Why??

makeclean:
Laziness, leads to confusion, perhaps I paint my colours here as a 'gluer', I'm usually making n libraries play nice at the same time, auto makes.it really hard to find which header I should be looking in. In fact in this case above isn't it just T?

I'm usually ok with auto if you had some multi-namespaced std:map of a type of a std:: pair

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

src/tally.cpp
pshriwise: Replying to [this comment](https://github.com/pyne/pyne/pull/1153#issuecomment-515799377) - yeah I can see that at times. You can run into some weird implicit type conversion too. I do think it’s valuable in for loops like this for the decreased verbosity and also in those cases like you’ve cited for other relatively verbose types like `std::pair<T,V>`.

src/tally.cpp
gonuke: It could be more general, but the key is that it is space delimited.  If you added an argument that was a string to be the delimiter, then it is likely more general.  You might name it `join_to_string` or something

src/tally.h
gonuke: I guess I see that use case... I'm kind of on the fence....  The user could also just change the class member variable before writing the it

src/tally.cpp
gonuke: Yes - maybe @makeclean can comment on why this, as well?

src/tally.h
bam241: I guess it depends on the canonical definition of a mesh tally, if it includes the "OUT" variable, then yes it should be there, otherwise I guess it should be an argument of the `mcnp()` method....

Does it make sense ?

src/tally.cpp
bam241: I understand @makeclean point, I am working with some other codes with thousands of auto... this is a pain... really!

but I really like the usage of `auto` for loop, it simplifies the syntax (`int i .... foo[i]....` is a bit verbosy for me....) and allows to loop over an array in a python way which I like...
but I don't have a strong opinion... specially in this case...
I'll switch back to the conservative way... :)

src/tally.cpp
bam241: I try to move this in `utils.cpp/utils.h`... but it won't compile (on macOS) if I keep the lib as shared...
@makeclean I remember you help me fix a similar issue on cyclus, to you have any idea how to fix it ?

src/tally.cpp
gonuke: in line 495, you make an array:
```
std::vector<double> meshes[3] = {i_meshs, j_meshs, k_meshs};
```
It is the only place that you use `i_meshs`, etc, so why not declare them initially to just be elements of `meshes[]`

bam241:
@gonuke @pshriwise @zxkjack123 @makeclean this should be ready for a second round of review.

tests/test_tally.py
```diff
@@ -436,6 +436,63 @@ def test_mcnp6_tally_vol_proton_volume_set():
...
+    origin = [1,2,3]
+    i = [5, 10, 20, 25]
+    j = [2, 12]
+    k = [45]
+    i_ints = [1, 2, 3, 1]
+    j_ints = [1]
+    k_ints = [1]
+    e = [0, 10, 100]
+    e_ints = [1, 1, 2]
+    tal_name = "Mesh Tally XYZ Proton"
```
makeclean: for the sake of confusion downstream - maybe change either the title to neutron or the particle to proton

pyne/tally.pyx
```diff
@@ -1,17 +1,20 @@
...
 ################################################
 """
 """
 from pyne cimport stlcontainers
 from libc.stdlib cimport free
 from libcpp.map cimport map as cpp_map
 from libcpp.string cimport string as std_string
+from libcpp.vector cimport vector
+import numpy as np
+cimport numpy as np
```
zxkjack123, line 17: The `numpy` is imported here twice.

pyne/tally.pyx
```diff
@@ -171,9 +174,102 @@ cdef class Tally:
         tal_name_bytes = tal_name.encode()
         self._inst = new cpp_tally.Tally(std_string(<char *> type_bytes), std_string(<char *> part_name_bytes), <int> ent, std_string(<char *> ent_type_bytes), std_string(<char *> ent_name_bytes), std_string(<char *> tal_name_bytes), <double> size, <double> norm)
 
+    
+    def _tally_tally_2(self, part_name, ent_geom, origin, i_mesh, j_mesh, k_mesh,
+            i_ints=[], j_ints=[], k_ints=[], e=[], e_ints=[], vec=[], axl=[], tal_name='', norm=1.0):
+        """Tally(self, type, part_name, ent, ent_type, ent_name, tal_name='', size=0.0, norm=1.0)
```
zxkjack123: The docstring is incompatible with the function parameters.

pyne/tally.pyx
```diff
@@ -171,9 +174,102 @@ cdef class Tally:
...
+        Each version will begin with a line of # characters.
+
+        ################################################################
+
+        Parameters
+        ----------
+        part_name : string
+        ent_geom : string
+        origin : (3d) list 
+        i_mesh : list
```
zxkjack123: Suggests to add the content type of these lists. For example, `i_mesh : list of float`, `i_ints : list of int (default: empty)`.

pyne/tally.pyx
```diff
@@ -603,7 +707,7 @@ cdef class Tally:
             pass
         raise RuntimeError('method from_hdf5() could not be dispatched')
 
-    def mcnp(self, tally_index=1, mcnp_version='mcnp5'):
+    def mcnp(self, tally_index=1, mcnp_version='mcnp5', out=''):
         """mcnp(self, tally_index=1, mcnp_version='mcnp5')
```
zxkjack123: Docstring didn't get update here.

src/tally.cpp
```diff
@@ -2,21 +2,21 @@
...
 
 #ifndef PYNE_IS_AMALGAMATED
-  #include "tally.h"
   #include "particle.h"
+  #include "tally.h"
 #endif
 
-enum entity_type_enum {VOLUME, SURFACE}; // Enumeration for entity types
-enum tally_type_enum  {FLUX, CURRENT};   // Enumeration for tally types
+enum entity_type_enum { VOLUME, SURFACE, MESH }; // Enumeration for entity types
```
zxkjack123: Are the blank spaces necessary?

src/tally.h
```diff
@@ -43,13 +48,31 @@ namespace pyne
...
+          std::string tally_name = "", double entity_size = 0.0, 
+          double normalization = 1.0);
+ 
+    /// Constructor from passed in vars
+    /// \param particle_name the name of the particle type
+    /// \param entity the entity id of the tally (eg. surface index,
+    ///          volume number)
+    /// \param entity_geom (xyz, cylinder)
+    /// \param tally_name string identifying the tally
+    /// \param origin double[3] origin of the meshtally
```
zxkjack123:
```suggestion
    /// \param orgn double[3] origin of the meshtally
```

src/tally.h
```diff
@@ -43,13 +48,31 @@ namespace pyne
...
+    /// \param i,j,k vector<double> coordinate of the coarse meshes
```
zxkjack123:
```suggestion
    /// \param i_mesh, j_mesh, k_mesh vector<double> coordinate of the coarse meshes
```

src/tally.h
```diff
@@ -43,13 +48,31 @@ namespace pyne
...
+    /// \param ints i,j,k vector<double> coordinate of the coarse meshes
```
zxkjack123:
```suggestion
    /// \param ints bins_i, bins_j, bins_k vector<double> coordinate of the coarse meshes
```

src/tally.h
```diff
@@ -43,13 +48,31 @@ namespace pyne
...
+    /// \param normalization the number required to normalize your tally
```
zxkjack123: Some parameters are not show up in the docstring.

tests/test_tally.py
```diff
@@ -436,6 +436,63 @@ def test_mcnp6_tally_vol_proton_volume_set():
                  "F14:h 12\n"+\
                  "SD14 100.000000\n",tally.mcnp(1,"mcnp6"))
 
+def test_mcnp_mesh_tally_xyz():
+    particle = "Neutron"
+    geometry = "Cartesian"
+    origin = [1,2,3]
+    i = [5, 10, 20, 25]
```
zxkjack123: Suggests to use `i_mesh`, `j_mesh` and `k_mesh` to replace `i, j, k` here for better understanding.

pyne/tally.pyx
bam241: one is an `import` the other one a `cimport` used by Cython.

I am not sure why this is used in `material.pyx` and seems to be normal according to [Cython docs](https://cython.readthedocs.io/en/latest/src/userguide/sharing_declarations.html)

pyne/tally.pyx
```diff
@@ -21,6 +24,44 @@ from pyne.utils import QAWarning
...
 warn(__name__ + " is not yet QA compliant.", QAWarning)
 
+cdef vector[int] to_vector_int(value):
+    cdef vector[int] value_proxy
+    cdef int ivalue
+    cdef int value_size
+    cdef int * value_data
+    # value is a ('vector', 'int32', 0)
+    value_size = len(value)
+    if isinstance(value, np.ndarray) and (< np.ndarray > value).descr.type_num == np.NPY_INT64:
```
gonuke: Could this conditional only choose how to set `value_data`, and then the allocation and assignment can be outside the conditional and appear once?
```
if .....
    value_data = <int * > np.PyArray.....
else:
    value_data = value
value_proxy = vector[int]....
```

pyne/tally.pyx
```diff
@@ -21,6 +24,44 @@ from pyne.utils import QAWarning
...
+
+
+cdef vector[double] to_vector_double(value):
+    cdef vector[double] value_proxy
+    cdef int ivalue
+    cdef int value_size
+    cdef double * value_data
+    # value is a ('vector', 'float64', 0)
+    value_size = len(value)
+    if isinstance(value, np.ndarray) and (< np.ndarray > value).descr.type_num == np.NPY_FLOAT64:
```
gonuke: see above

src/tally.h
```diff
@@ -43,13 +48,35 @@ namespace pyne
...
-    double normalization = 1.0);
-
-    ~Tally (); /// default destructor
-
-
+          std::string entity_type, std::string entity_name,
+          std::string tally_name = "", double entity_size = 0.0, 
+          double normalization = 1.0);
+ 
+    /// Constructor from passed in vars
```
gonuke: Better comment to indicate how these two constructors are different: one is for a cell/surface-based tally and the other is for a mesh tally.

src/tally.cpp
```diff
@@ -420,59 +448,131 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
+    std::vector<double> origin, std::vector<double> meshes[3],
+    std::vector<int> ints[3], std::vector<double> e_bounds,
+    std::vector<int> e_ints, std::string out) {
+  std::stringstream mtally_stream;
+  // indentation block
+  std::string indent_block = "           ";
+
+  mtally_stream << "FMESH" << tally_index << "4:" << particle_token << " ";
+  mtally_stream << "GEOM=";
+  std::stringstream sup_var;
```
gonuke: Why this variable? Why not append directly to `mtally_stream`?

src/tally.cpp
```diff
@@ -420,59 +448,131 @@ std::string pyne::Tally::mcnp(int tally_index, std::string mcnp_version) {
...
+
+  if (entity_geometry.find("Cartesian") != std::string::npos) {
+    mtally_stream << "XYZ ";
+  } else if (entity_geometry.find("Cylinder") != std::string::npos) {
+    mtally_stream << "CYL ";
+    if (!is_zero(axs)) {
+      sup_var << indent_block << "AXS=" << pyne::join_to_string(axs) << "\n";
```
gonuke:
```suggestion
      sup_var << indent_block << "AXS=" << pyne::join_to_string(axs) << std::endl;
```
(also in all other locations)

gonuke (changes requested):
A few small changes and I think we'll be ready to merge

src/tally.cpp
bam241: I wanted the tally card to be form in the same order:
GEOM, ORIGIN, MESH, INTS
(it is how it is explained in the MCNP docs, I am not sure if the order matters.)
if the order does not matter, I can append it directly to the `tally_stream`

pyne/tally.pyx
bam241: nice catch !

bam241:
@gonuke This should be good to go, unless you feel like  the `AXS` and `VEC` should directly be appended after the `GEOM`...

src/tally.cpp
gonuke: Order doesn't matter

bam241:
everything should be good now :)

pyne/tally.pyx
gonuke: will you update this like the previous function?

pyne/tally.pyx
```diff
@@ -21,6 +24,40 @@ from pyne.utils import QAWarning
...
+    cdef int ivalue
+    cdef int value_size
+    cdef int * value_data
+    # value is a ('vector', 'int32', 0)
+    value_size = len(value)
+    if isinstance(value, np.ndarray) and (< np.ndarray > value).descr.type_num == np.NPY_INT64:
+        value_data = <int * > np.PyArray_DATA( < np.ndarray > value)
+    value_proxy = vector[int](< size_t > value_size)
+    for ivalue in range(value_size):
+        value_proxy[ivalue] = <int > value[ivalue]
```
gonuke, line 35: This won't work because you don't use `value_data`?

bam241: yes sorry... tests were working so I did not look at it further...
should be fixed in the next commit.

bam241: I am not sure this implementation will work, 
I cannot cast `value` in `value_data` I think that's why the former implementation looked like this...

bam241: I re-implemented the previous implementation (with a slight improvement...)

the problem, is as we don't know the type of `value` we cannot cast it directly into value_data (which is a `cdef double*` or `cdef int*`)

pyne/cpp_tally.pxd
```diff
@@ -4,61 +4,83 @@
...
-from libcpp.string cimport string as std_string
+from libcpp.string cimport string as cstr
+from libcpp.vector cimport vector
 
 cdef extern from "tally.h" namespace "pyne":
-
+    ctypedef double vec3[3]
+    ctypedef vector[double] vdbl
+    ctypedef vector[int] vint
     cdef cppclass Tally:
```
pshriwise, line 22: Let's add a couple of lines here. I'll cite PEP8 even though this is Cython.

pyne/tally.pyx
```diff
@@ -21,6 +24,42 @@ from pyne.utils import QAWarning
...
+            value_proxy[ivalue] = <int > value[ivalue]
+    return value_proxy
+
+
+cdef vector[double] to_vector_double(value):
+    cdef vector[double] value_proxy
+    cdef int ivalue
+    cdef int value_size
+    cdef double * value_data
+    # value is a ('vector', 'float64', 0)
```
pshriwise: Commented line can go

src/tally.cpp
```diff
@@ -59,10 +59,37 @@ pyne::Tally::Tally(std::string type, std::string part_name,
...
+  vec = vec_;
+  axs = axs_;
+  meshes[0] = mesh_i;
+  meshes[1] = mesh_j;
+  meshes[2] = mesh_k;
+  ints[0] = ints_i;
+  ints[1] = ints_j;
+  ints[2] = ints_k;
+  e_bounds = e_bounds_;       ///< Energy Mesh
+  e_ints = e_ints_;  ///< Bin per energy
```
pshriwise: Consistent whitespace here.

src/tally.h
```diff
@@ -22,7 +22,12 @@
 
 namespace pyne
 {
-  class Tally
+  // simplifying implementation...
+  using std::vector;
+  // default empty parameters for constructors
+  const vector<double> null_v_dbl;
+  const vector<int> null_v_int;
+  class Tally
```
pshriwise, line 31: Let's add a line here between these vectors and the `Tally` class.

src/tally.h
```diff
@@ -101,10 +138,21 @@ namespace pyne
...
+    /// fundemental mesh tally variables
+    /// Mesh tally definition variable
+    std::string entity_geometry;
+    std::vector<double> origin;
+    std::vector<double> vec;
+    std::vector<double> axs;
+    std::vector<double> meshes[3];
+    std::vector<int> ints[3];
+    std::vector<double> e_bounds;    ///< Energy Mesh
+    std::vector<int> e_ints;  ///< Bin per energy
```
pshriwise: same as above, consistent whitespace here

tests/test_tally.py
```diff
@@ -436,6 +436,64 @@ def test_mcnp6_tally_vol_proton_volume_set():
...
+    k_mesh = [45]
+    i_ints = [1, 2, 3, 1]
+    j_ints = [1]
+    k_ints = [1]
+    e = [0, 10, 100]
+    e_ints = [1, 1, 2]
+    tal_name = "Mesh Tally XYZ Neutron"
+    vec = [ -1, 4, -2]
+    axs = [ 12, -2, 5]
+
```
pshriwise, line 481: Totally optional, but you can remove one of these blank lines probably.

pshriwise (commented):
some very small comments. Going to try this out today!

pshriwise:
Tried it out and it worked just fine for me!

src/tally.cpp
bam241: removed the comment there, there were the only one in the constructor, and they should be commented in the header.

src/tally.h
bam241: added constituent commenting as well !

bam241 (commented):
> some very small comments. Going to try this out today!

thx for the review !

pyne/tally.pyx
```diff
@@ -21,6 +24,32 @@ from pyne.utils import QAWarning
...
+    cdef vector[double] value_proxy
+    cdef int ivalue
+    cdef int value_size
+    cdef double * value_data
+    
+    value_size = len(value)
+    value_proxy = vector[double](< size_t > value_size)
+    for ivalue in range(value_size):
+        value_proxy[ivalue] = <double > value[ivalue]
+    return value_proxy
```
bam241, line 48: @gonuke this simplified version seems to be working with python3, let's check if it works with python2 as well...

gonuke: Is it being tested over a broad enough range of use cases?  It seems like something we could throw a lot of tests at without much effort.

pyne/tally.pyx
```diff
@@ -603,7 +696,7 @@ cdef class Tally:
             pass
         raise RuntimeError('method from_hdf5() could not be dispatched')
 
-    def mcnp(self, tally_index=1, mcnp_version='mcnp5'):
+    def mcnp(self, tally_index=1, mcnp_version='mcnp5', out=''):
         """mcnp(self, tally_index=1, mcnp_version='mcnp5')
```
gonuke:
```suggestion
        """Write the tally in a format suitable for MCNP.
```

pyne/tally.pyx
```diff
@@ -21,6 +24,32 @@ from pyne.utils import QAWarning
...
+    for ivalue in range(value_size):
+        value_proxy[ivalue] = <int > value[ivalue]
+    return value_proxy
+
+
+cdef vector[double] to_vector_double(value):
+    cdef vector[double] value_proxy
+    cdef int ivalue
+    cdef int value_size
+    cdef double * value_data
```
gonuke:
```suggestion
```

pyne/tally.pyx
```diff
@@ -21,6 +24,32 @@ from pyne.utils import QAWarning
 
 warn(__name__ + " is not yet QA compliant.", QAWarning)
 
+cdef vector[int] to_vector_int(value):
+    cdef vector[int] value_proxy
+    cdef int ivalue
+    cdef int value_size
+    cdef int * value_data
```
gonuke:
```suggestion
```

gonuke (approved):

gonuke:
Thanks for your work on this @bam241

Issue #1155, opened 2019-07-23
Tally normalization must be >1.0 for MCNP output


gonuke:
It's not clear why the MCNP formatted tally writer requires that the tally normalization be >1.0 in order to generate the normalization card.  Shouldn't it always write a normalization card if the normalization is set?

gonuke:
Closed with #1156

PR #1156, opened 2019-07-23
Now add normalization tally card when normalization factor != 1

bam241:
Adds the tally card if normalisation != 1

this should address #1155

### DIFF
--- a/src/tally.cpp
+++ b/src/tally.cpp
@@ -466,7 +466,7 @@ std::string pyne::Tally::form_mcnp_tally(int tally_index,
   if (entity_size > 0.0)
     tally_stream << "SD" << tally_index << type << " " << entity_size << std::endl;
   
-  if (normalization > 1.0)
+  if (normalization != 1.0)
     tally_stream << "FM" << tally_index << type << " " << normalization << std::endl;
 
   return tally_stream.str();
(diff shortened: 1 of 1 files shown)


bam241:
is that the expected behavior ? @gonuke

gonuke:
Makes sense to me - probably should see if @makeclean had something else in mind originally?

gonuke (approved):
Thanks @bam241 - this LGTM - still hoping for some input from @makeclean - or I'll look to merge next week

makeclean:
Sorry for the delay! So the reason why it was originally > 0 is that a negative tally multiplier (in MCNP speak) means multiply by atomic density, I expect other MC codes have the similar side affects. I always figured that we would eventually have some infrastructure that would take care of knowing that the normalisation referred purely to the number of source particles per second, and then other code would know which cell the tally refers to, with knowledge of the atomic density, the type of tally, e.g. heat would eventually be able to deal with the edge cases. Therefore, the assumption here is that the normalisation should be greater than 0 since it refers to the true normalisation. 

What stopped progress dead (for me) is/was the limitation of Cubit groups, as an analyst one often wants better grouping of tallies for your down stream purpose. Now, of course with the right downstream infrastructure, one should usually be able to combine the tally data into whatever you want.

gonuke:
But the previous code was ">1" and not ">0"...

gonuke:
I think the issue of `!=1` vs `>1` is sufficiently vague and this seems more consistent.

Thanks @bam241

PR #1157, opened 2019-07-23
Add ebins in source.h5m

zxkjack123:
DO NOT MERGE!
This PR adds energy boundaries data into `source.h5m.

For PyNE R2S coupling with MCNP, the energy boundaries are stored in a text file `e_bounds`. Well, it will be useful if we put more information in the h5m file:
- Energy bins;
- Source data (already there);
- Total photon source intensities;
- Cooling time (s).

Adding these into the `source.h5m` will be useful for:
- Check the content of energy by `h5dump` to make sure right file is used;
- Read the energy bins and other data directly from h5m file.

TO DO:
- [ ] : add e_bins
- [ ] : add total photon source intensities
- [ ] : cooling time


pyne/r2s.py
```diff
@@ -222,3 +223,18 @@ def total_photon_source_intensity(m, tag_name, sub_voxel=False):
...
+    photon source file 'source_x.h5m'.
+    Parameters
+    ----------
+    filename : string
+        Filename of the 'source_x.h5m'.
+    e_bins : numpy array
+        Energy boundaries, unit: MeV.
+    """
+    with tb.open_file(filename, 'r+') as h5f:
+        h5f.create_array('/', 'energy_boundaries', obj=e_bins)
```
gonuke, line 248: should we put this deeper in the HDF hierarchy?  (perhaps other info should also be deeper??)

gonuke (commented):
Just a comment/question about where to put this in the HDF hierarchy.

This will be very useful for MCNP as well and we should look to read this into MCNP with the source sampling routines.

gonuke:
Have we documented the HDF5 hierarchy for source mesh files?  That might be useful to help us plan/ design how we add things

zxkjack123:
@gonuke No. I didn't find any documentation about the HDF5 hierarchy for source mesh files. I usually use `h5dump` to find out the place of the data I want. We should add that.

zxkjack123:
I original motivation of this PR is to add the `e_bounds` information for OpenMC. But I find that defining the `e_bounds` in openmc python API is a better way. 

Therefore, this PR is not longer valuable for OpenMC R2S.

But for DAG-MCNP purpose, I think we can define the energy boundaries in `rdum` rather than reading `e_bounds` file. Because reading `e_bounds` from external files (`e_bounds` or `source.h5m`), always requires extra code and API. I think it will be easier if we use native supported functions to get `e_bounds` information.

@gonuke How do you think about this change?

gonuke:
I definitely prefer to have the `e_bounds` info in the file.  Best practice suggests that you want to keep data like that tightly bound together.  If not, it because too easy for users to include the wrong `e_bounds` info in their input files.

gonuke:
The photon source data is tagged on a mesh via MOAB.  Can’t we use a MOAB tag for this data, also?

zxkjack123:
@gonuke We can, but the data of `e_bounds` will be repeated for many times.
@pshriwise Is there a MOAB tag works on the entire mesh, but not each mesh element?

gonuke:
You can put a sparse tag on the root set and it will only occur once - there are other examples in a typical H5M DAGMC geometry

gonuke:
Are you planning to change this to a sparse tag on the root set?

zxkjack123:
@gonuke I plan to write `e_bounds` and total photon source intensities as sparse tag on the root set. I will implement this change after the problem of PR of `activation_response` get fixed.

gonuke:
@zxkjack123 - what's the status here?  Will you continue this effort?

zxkjack123:
@gonuke Sorry for the late response. I can't work on this PR recently for some personal reasons. It's better to set is as deferred, or someone else could work on this if it's needed in a hurry.

zxkjack123:
This PR is replaced by #1341

Issue #1111, opened 2019-03-14
allow user to turn off void rejection in source sampling routine


gonuke:
In the source sampling routine included for MCNP, the second value on the `idum` card indicates how many times to reject a sample because it has been born in the void.  This is based on the origin of this sampling capability - sampling a photon source from activated material. 

However, if one wants to use those same routines to sample a mesh for other purposes - e.g. a neutron source in the "vacuum" of a plasma chamber, then it is not appropriate to reject the voids.

One simple solution is to use the second `idum` entry; if it's value is < 0 then void rejection is turned off.

zxkjack123:
Close it, as it has been resolved in PR #1158.

PR #1158, opened 2019-07-24
R2s allow turn off void rejection

zxkjack123:
As suggested in #1111 , this PR makes the changes to allow user turn off the void rejection.

Using the variable `idum(2)`:
- If it is a positive integer, the void rejection is turned on
- If it is a negtive integer, the void rejection is turned off

### DIFF
--- a/docs/usersguide/source_sampling.rst
+++ b/docs/usersguide/source_sampling.rst
@@ -226,11 +226,14 @@ groups used in the activation calculations. An "idum" card must be used
 in the MCNP5 input file. This card should have three arguments. The first is the
 sampling mode (0: DEFAULT_ANALOG, 1: DEFAULT_UNIFORM, 2: DEFAULT_USER,
 3: SUBVOXEL_ANALOG, 4: SUBVOXEL_UNIFORM, 5: SUBVOXEL_USER). The second is the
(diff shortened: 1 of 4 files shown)


docs/usersguide/source_sampling.rst
```diff
@@ -226,11 +226,14 @@ groups used in the activation calculations. An "idum" card must be used
...
-cell number, the source position is resampled within the selected mesh volume
-element until either a correct position is found, or this user-specified limit
-is researched. The third argument should specify the particle type: 1 for
+resample limit for void and cell rejections. If the second argument is set to
+be a postive integer, the void rejection will be applied, i.e., for a given
+particle, if a source position is selected in void (MCNP material 0) or in a
+cell that disagrees with the cell number, the source position is resampled
+within the selected mesh volume element until either a correct position is
+found, or this user-specified limit is researched. If the second argument is
+set to be a negtive integer (-1 is recommended), then the void rejection will
```
gonuke: What if we recommend 0 to turn it off.  This is more consistent with other usage, I think.

gonuke (commented):
Thanks @zxkjack123 - I thought this feature was already available?!?  Thanks for making sure that it is

gonuke:
Thanks @zxkjack123

Issue #1137, opened 2019-05-23
Add statistical summary for failure modes of `find_cell`


gonuke:
The `find_cell` function added in #1118 adds important performance improvements and identifies some possible failure modes in finding the cell number.

It would be useful to know, at the end of each simulation, how many times each failure mode was encountered.  This could be important diagnostic information for improving the modeling of a given problem, e.g increasing the number of rays in `discretize_geom`/R2S step 1

gonuke:
Resolved in #1159

PR #1159, opened 2019-07-26
Find cell failure statistics

zxkjack123:
DO NOT MERGE!

This PR adds statistics summary output in the source sampling as suggested in #1137 .

There are three types of `find_cell` failure:

| Type  | Resaon  | Frenquency | Treatment | Suggestion |
| ------------- | ------------- | -------------- | -------------- | -------------- |
|  1  | Sorce particle is located in a cell that exist in neutron transport but removed in photon transport. Therefore, the cell number does not exist in the `ncl` list.  | Small frequency if part of geometry removed. | Skip, re-sample | None |
|  2  | Source particle is not located in the given `cell_list`. Because `cell_list` from `discretize_geom()` missed some cells with very small volume fractions. Caused by random error,. | Small frequency |  Skip, re-sample, warning | Increase `num_rays` in R2S step1 |
| 3 | Source partilce can't be found in any cell.  This is an error. When this error happens, it means that there is something wrong in either the source.F90 file or the DAGMC geometry representation. | Low frenquency for some complex geometry | Skip, re-samling, error message | Check the code and geometry |

The following is an (FNG photon transport) example summary message:
```
Cell not found error summary:
 Type1 warning counts:       48582 /              1000000
 Type2 warning counts:        1517 /              1000000
 Type3 error counts:           0 /              1000000
 run terminated when     1000000  particle histories were done.
```
To do:
- [ ] : parallel run support
- [ ] : continue run support

### DIFF
--- a/share/source.F90
+++ b/share/source.F90
@@ -105,8 +105,8 @@ subroutine find_cell(cell_list, cell_list_size, icl_tmp, count_1, count_2, count
           ! reset the icl_tmp to -1 because of the type 2 not found
           icl_tmp = -1
           count_2 = count_2 + 1
-          return
         endif
+        return
       endif
     enddo
     ! icl now is -1, it is a type 3 error.
--- a/share/source.F90
+++ b/share/source.F90
@@ -137,7 +137,6 @@ subroutine source
   logical, save :: first_run = .true.
   real(dknd), dimension(6) :: rands
   integer :: icl_tmp ! temporary cell index variable
-  !integer :: find_cell
   integer :: tries
   integer, save :: cell_list_size = 0
   integer, dimension(:), allocatable, save :: cell_list
--- a/share/source.F90
+++ b/share/source.F90
@@ -12,7 +12,7 @@
 ! Full instructions on compiling and using MCNP5 with this subroutine are found
 ! in the PyNE user manual.
 
-function find_cell(cell_list, cell_list_size) result(icl_tmp)
+subroutine find_cell(cell_list, cell_list_size, icl_tmp, count_1, count_2, count_3)
 ! This function determines the current MCNP cell index location.
 ! Return a positive integer if a valid cell is found, otherwise it returns -1.
 ! This only works if there are no repeated geometries or universes present in
@@ -60,9 +60,10 @@ function find_cell(cell_list, cell_list_size) result(icl_tmp)
 
   integer :: i ! iterator variable
   integer :: j ! temporary cell test
-  integer :: icl_tmp ! temporary cell variable
   integer, intent(in) :: cell_list_size
   integer, dimension(cell_list_size), intent(in) :: cell_list
+  integer, intent(out) :: icl_tmp ! temporary cell variable
+  integer, intent(out) :: count_1, count_2, count_3 ! failure counter
   integer :: cidx ! cell index
 
   icl_tmp = -1
@@ -78,13 +79,14 @@ function find_cell(cell_list, cell_list_size) result(icl_tmp)
       cidx = namchg(1, cell_list(i))
       if (cidx .eq. 0) then
         ! Type 1: cell index not found, skip and resampling
-        exit
+        count_1 = count_1 + 1
+        return
       endif
       call chkcel(cidx, 0, j)
       if (j .eq. 0) then
         ! valid cell found
         icl_tmp = cidx
-        exit
+        return
       endif
     enddo
   endif
@@ -102,19 +104,22 @@ function find_cell(cell_list, cell_list_size) result(icl_tmp)
           ! this is a type 2 problem, skip
           ! reset the icl_tmp to -1 because of the type 2 not found
           icl_tmp = -1
+          count_2 = count_2 + 1
+          return
         endif
-        exit
       endif
     enddo
     ! icl now is -1, it is a type 3 error.
     ! Skip and print error message
     if(icl_tmp .le. 0) then
+      count_3 = count_3 + 1
       write(*,*) 'ERROR: history ', nps, 'at position ', &
       &          xxx, yyy, zzz, ' not in any cell'
       write(*,*) 'Skipping and resampling the source particle'
     endif
   endif
-end function find_cell
+  return
+end subroutine find_cell
 
 subroutine source
   ! This subroutine is called directly by MCNP to select particle birth
@@ -125,21 +130,28 @@ subroutine source
   use mcnp_random, only: rang
   use pblcom, only: xxx, yyy, zzz, erg, tme, wgt, icl, ipt, jsu
   use mcnp_debug
+  use varcom, only: nps, npp
 
   implicit none
 
   logical, save :: first_run = .true.
   real(dknd), dimension(6) :: rands
   integer :: icl_tmp ! temporary cell index variable
-  integer :: find_cell
+  !integer :: find_cell
   integer :: tries
   integer, save :: cell_list_size = 0
   integer, dimension(:), allocatable, save :: cell_list
+  ! counter for type 1, 2 and 3 failure.
+  integer, save :: count_1, count_2, count_3
 
   if (first_run .eqv. .true.) then
     ! set up, and return cell_list_size to create a cell_list
     call sampling_setup(idum(1), cell_list_size)
     allocate(cell_list(cell_list_size))
+    ! initialize failure counter
+    count_1 = 0
+    count_2 = 0
+    count_3 = 0
     first_run = .false.
   endif
 
@@ -155,7 +167,7 @@ subroutine source
 
   call particle_birth(rands, xxx, yyy, zzz, erg, wgt, cell_list)
   ! Loop over cell_list to find icl_tmp
-  icl_tmp = find_cell(cell_list, cell_list_size)
+  call find_cell(cell_list, cell_list_size, icl_tmp, count_1, count_2, count_3)
 
   ! check whether this is a valid cell
   if (icl_tmp .le. 0) then
@@ -185,5 +197,17 @@ subroutine source
   ipt = idum(3)
   jsu = 0
 
+  if ((nps .eq. npp) .and. (count_1 + count_2 + count_3 > 0)) then
+    write(*, *) "Cell not found error summary:"
+    write(*, *) "Type1 warning counts:", count_1, "/", npp
+    write(*, *) "Type2 warning counts:", count_2, "/", npp
+    if (count_2 / DBLE(npp) > 0.05) then
+      write(*, *) "Suggest to increase num_rays in r2s step1"
+    endif
+    write(*, *) "Type3 error counts:", count_3, "/", npp
+    if (count_3 > 0) then
+        write(*, *) "Check the geometry!"
+    endif
+  endif
   return
 end subroutine source
--- a/share/source_mcnp6.F90
+++ b/share/source_mcnp6.F90
@@ -100,8 +100,8 @@ subroutine find_cell(cell_list, cell_list_size, icl_tmp, count_1, count_2, count
           ! reset the icl_tmp to -1 because of the type 2 not found
           icl_tmp = -1
           count_2 = count_2 + 1
-          return
         endif
+        return
       endif
     enddo
     ! icl now is -1, it is a type 3 error.
--- a/share/source_mcnp6.F90
+++ b/share/source_mcnp6.F90
@@ -12,7 +12,7 @@
 ! Full instructions on compiling and using MCNP6 with this subroutine are found
 ! in the PyNE user manual.
 
-function find_cell(cell_list, cell_list_size) result(icl_tmp)
+subroutine find_cell(cell_list, cell_list_size, icl_tmp, count_1, count_2, count_3)
 ! This function determines the current MCNP cell index location.
 ! Return a positive integer if a valid cell is found, otherwise it returns -1.
 ! This only works if there are no repeated geometries or universes present in
@@ -57,6 +57,8 @@ function find_cell(cell_list, cell_list_size) result(icl_tmp)
   integer :: icl_tmp ! temporary cell variable
   integer, intent(in) :: cell_list_size
   integer, dimension(cell_list_size), intent(in) :: cell_list
+  integer, intent(out) :: icl_tmp ! temporary cell variable
+  integer, intent(out) :: count_1, count_2, count_3 ! failure counter
   integer :: cidx ! cell index
 
   icl_tmp = -1
@@ -72,13 +74,14 @@ function find_cell(cell_list, cell_list_size) result(icl_tmp)
       cidx = namchg(1, cell_list(i))
       if (cidx .eq. 0) then
         ! Type 1: cell index not found, skip and resampling
-        exit
+        count_1 = count_1 + 1
+        return
       endif
       call chkcel(cidx, 0, j)
       if (j .eq. 0) then
         ! valid cell found
         icl_tmp = cidx
-        exit
+        return
       endif
     enddo
   endif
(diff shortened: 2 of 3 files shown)


zxkjack123:
@ljacobson64 The `source_mcnp6.F90` is changed in this PR, but I can't test it, not sure whether it works well. Could you please help me with it?

pshriwise:
The Type 1 failure rate is fairly high. This is jut a passing a thought, but would there be a way to “alias” cells that were present in the neutron transport model but not in the photon model? Meaning that the space cell X occupied in the neutron transport model is now occupied by cell Y. Perhaps this aliasing could be added to the discretize geom file?

This might only be worth the effort if changes like this are fairly common when using R2S.

zxkjack123:
@pshriwise If we want to add that `alias` by performing `discretize_geom`, we need to actually compare the CAD models for both neutron transport and photon transport. Is is a fairly a difficult task. And a cell X in neutron transport, could be:
- A: Removed in photon transport -> cell X no longer in photon transport model
- B: Replaced by cell Y
- C: Replcaed by cell Y, Z, ...
- D: Partly replaced by cell Y, Z, ...

For the first case, we don't need to add the `alias`, because in the source sampling procedure, if we don't find this `X`, the code will iterate in the entire model to find out the place it's located. And then, it will be found to be in void, then rejected. Then, the results would not be influenced.

However, for the rest case (B, C, D), the source sampling would lead to a systematic error as the material here is not void, and the photons will sampled in these new cells with new material, leading to an error.

Theoretically, an `alias` table for cell `X` would be good.  However, in practice, `discretize_geom` itself is not  so accurate when we use random rays. This is also the reason of Type 2 warning.

Therefore, my option is, under the current situation, it's hard and even not achievable create an accurate `alias` table for cell X by the `discretize_geom`.

pshriwise:
This is true. I guess I wasn’t thinking of it as an automated process but an option we could add. In my mind the alias table would be manually defined by the user.

Regarding the Type 2 errors, I was mainly looking to address the Type 1 errors as they seem to be more common. Type 2 errors are definitely going to be prone to random error and it’s good to see that the frequency at which they occur is relatively low compared to the Type 1 frequency. 

Anyway, it was just a thought. If you don’t think it’ll be a useful feature then it probably isn’t worth the time. The better answer is probably for the user to reassign the materials of those cells rather than remove then from the model, yeah?

zxkjack123:
@pshriwise Yes, manually define material for the cells with changed materials in neutron/photon transport would be a better way than remove it. And remove it only works when this cell will be replaced by void.

And I think we can go further into this material define process. Here is what I am thinking: In Trelis, if we define the materials for both neutron and photon transport, it would be convenient for better accuracy. 
- For most volumes, they have the same material in neutron and photon transport. Define the mateerial as the current workflow (`mat:<name of material>/rho:<density>`).
- For some volumes, which will change the material in photon transport, define the two materials. Eg. `matn:<name_of_material>/rho:<density>|matp:<name of material>/rho:<density>`). Then use `matn` in neutron transport, and use `matp` in photon transport.

By this way, we can:
- Avoid preparing two models for neutron and photon transport if there is only material changes
- Know the cells with material changes in advance, do not activate it, or do not sample photon particle in that cell
- If the material of cell X is the same in neutron and photon transport, they can still be changed. It could be used to represent the component replacement.

How do you think about this idea?

pshriwise:
That seems like a pretty specific naming scheme for R2S. I think maybe it would be better to have R2S automatically update the material at some point in the workflow. Now that PyNE relies on PyMOAB this could be a function we add to `R2S.py` to change the material assignment for photon transport.

zxkjack123:
In the current situation, I agree that do not remove a cell, but change its material to void is a better choice.  Then, there should be no Type1 error at all. But we still need to keep the detection of Type1 warning, as remove of a cell is a supported choice.

Talking about changing material in photon transport, I agree that my suggestion is beneficial to R2S. But there is something tricky: if the users want to change the materials of some volumes, we must have ability to distinguish them, they must set them is a separate group in the material assignment process.

For example, there are two volumes A and B, they use the same material `water` in neutron transport. But their materials are `water` and `air` in photon transport. That means, you have to assgin them in different groups/names. On the extreme condition, the user need to assign each volume a specific group. That's even harder than prepare a photon transport model.

Therefore, if we want to change the material at some time, we'd better change the material based on volume rather than material name. I don't think there is a better way than assign multiple material for each volume in Trelis if we want to change the material.

pshriwise:
I don’t have any particular problem with what you’ve proposed :) I was thinking ahead to how we’d update those material assignments using the “special” groups to be created in Trelis and I was thinking we’d want to handle that in PyNE, maybe using PyMOAB directly. 

Here’s how I envision it working, let me know if you have a different plan in mind:


*In Trelis*
  - material groups are assigned as usual in Trelis for neutron transport (`mat:steel/rho:8.08`) and applied as normal
  - volumes whose materials change for the photon transport step will have a different material name, perhaps something like `mat-photon:aluminum/rho:2.7`, where “mat-photon” indicates that any volumes in this group should be filled with aluminum for photon transport only

*During R2S*

  - No updates to the DagMC file are necessary for the neutron transport
  - Before photon transport is performed, a function in PyNE relying on PyMOAB will find any groups with the “mat-photon” keyword (or any other keyword we decide upon) and reassign those volumes appropriately
   - photon transport commences

The reason I think this is a good approach is that it requires no changes to the UWUW library in terms of recognizing a special group name specialized for the purpose of R2S.

That was what I was not-so-clearly trying to convey before. Thoughts?

zxkjack123:
@pshriwise Got your idea.
I rephrase my thoughts using your format below:
In Trelis

- volumes whose materials will not change are assigned as usual in Trelis for neutron transport (`mat:steel/rho:8.08`) and applied as normal
- volumes whose materials change for the photon transport step will have a different material name,  something like `mat:steel/rho:8.08|mat-photon:aluminum/rho:2.7`

During R2S

- **Updates required** for the DagMC material recognize, add `mat-photon` material name support
- Neutron transport uses materials in neutron format, such as `mat:steel/rho:8.08`
- **No changes** in R2S workflow
- photon transport commences, uses materials in photon format, such as `mat-photon:aluminum/rho:2.7`

Your approach involves no change to the existing DAGMC code (especially for UWUW material lib name protocol), it will have a better compatibility with older version files. My plan involves changes in DAGMC code, only a little changes need to make in PyNE.

I must admit that your approach is better than mine. Make changes in pain is easier to do than DAGMC.

There is another thing I am concerned about. In your plan, we need to assign a volume into two groups if its material will change. Will this lead to conflicts in material assignment and exportation? 

And I think the material name recognition process (part of UWUW library) requires minor change to support this feature. Here are two possible changes in my mind:
- A material assignment checking.
    - Make sure every volume has a material;
    - And no volume has more than three materials.  (If there is no conflict when we assign material in Trelis, that means we may assignment two different neutron or photon material for a volume)
    - If a volume is in two groups, they must be a `mat` and `mat-photon`
- Add `mat-photon` support. At least we need to let the code know that we want to store these materials into the geometry.

pshriwise:
Glad we’re on the same page.

> There is another thing I am concerned about. In your plan, we need to assign a volume into two groups if its material will change. Will this lead to conflicts in material assignment and exportation?

I don’t foresee any problems exporting groups names from Trelis (other than maybe hitting the unfortunate 32 character limit on group names we have now). 

I would really rather avoid changing the UWUW library at all as this use case is specific to the R2S workflow and R2S’s goals are separate from UWUW’s. If we modify the .h5m file and update material assignments before the simulation begins by making this an additional sub-step in the R2S process, UWUW can treat the material assignments like any other simulation.

zxkjack123:
@pshriwise Agreed! And this deserves some effort to implement.

gonuke:
What if we call it `r2s-photon-mat` (or something that makes it even more clear when/how it’s used)

zxkjack123:
@gonuke @pshriwise I am thinking of the usage of allowing a volume to be assigned into multiple groups. It's not only benefit R2S. In R2S, we use `mode` as the trigger the use different material set. By allowing other triggers, we can achieve things more than R2S:
- Use time stamp as a trigger: Time step changes geometry. Such as time step R2S of moving system that Chelsea has studied.
- User defined trigger cases: Some (groups of) volumes have different material setting in different cases for material sensitivity analysis.
- Material updates considering the depletion/transmutation: coupling with depletion/transmutation codes, and with the help of PyNE/PyMOAB to change the material, a time dependent simulation could be achieved

These are some applications I am thinking of. Of course, it can be used on R2S, with a user defined trigger. But name it as `r2s-photon-mat` will limit the usage to R2S. I think we should consider more application of this material change function.

The core of this function is to use a new material to replace an old material, R2S material change is just an application of this function. And for the R2S case, it is replacing the `neutron` material with `photon` material. Set it as `r2s-photon-mat` looks good to me.

gonuke:
Maybe we should take this conversation out of this PR. 🙂

It’s an important conversation, but it’s buried here.

zxkjack123:
@gonuke Agreed!. the topic seems to be "the implementation and application of material change using pymoab". I am not sure whether we should convert it to an issue or a project.

pshriwise:
@zxkjack123  I'll vote for an issue for now. Once we've finished this discussion, we'll then have a project scoped out and can start one up. Sound good?

zxkjack123:
@pshriwise Sounds like a good idea.

zxkjack123:
@gonuke @pshriwise I moved the content of material change to #1163 . Now let's focus on this PR. Any suggestion about this PR?

share/source.F90
```diff
@@ -102,19 +104,22 @@ end function namchg
           ! this is a type 2 problem, skip
           ! reset the icl_tmp to -1 because of the type 2 not found
           icl_tmp = -1
+          count_2 = count_2 + 1
+          return
```
gonuke: The exit point of this method moved inside the conditional... doesn’t that change the behavior?

share/source.F90
```diff
@@ -125,21 +130,28 @@ subroutine source
...
   use mcnp_debug
+  use varcom, only: nps, npp
 
   implicit none
 
   logical, save :: first_run = .true.
   real(dknd), dimension(6) :: rands
   integer :: icl_tmp ! temporary cell index variable
-  integer :: find_cell
+  !integer :: find_cell
```
gonuke: Delete this line instead of comment

gonuke (changes requested):
Thanks for this @zxkjack123.

As a separate issue, we should consider how we can make the FORTRAN modular and reuse it across different MCNP versions.

share/source.F90
zxkjack123: Well, I tested with it. It turns out that `return` is the right one. The previous `exit` only exit the current block and goes to the following code, which slows down the `find_cell`.

share/source.F90
zxkjack123: Changed in https://github.com/pyne/pyne/pull/1159/commits/e23c72e7bf794e3521cdfedc9dd4ea857456f191.

zxkjack123:
@gonuke To reuse the code, I think we need to:
- Put the `find_cell` in a separate file/module.
- Convert the print warning message part as a function. Also put in a separate  file/module.

Maybe patch could be used to write these functions to the chosen modlue file.

share/source.F90
gonuke: But there is no way to leave the `mxa` loop if a match is found.  It looks like it will always test every volume in the problem!

gonuke: Maybe you need both exit points?  One to exit with a type 2 error and one to exit with a successful match?

gonuke: Actually, I think the previous exit point should be faster.  If you find a matching cell with `chkcel` you always want to immediately return, either with that cell ID or with the error.

share/source_mcnp6.F90
```diff
@@ -96,19 +99,22 @@ function find_cell(cell_list, cell_list_size) result(icl_tmp)
           ! this is a type 2 problem, skip
           ! reset the icl_tmp to -1 because of the type 2 not found
           icl_tmp = -1
+          count_2 = count_2 + 1
+          return
```
gonuke: Just leaving a note here for the same discussion as above

gonuke:
Because of where this lives in the FORTRAN, it might be difficult to modularize!

gonuke:
Maybe a FORTRAN include?

share/source_mcnp6.F90
```diff
@@ -177,6 +191,17 @@ subroutine source
   pbl%r%tme = 0.0
   pbl%i%ipt = idum(3)
   pbl%i%jsu = 0
-
+  if ((nps .eq. npp) .and. (count_1 + count_2 + count_3 > 0)) then
```
pshriwise, line 196: What does this output look like for a parallel run? Does it print for each process?

share/source.F90
zxkjack123: > Actually, I think the previous exit point should be faster. If you find a matching cell with `chkcel` you always want to immediately return, either with that cell ID or with the error.

Yes, this is what we want. But the `exit` statement just ends the current `do` loop. The following type 3 check will still be running and is always true if a type 2 warning is found. this causes the wrong count of type3 error.

The `return` here is used to end of the entire `find_cell` subroutine and returns the current output value. The `return` statement seems to be obsolete in Fortran version newer than 95. 

There needs another variable to identify the warning type to differentiate type 2 and 3 if the `exit` is used.

share/source_mcnp6.F90
zxkjack123: Oh, forgot to test it in parallel. Just did it and found ugly output message. Each thread prints the message once. And their numbers are slightly different.

pshriwise: Right on. Thanks for looking into it. We might want to do a check that we’re on the master process here and accumulate those stats here too

share/source.F90
gonuke: What about a `return` statement where the `exit` statement used to be?

I don't think `return` is obsolete.  I think it is obsolete to provide a return value in a `return` statement.

zxkjack123: > What about a `return` statement where the `exit` statement used to be?
> 
> I don't think `return` is obsolete. I think it is obsolete to provide a return value in a `return` statement.

Well, in  this PR, it is a `return` statement to replace the `exit`.
And for the `return` statement, I checked the tutorial I am referring: http://wwweic.eri.u-tokyo.ac.jp/computer/manual/altix/compile/Fortran/Intel_Fdoc100/main_for/mergedProjects/lref_for/source_files/rfreturn.htm You are right, it says `alternate return is an obsolescent feature in Fortran 95 and Fortran 90`. It's `altrnate return` (with a return value) not `return`.

gonuke: I don't have a problem with `return`, but it's in the wrong place.  It should be outside the `if` block

zxkjack123: Updated in https://github.com/pyne/pyne/pull/1159/commits/646b00c01be1073dacc3fb93a045d34cd7a82768.

share/source_mcnp6.F90
gonuke: we probably need to do some work to ensure that these variables are accumulated over the difference processes - that might require some assistance from someone with MCNP source code access

gonuke: Similarly, this is probably not handled correctly for a continue run.

zxkjack123: I didn't tested with a continue run, it will probably get the wrong value.

gonuke (approved):

gonuke:
Thanks @zxkjack123 - I'm going to merge this.  I don't think it's easy to handle this properly for MPI and/or continue run.  It will require and MCNP patch and that is much more intrusive than the current approach of providing a stand-alone source sampling capability.

PR #1167, opened 2019-08-22
Openmc module change name

zxkjack123:
Rename the `pyne/openmc.py` to `pyne/openmc_utils.py`, for #1166 .


gonuke:
Thanks @zxkjack123

PR #1168, opened 2019-08-27
Add redistribute premission tendl

zxkjack123:
This PR adds the permission to redistribute TENDL files as part of PyNE package.


gonuke:
Thanks @zxkjack123

PR #1170, opened 2019-09-05
DagOpenMC support for PyNE R2S

zxkjack123:
DO NOT MERGE!

This PR adds OpenMC python API as an  optional dependency to add OpenMC support for PyNE R2S workflow. 

Functions that read the neutron flux from OpenMC and write the photon source information to `h5m` files are added:
- Move the `MeshTally` class from `pyne.mcnp` to `pyne.mesh`
- Reads the OpenMC state point file (containing mesh tally) and creates a `MeshTally` instance for it
- Setup R2S using data from OpenMC state point file
- A docker image: `py3_dagmc_pymoab_openmc` is added and the functions depends on OpenMC Python API is tested via CI.

Photon source sampling changes for OpenMC is NOT included in this PR.

This PR will be replaced by several smaller PRs:
- [x] : cleanup of dockerfiles (see comment)
- [x] : addition of OpenMC mesh tallies
- [ ] : addition of OpenMC photon source files
- [ ] : changes to R2S to support OpenMC


zxkjack123:
This PR replaces the previous PR: https://github.com/pyne/pyne/pull/1146. And this PR is also part of the solution of #1166 .

docker/ubuntu_18.04-dev.dockerfile
```diff
@@ -125,9 +125,9 @@ ARG build_pyne=YES
 # Build/Install PyNE
 RUN if [ "$build_pyne" = "YES" ]; then \
         cd $HOME/opt \
-        && git clone https://github.com/cnerg/pyne.git \
+        && git clone https://github.com/pyne/pyne.git \
```
gonuke, line 128: These are important changes that we missed!!  Can you make a separate PR with these corrections - and any similar ones that remove reference to CNERG or specific repo branches?

pyne/mcnp.py
```diff
@@ -1998,201 +1998,180 @@ def _read_tallies(self, f):
...
         tag_names : iterable, optional
             Four strs that specify the tag names for the results, relative
             errors, total results and relative errors of the total results.
             This should come from the Meshtal.tags attribute dict.
-        mesh_has_mats : bool
-             If false, Meshtally objects will be created without PyNE material
-             objects.
+        mesh_has_mats: bool
+            If false, Meshtally objects will be created without PyNE material
+            objects.
```
gonuke, line 2027: Include return object in docstring

pyne/mesh.py
```diff
@@ -1510,6 +1509,162 @@ def _do_op(self, other, tags, op, in_place=True):
...
+    
+            self.tag(self.tag_names[3], size=1,
+                     dtype=float, tagtype='nat_mesh')
+            rel_err_tot_tag = self.get_tag(self.tag_names[3])
+    
+            res_tot_tag[:] = res_tot
+            rel_err_tot_tag[:] = rel_err_tot
+
+
+    def tag_flux_error_from_openmc_tally_results(self, result, rel_err,
```
gonuke, line 1621: It seems like we'll get different tag structures depending on whether it's MCNP or OpenMC. Is that right?  Shouldn't we figure out a single canonical structure?  Maybe that will break backwards compatibility for MCNP?

scripts/r2s.py
```diff
@@ -183,7 +203,10 @@ def step2():
         tags = {('TOTAL', dc): tag_name}
         photon_source_hdf5_to_mesh(mesh, h5_file, tags, sub_voxel=sub_voxel,
                                    cell_mats=cell_mats)
-        mesh.write_hdf5('{0}_{1}.h5m'.format(output, i+1))
+        p_src_filename = '{0}_{1}.h5m'.format(output, i+1)
+        mesh.write_hdf5(p_src_filename)
+        if transport_code.lower() == 'openmc':
```
gonuke, line 208: Does it break anything in MCNP if the filetype attribute is also set there?  Seems like a useful thing to do in all cases.

gonuke (changes requested):
I made a few comments already, but I'm tempted to request this be split into a few different PR's:
* cleanup of dockerfiles (see comment)
* addition of OpenMC mesh tallies
* addition of OpenMC photon source files
* changes to R2S to support OpenMC

These different parts all require different expertise/knowledge and will ensure a thorough review.

pyne/mesh.py
zxkjack123: Yes, the results from `MCNP` and `openmc` has different shapes. The one from `MCNP` need to transpose before tagging data to mesh. We can use the same tag method if we make the conversion before tagging.

scripts/r2s.py
zxkjack123: It will not break anything in MCNP. This parameter is currently not used in DAG-MCNP. We can keep it there is possible usage of it.

pyne/mesh.py
gonuke: If possible, I think it would be good for mesh tally files to have the same structure regardless of their source

docker/ubuntu_18.04-dev.dockerfile
zxkjack123: The separate PR about dockerfile see https://github.com/pyne/pyne/pull/1172. I searched the code with `cnerg/pyne` and `pymoab_clean`, not any outdated ones found.

pyne/mesh.py
zxkjack123: Will do.

zxkjack123:
Close this PR, it will be split into several smaller PRs.

PR #1175, opened 2019-09-11
removing all tabs and trailing space

bam241:
just some cleaning to remove mix indentation....

### DIFF
--- a/src/measure.cpp
+++ b/src/measure.cpp
@@ -200,4 +200,3 @@ double measure( moab::EntityType type,
   }
 }
 
-
--- a/src/measure.cpp
+++ b/src/measure.cpp
@@ -32,25 +32,25 @@
 class CartVect {
   private:
     double coords[3];
-    
+
   public:
-  
+
     inline CartVect() {}
-    
+
     inline CartVect( double tx, double ty, double tz ) { set(tx,ty,tz); }
-    
+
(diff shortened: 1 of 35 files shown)


gonuke:
This needs a news file

bam241:
also removed tab from python files (replacing them with 4spaces)

src/measure.cpp
```diff
@@ -199,4 +199,5 @@ double measure( moab::EntityType type,
       return 0.0;
   }
 }
-      
+
+
```
gonuke: Why this extra line?

gonuke (commented):
Looks like you added a line (by accident?)

src/measure.cpp
bam241: an accident.

nice catch tho !

gonuke: It was lucky...  When I review these whitespace PR's, I set Github to ignore whitespace and then I should see no other changes.  Although I would consider this to also be whitespace, it showed it anyway

gonuke:
I think you need to restart tests...

bam241:
We need to do something about those test... endf website seems not be reliable enough...
They might kick CI out because of the amount of queries it does...

gonuke:
Thanks @bam241

PR #1176, opened 2019-09-13
add news checker in the CI-test

bam241:
This add a new CI-test to check if a news file have been created in the news folder.

I guessed as we rely on them, we might as well enforce them!


gonuke:
Thanks @bam241

PR #1178, opened 2019-09-26
fix news check trigger

bam241:
News check also ran after a merge on `develop` and fails because `develop` branch and `develop`  branch have the same number of news files...
this should prevent the news check test to be triggered after a merge.

News check should still be triggered on all the PRs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -238,7 +238,7 @@ workflows:
       - news_update:
           filters:
             branches:
-              ignore: master
+              ignore: develop
             tags:
               ignore: /.*/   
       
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -235,8 +235,13 @@ workflows:
   version: 2
   build_and_test:
     jobs:
-      - news_update
-        
+      - news_update:
+          filters:
+            branches:
+              ignore: master
```
gonuke: Should this be `develop`?

bam241: Nice catch ! stole that form Cyclus... forgot to change it !!

gonuke:
I think this looks good to me - thanks @bam241

PR #1179, opened 2019-10-01
Attempt to fix the permission to push the website

bam241:
I check this in that order:
1. no fingerprints specified (to check if it pull all of it, as it should ... )
2. if 1. not working add the fingerprint (which is working...)
3. put back the trigger and filters

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -325,22 +325,23 @@ workflows:
      # Build/Push test website
       - build_push_website_test:
           requires:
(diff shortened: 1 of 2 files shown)


gonuke:
Perhaps a comment to @pyne-dev will result in an email arriving in their account.  If you receive this, please reply so that we can discuss access to the @pyne-dev account.

gonuke:
I have access to the pyne-dev account now, so we can do what we need to do...

Issue #720, opened 2015-04-22
Style review of endf module


scopatz:
Many issues came up when reviewing endl.  See #717 for some examples.

gonuke:
I think the issues in endf go beyond style.  ENDF is a fixed format FORTRAN based format and the ENDF module has been written with seek, reed, tell paradigms that need to determine a line length, etc.  I think a readline and tokenizing paradigm will be better.

In reviewing #1180 it became apparent that the existing code was hard to read and maintain.

PR #1180, opened 2019-10-02
Improve/fix the regular expressions in endf.pyx

furutaka:
The CONTENTS_R regular expression in endf.pyx is not so restrictive, and for example the last line of 
the following, excerpted from n-058_Ce_136.endf file of ENDF/B-VII.1, is thought of as a contents line, causing an error:

>       Collective levels used in direct calcualtions                   5825 1451  138
>                                                                       5825 1451  139
>         Ncoll  Lmax IDef  Kgs  (Def(1,j),j=2,IDef,2)                  5825 1451  140
>           12    4    4   0.0  0.170E+00  0.000E+00                    5825 1451  141

Other two regular expressions were also improved (IMHO) in the same way.


gonuke:
Thanks for this contribution @furutaka!  It is always a pleasure to receive new input from users.

I have two requests for updates to this modification:
1. Can you please add a `news` file?  You can see examples in the `news` directory from previous pull requests.  This helps us ensure that we can explain all the changes as we publish updates.

2. Can you add one or more tests?  Such tests will ideally demonstrate the failure using the old code and then demonstrate success using your update.  You may need to add a file to `tests/file_test_endf` as well as adding/modifying methods in `tests/test_endf.py`

pyne/endf.pyx
```diff
@@ -40,11 +40,10 @@ libraries = {0: 'ENDF/B', 1: 'ENDF/A', 2: 'JEFF', 3: 'EFF',
              4: 'ENDF/B High Energy', 5: 'CENDL', 6: 'JENDL',
              31: 'INDL/V', 32: 'INDL/A', 33: 'FENDL', 34: 'IRDF',
              35: 'BROND', 36: 'INGDB-90', 37: 'FENDL/A', 41: 'BROND'}
-FILE1_R = re.compile(r'1451 *\d{1,5}$')
-CONTENTS_R = re.compile(' +\d{1,2} +\d{1,3} +\d{1,10} +')
-SPACE66_R = re.compile(' {66}')
-NUMERICAL_DATA_R = re.compile('[\d\-+. ]{80}\n$')
+FILE1_R = re.compile(r'1451[ \d]{4}\d$')
```
gonuke: This still isn't quite right.

The old one will match incorrectly on  `1451  5` - not enough spaces and/or digits.
This one will match incorrectly on `1451  5 5` - a mixture of the correct number of spaces or digits.
Perhaps it should be:
```suggestion
FILE1_R = re.compile(r'1451(?=[ ]*\d+$)[ \d]{5}$')
```
This adds a look head to ensure that there is one group of 0+ spaces and the one group of 1+ digits before the new line, and then makes sure there are 5 of them, combined.

gonuke: It might be useful to add tests for these regex expressions that don't rely on a file, but simply document a set of developer-constructed cases that either should or should not match and then a test function that shows if/when they do/don't.

gonuke (changes requested):
I think your expressions are definitely an improvement, but since we are trying to use regex to match FORTRAN fixed format numbers, it's still not right.  I learned something useful here: https://stackoverflow.com/questions/5116422/regex-to-read-fixed-width-numeric-fields

Maybe your new expressions can be updated along these lines?

pyne/endf.pyx
furutaka: Let me sleep on it...

gonuke:
This is definitely looking a little cleaner - I like the idea of breaking each line into blocks of 11 characters.

However, I suddenly remembered that we already have some generic FORTRAN fixed-format reading capability: https://github.com/pyne/pyne/blob/develop/pyne/fortranformat/FortranRecordReader.py
I wonder if it's worth retooling for that?

furutaka:
Well... It seems to me that the FortranRecordReader.py is a tool for reading/converting values after knowing what a line is for.  But the problem I had was incorrect recognition of the Contents lines in MT=451, i.e. how to know/tell the Contents lines from others.

pyne/endf.pyx
```diff
@@ -152,18 +162,14 @@ class Library(rxdata.RxLib):
         # get mat_id
         line = fh.readline()
         mat_id = int(line[66:70].strip() or -1)
-        # store position of read
-        pos = fh.tell()
         # check for isomer (LIS0/LISO entry)
-        matflagstring = line + fh.read(3*self.line_length)
+        matflagstring = line + fh.read(3*len(line))
```
gonuke: Is there a reason to not use the member variable here?  The purpose of this variable is to be robust across Window vs Mac vs Linux.

pyne/endf.pyx
```diff
@@ -175,27 +181,24 @@ class Library(rxdata.RxLib):
...
                 stop = start + mt_length
                 self.mat_dict[nuc]['mfs'][mf, mt] = (self.line_length*start-self.offset,
                                                      self.line_length*stop-self.offset)
                 start = stop + 1
                 line = fh.readline()
             # parse comment
-            elif SPACE66_R.match(line):
-                self.structure[nuc]['docs'].append(line[0:66])
-                line = fh.readline()
-            elif NUMERICAL_DATA_R.match(line):
```
gonuke, line 196: Why don't we need this case any more?

gonuke (commented):
I really like how this is cleaned up by your changes.

I'm not an expert on ENDF file formats (perhaps we can get some help from @paulromano or someone else on this).

I have a few questions in this review to help gain confidence without my expertise.

pyne/endf.pyx
furutaka: Well... I'm trying to use pyne.endf to read JENDL4 files.  Unfortunately, some of which are with '\r\n' and others have '\n', and the reading failed for files w/ '\r\n'.  So I followed the function of Library._read_headers() and found that for the files w/ '\r\n' the fh.read(3*self.line_length) goes too further!  I think that it is because of the universal newline mode; by experimenting, it seems to me that the size argument of read() is not the number of characters in the file but the characters it returns, although I'm not quite sure...

pyne/endf.pyx
furutaka: My understanding is that there're three parts in a (MF,MT)=(1,451) file, the data part which comes first, then the document part, and at last the contents/dictionary part.  As the meaningful values in the data part has already read with fh.read(3*len(line)) (and I don't think we have to go back to the top to read the same data lines again), we only have to distinguish the contents lines from others, i.e. 'docs' lines.

furutaka: Ah, according to the [manual](https://www.nndc.bnl.gov/csewg/docs/endf-manual.pdf#page=54), there're only two parts, 1. a documentation, and 2. a dictionary, and 

> The first three records of the descriptive information contain a standardized presentation
of information on the material, projectile, evaluators, and modification status.

pyne/endf.pyx
furutaka: ##### Sr087.dat: ASCII text
_read_tpid: chars_til_now =  81
_read_tpid: line_length   =  81
_read_tpid: offset        =  0
_read_tpid: len(line)     =  81
##### Sr088.dat: ASCII text, with CRLF line terminators
_read_tpid: chars_til_now =  82
_read_tpid: line_length   =  82
_read_tpid: offset        =  0
_read_tpid: len(line)     =  81


This way, regardless of the newline characters (`'\r\n'` or `'\n'`) and the value of `self.line_length`, `len(line)` always returns 81, and `self.offset` is always 0.

`self.chars_til_now = len(line) + self.line_length - 81`
`self.offset = self.line_length - self.chars_til_now`

furutaka: It seems to me that the same thing (reading too much characters) occurs in endf.Library._read_nucmfmt(), and I'm wondering how to fix them...

furutaka:
What do you think of this?

gonuke:
Unfortunately, this is now failing the tests for the ENDF reader.  

(As I look at your changes, it seems like perhaps the underlying problem was the mixing of `read()` and `readline()` in the original code.  I think the original author may not have had an appreciation of FORTRAN fixed format and that the there might be a more straightforward way to do this whole module!)

furutaka:
The cause of the failure is obvious; the first line of the test file, sampletape1, has 75 characters only!  This is the first time I saw an ENDF-format evaluated nuclear data file with <80 characters...

Can I correct the test file?

By the way, I made a mistake and I'll correct it.

gonuke:
I submitted a PR for that incorrect test file.

gonuke (approved):

bam241:
thank you @furutaka !

gonuke:
I was just writing a message to merge this since it is definitely an improvement!  Thanks @furutaka!

However, I also think there is room for an overhaul of the ENDF reader since I think the overall approach is not well matched to the fixed format FORTRAN expectations.  It is not easy to parse or maintain this code.  I'll add to issue #720 that perhaps first hinted at this need.

PR #1188, opened 2019-10-17
Form UWUW material name

bam241:
I figured that as we are slowly moving on with the MaterialLibrary C++ API at some point we would need something like that...
this allows to form a UWUW material name, including the density if provided.

this solves #1188

### DIFF
--- a/pyne/material.pyx
+++ b/pyne/material.pyx
@@ -346,12 +346,12 @@ cdef class _Material:
         card = self.mat_pointer.mcnp(frac_type.encode())
         return card.decode()
     
-    def uwuw(self):
-        """uwuw()
+    def get_uwuw_name(self):
+        """get_uwuw_name()
         Return a uwuw material name
         """
         cdef std_string uwuw_name
-        uwuw_name = self.mat_pointer.uwuw()
+        uwuw_name = self.mat_pointer.get_uwuw_name()
         return uwuw_name.decode()
 
     def openmc(self, frac_type='mass'):
--- a/pyne/material.pyx
+++ b/pyne/material.pyx
@@ -346,7 +346,7 @@ cdef class _Material:
         card = self.mat_pointer.mcnp(frac_type.encode())
         return card.decode()
     
-    def mcnp(self):
+    def uwuw(self):
         """mcnp(frac_type)
         Return a uwuw material name
         """
--- a/pyne/material.pyx
+++ b/pyne/material.pyx
@@ -347,7 +347,7 @@ cdef class _Material:
         return card.decode()
     
     def uwuw(self):
-        """mcnp(frac_type)
+        """uwuw()
         Return a uwuw material name
         """
         cdef std_string uwuw_name
--- a/pyne/material.pyx
+++ b/pyne/material.pyx
@@ -345,6 +345,14 @@ cdef class _Material:
         cdef std_string card
         card = self.mat_pointer.mcnp(frac_type.encode())
         return card.decode()
+    
+    def mcnp(self):
+        """mcnp(frac_type)
+        Return a uwuw material name
+        """
+        cdef std_string uwuw_name
+        uwuw_name = self.mat_pointer.uwuw()
+        return uwuw_name.decode()
 
     def openmc(self, frac_type='mass'):
         """openmc(frac_type)
--- a/src/material.cpp
+++ b/src/material.cpp
@@ -594,6 +594,26 @@ std::string pyne::Material::openmc(std::string frac_type) {
   return oss.str();
 }
 
+///---------------------------------------------------------------------------//
+std::string pyne::Material::uwuw() {
+  // standard uwuw material name is : "mat:<Name of Material>/rho:<density>" 
+  if (! metadata.isMember("name")) {
+    pyne::warning("The material has no name!!");
+    return "";
+  }
+  std::string uwuw_name = "mat:<";
+  uwuw_name += metadata["name"].asString();
+  if (density > 0) {
+    uwuw_name += "/rho:" +std::to_string(density);
+  } else {
+    pyne::warning("No Density defined for this Material");
+  }
+
+  uwuw_name += ">";
+  return uwuw_name;
+}
+
+///---------------------------------------------------------------------------//
 std::string pyne::Material::mcnp(std::string frac_type) {
   //////////////////// Begin card creation ///////////////////////
   std::ostringstream oss;
--- a/src/material.cpp
+++ b/src/material.cpp
@@ -595,21 +595,22 @@ std::string pyne::Material::openmc(std::string frac_type) {
 }
 
 ///---------------------------------------------------------------------------//
-std::string pyne::Material::uwuw() {
+std::string pyne::Material::get_uwuw_name() {
   // standard uwuw material name is : "mat:<Name of Material>/rho:<density>" 
   if (! metadata.isMember("name")) {
     pyne::warning("The material has no name");
     return "";
   }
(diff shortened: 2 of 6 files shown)


pyne/material.pyx
```diff
@@ -345,6 +345,14 @@ cdef class _Material:
         cdef std_string card
         card = self.mat_pointer.mcnp(frac_type.encode())
         return card.decode()
+    
+    def mcnp(self):
```
pshriwise: The method name here is incorrect.
```suggestion
    def uwuw(self):
```

pyne/material.pyx
```diff
@@ -345,6 +345,14 @@ cdef class _Material:
...
+        """mcnp(frac_type)
```
pshriwise: Docstring should be updated as well. There is no `frac_type` argument and function name needs updating, same as above.

src/material.cpp
```diff
@@ -594,6 +594,26 @@ std::string pyne::Material::openmc(std::string frac_type) {
   return oss.str();
 }
 
+///---------------------------------------------------------------------------//
+std::string pyne::Material::uwuw() {
+  // standard uwuw material name is : "mat:<Name of Material>/rho:<density>" 
+  if (! metadata.isMember("name")) {
+    pyne::warning("The material has no name!!");
```
pshriwise:
```suggestion
    pyne::warning("The material has no name");
```

tests/test_material.py
```diff
@@ -1290,7 +1290,47 @@ def test_mcnp_mat0():
     assert_equal(mass, mass_exp)
 
 
+def uwuw():
```
pshriwise: This test won't be found by `nosetests`. It needs to have a name like `test_*`
```suggestion
def test_uwuw():
```

src/material.cpp
```diff
@@ -594,6 +594,26 @@ std::string pyne::Material::openmc(std::string frac_type) {
...
+  }
+  std::string uwuw_name = "mat:<";
+  uwuw_name += metadata["name"].asString();
+  if (density > 0) {
+    uwuw_name += "/rho:" +std::to_string(density);
+  } else {
+    pyne::warning("No Density defined for this Material");
+  }
+
+  uwuw_name += ">";
```
pshriwise:
```suggestion
```

src/material.cpp
```diff
@@ -594,6 +594,26 @@ std::string pyne::Material::openmc(std::string frac_type) {
...
 }
 
+///---------------------------------------------------------------------------//
+std::string pyne::Material::uwuw() {
+  // standard uwuw material name is : "mat:<Name of Material>/rho:<density>" 
+  if (! metadata.isMember("name")) {
+    pyne::warning("The material has no name!!");
+    return "";
+  }
+  std::string uwuw_name = "mat:<";
```
pshriwise:
```suggestion
  std::string uwuw_name = "mat:";
```

pshriwise (changes requested):
Good idea.

The format for names isn't quite right though. UWUW like the leading and trailing `<`/`>` characters.

tests/test_material.py
```diff
@@ -1290,7 +1290,47 @@ def test_mcnp_mat0():
...
+    leu2 = Material(nucvec={'U235': 0.04, 'U238': 0.96},
+                   metadata={'mat_number': 2,
+                          'table_ids': {'92235':'15c', '92238':'25c'},
+                          'mat_name':'LEU',
+                          'source':'Some URL',
+                          'comments': ('this is a long comment that will definitly '
+                                       'go over the 80 character limit, for science'),
+                          'name':'leu'})
+    uwuw_name = leu2.uwuw()
+    name_exp = ('<mat:leu>')
```
bam241:
```suggestion
    name_exp = ('mat:leu')
```

tests/test_material.py
```diff
@@ -1290,7 +1290,47 @@ def test_mcnp_mat0():
...
+                          'table_ids': {'92235':'15c', '92238':'25c'},
+                          'mat_name':'LEU',
+                          'source':'Some URL',
+                          'comments': ('this is a long comment that will definitly '
+                                       'go over the 80 character limit, for science'),
+                          'name':'leu'},
+                   density=19.1)
+
+    uwuw_name = leu.uwuw()
+    name_exp = ('<mat:leu/rho:19.1>')
```
bam241:
```suggestion
    name_exp = ('mat:leu/rho:19.1')
```

pyne/material.pyx
```diff
@@ -345,6 +345,14 @@ cdef class _Material:
         cdef std_string card
         card = self.mat_pointer.mcnp(frac_type.encode())
         return card.decode()
+    
+    def uwuw(self):
+        """mcnp(frac_type)
```
bam241:
```suggestion
        """uwuw()
```

bam241:
thx for the review @pshriwise and for the suggestion made my work so easy !!

gonuke (commented):
My only comment is about the name of the method.  `uwuw` doesn't communicate what it's doing very well.  What about `get_uwuw_name`?

(Note: in general, I have a preference for method names to be verbs, but there are already many that are not, so I can't impose it as a norm)

bam241:
> My only comment is about the name of the method. `uwuw` doesn't communicate what it's doing very well. What about `get_uwuw_name`?
> 
> (Note: in general, I have a preference for method names to be verbs, but there are already many that are not, so I can't impose it as a norm)

I am fine with that suggestion, but does it mean we should eventually change the `mcnp()` and the `fluka() ones ?

tests/test_material.py
```diff
@@ -1290,7 +1290,47 @@ def test_mcnp_mat0():
     assert_equal(mass, mass_exp)
 
 
+def test_uwuw():
+    leu = Material(nucvec={'U235': 0.04, 'U238': 0.96},
+                   metadata={'mat_number': 2,
+                          'table_ids': {'92235':'15c', '92238':'25c'},
+                          'mat_name':'LEU',
```
pshriwise: Help me out here. Why both `mat_name` and `name` in the material def?

bam241: this is actually a good question, I copy pasted the one form `mcnp()` test... I should probably fix both of them !

bam241: looking at the source code of material.cpp, it seems that openmc is using the `mat_name` metadata to name the material where all the others uses `name`.

Here is what I suggest:
In this PR: fix my test about `uwuw` removing mat_name
Issue an issue: to remove mention to mat_name in `openmc()` and in the test_material.py 
@pshriwise what do you thing ?

Shall we also implement something to catch old `mat_name` and convert them into name ? (for backward compatibility)

pshriwise: That sounds like a good plan to me!

gonuke:
> > My only comment is about the name of the method. `uwuw` doesn't communicate what it's doing very well. What about `get_uwuw_name`?
> > (Note: in general, I have a preference for method names to be verbs, but there are already many that are not, so I can't impose it as a norm)
> 
> I am fine with that suggestion, but does it mean we should eventually change the `mcnp()` and the `fluka() ones ?

Perhaps - but they are qualitatively different.  This new method extracts a single string out of the metadata in a particular format.  The others are a little more comprehensive (although still not verbs).

bam241:
@pshriwise  ?

pshriwise (approved):

pshriwise:
Will merge this evening if there are no further comments. Thanks @bam241!

gonuke:
Thanks @bam241

PR #1191, opened 2019-10-21
openMC material name 

### LINKED ISSUE #1189: Material metdata: mat_name vs name
Today with @pshriwise we spotted that in Material.cpp the openmc() methods expect the material name to be in the metadata as `mat_name` where all the other method uses `name`.

For consistency this has to be fixed.

But shall we break backward compatibility and suppress mention of `mat_name` or try to find a fix for it ?

bam241:
openmc() method in material.cc was looking for the "mat_name" for the name of the material.

this enforce the metadata convention: `name` for the material name.

I added some method to still catch `mat_name` if present in order to allow backward compatibility.

this should partially fix #1189

### DIFF
--- a/src/material.cpp
+++ b/src/material.cpp
@@ -502,15 +502,10 @@ std::string pyne::Material::openmc(std::string frac_type) {
   }
 
   // add name if specified
-  if (temp_mat.metadata.isMember("mat_name")) {
-    std::cout << "Usage of mat_name ad a Metadata is deprecated, please use \"name\" instead." <<std::endl;
-    temp_mat.metadata["name"] = temp_mat.metadata["mat_name"].asString();
-  }
 
   if (temp_mat.metadata.isMember("name")) {
     oss << "name=" << new_quote << temp_mat.metadata["name"].asString() << end_quote;
-  }
-
+  } 
   // close the material tag
   oss << ">";
   // new line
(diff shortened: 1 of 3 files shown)


src/material.cpp
```diff
@@ -503,7 +503,12 @@ std::string pyne::Material::openmc(std::string frac_type) {
 
   // add name if specified
   if (temp_mat.metadata.isMember("mat_name")) {
-    oss << "name=" << new_quote << temp_mat.metadata["mat_name"].asString() << end_quote;
+    std::cout << "Usage of mat_name ad a Metadata is deprecated, please use \"name\" instead." <<std::endl;
```
pshriwise: I don't think we need this output and update of the value. I simply wasn't using the right value in the first place and I'm hoping that this capability hasn't been around for long enough to cause a backwards compatibility problem.

Additionally, what if the metadata has both a `name` and `mat_name`? We'd overwrite the exisitng, and possibly preferred, infromation.

pshriwise (changes requested):
Thanks for taking care of this @bam241. Just one comment about the additional output and 'mat_name' &rarr; `name` update.

src/material.cpp
bam241: I removed the search for `mat_name` as advised.
thx for the review !

pshriwise (approved):

PR #1193, opened 2019-11-12
Fix variable issue for mcnp6 source file

ngranda:
This pull request removes two variables that are not needed with the latest changes. These variables cause errors to come up when building the source. 
1. `icl_tmp`: has two declarations. Kept the one with intent declaration 
2. `find_cell`: has integer declaration, but `find_cell` is a subroutine now

### DIFF
--- a/news/fix_variable_issue_mcnp6_source.rst
+++ b/news/fix_variable_issue_mcnp6_source.rst
@@ -1,0 +1,14 @@
+
+**Added:** None
+
+**Changed:**
+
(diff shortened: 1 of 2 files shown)


gonuke:
This LGTM, but it needs a news file to describe the update.

gonuke:
Thanks @ngranda - this was tested in MCNP5, but not in MCNP6 which is probably why the difference.  Nice catch!

Issue #1195, opened 2019-12-04
Multi-particles tally definition


bam241:
I want/need to allow the `PyNE::Tally` to deal with multiple particles tallies.

It will ultimately only change the `std::string particle_name` into a `std::vector<std::string>` particles_name
(for the tally information data structure)

My problem is in the `write_hdf5` and `read_hdf5`, that now it is basically forming and writing:
```
typedef struct tally_struct {
    int entity_id;
    int entity_type;
    int tally_type;
    const char* particle_name;
    const char* entity_name;
    const char* tally_name;
    double entity_size;
    double normalization;
  } tally_struct;
```
into the hdf5 file.
I could probably replace the `const char*` with a `vector<const char*>` but I am not sure how to maintain backward compatibility.

I was thinking about concatenating the list of particle names into a single `string` with comma separator, allowing to write down a list of particle as a single `const char*`

Does it make sense ? What do you think ?

I wanted to have your though about this before I start the implementation of this part...

gonuke:
closed by #1196

PR #1196, opened 2019-12-04
Multi particle tally

bam241:
This should allow to define tally tallying multiple particles at once for `MCNP`.

This would revolve #1195 .

I am still not sure my implementation is correct, please focus on the `C++` part for the moment.

This is not yet working (still working on fixing everything)...
but I want to be sure the proposed implementation is OK.


bam241:
the existing version of `particle_name` is a `std::string` that I replaced with a `std::vector<std::string>`

When writing an `hdf5`file, I concatenated the `vector<string>` into a single string with `,` delimiters: `{"part1", "part2", "part3"}` -> `"part1,part2,part3"` and used in place of the previous `particle_name`
When reading an `hdf5` file, it splits the particle name using a `,` delimiter and fill out the vector<string>.

This should allow backward compatibility.

pyne/tally.pyx
```diff
@@ -385,15 +397,18 @@ cdef class Tally:
...
-    property particle_name:
-        """no docstring for particle_name, please file a bug report!"""
+    property particle_names:
+        """no docstring for particle_names, please file a bug report!"""
         def __get__(self):
-            return bytes(<char *> (<cpp_tally.Tally *> self._inst).particle_name.c_str()).decode()
+            proxy_value = []
+            for i in range(len((<cpp_tally.Tally *> self._inst).particle_names)):
+                b_value = bytes(<char *> (<cpp_tally.Tally *>
+                     self._inst).particle_names[i].c_str()).decode()
```
gonuke, line 457: There is probably a more readable place to break the line here?

src/tally.h
```diff
@@ -52,6 +52,20 @@ namespace pyne
           std::string entity_type, std::string entity_name,
           std::string tally_name = "", double entity_size = 0.0, 
           double normalization = 1.0);
+    /// Constructor for regular tally
```
gonuke, line 56: insert a blank line

src/utils.cpp
```diff
@@ -300,17 +300,40 @@ std::string pyne::natural_naming(std::string name) {
...
+    output_vector.push_back(lists.substr(prev_pos, pos));
+    prev_pos = pos + delimiter.length();
+  }
+  return output_vector;
+}
+
+
+
 template<typename T>
 std::string pyne::join_to_string(std::vector<T> vect, std::string delimiter){
```
gonuke, line 324: We should add tests for this

gonuke (changes requested):
This looks good to me - a few minor comments.

Do you check that multi-particle tallies are supported for each of the MCNP tally types you have?

src/utils.cpp
bam241: yes 💯 
I am not done yet :) still segfaulting :)

bam241:
these is still missing testing!

\***edit**\* this is also missing real implementation of multi particles constructor, appending particles

bam241:
this should be good to break reviewed !

src/utils.cpp
bam241: fixed

tests/test_tally.py
```diff
@@ -396,6 +383,18 @@ def test_tally37():
     new_tally.from_hdf5("test_tally.h5","bob_geldof",1)
     assert_equal(tally.tally_type,new_tally.tally_type)
 
+# test multi-particule tally
```
gonuke:
```suggestion
# test multi-particle tally
```

pyne/tally.pyx
```diff
@@ -160,6 +170,42 @@ cdef class Tally:
         tal_name_bytes = tal_name.encode()
         self._inst = new cpp_tally.Tally(std_string(<char *> type_bytes), std_string(<char *> part_name_bytes), <int> ent, std_string(<char *> ent_type_bytes), std_string(<char *> ent_name_bytes), std_string(<char *> tal_name_bytes), <double> size, <double> norm)
 
+
+    def _tally_tally_1b(self, type, part_names, ent, ent_type, ent_name, tal_name='', size=0.0, norm=1.0):
+        """Tally(self, type, part_names, ent, ent_type, ent_name, tal_name='', size=0.0, norm=1.0)
+         This method was overloaded in the C-based source. To overcome
+        this we ill put the relevant docstring for each version below.
```
gonuke:
```suggestion
        this we will put the relevant docstring for each version below.
```

pyne/tally.pyx
```diff
@@ -261,7 +316,7 @@ cdef class Tally:
     def __init__(self, *args, **kwargs):
         """Tally(self, type, part_name, ent, ent_type, ent_name, tal_name='', size=0.0, norm=1.0)
          This method was overloaded in the C-based source. To overcome
-        this we ill put the relevant docstring for each version below.
+        this we ilL put the relevant docstring for each version below.
```
gonuke, line 319:
```suggestion
        this we will put the relevant docstring for each version below.
```

pyne/tally.pyx
```diff
@@ -385,15 +448,20 @@ cdef class Tally:
             (<cpp_tally.Tally *> self._inst).normalization = <double> value
 
 
-    property particle_name:
-        """no docstring for particle_name, please file a bug report!"""
+    property particle_names:
+        """no docstring for particle_names, please file a bug report!"""
```
gonuke: Why don't you write a docstring for this?

src/tally.cpp
```diff
@@ -536,32 +558,32 @@ std::string pyne::Tally::form_mcnp_meshtally(
   } else if (entity_geometry.find("Cylinder") != std::string::npos) {
     mtally_stream << "CYL" << std::endl;
     if (!is_zero(axs)) {
-      mtally_stream << indent_block << "AXS=" << pyne::join_to_string(axs) << std::endl;
+      mtally_stream << indent_block << "AXS= " << pyne::join_to_string(axs) << std::endl;
```
gonuke: Why this space?  I think these are better without the space.

tests/test_tally.py
```diff
@@ -278,18 +277,6 @@ def test_tally28():
 
     assert_equal(tally.entity_size,new_tally.entity_size)
 
-def test_tally28():
```
gonuke, line 281: Why did we get rid of this test?

src/utils.cpp
gonuke: Do we have tests for this yet?

gonuke (changes requested):
A few minor edits and some questions

pyne/tally.pyx
bam241: this is how it was before, I don't know why there is not doctoring there on this one or the others....

src/tally.cpp
bam241: I changed the join_to_string method, to not have the delimiter before the first joined item...
the test included the space, so I added it back here.

I will remove that one and fix the test accordingly

tests/test_tally.py
bam241: my mistake, this was not intended

bam241:
@gonuke, about the `pyne::join_to_string` we don't have a dedicated test.
but I am wondering if we have test for utils.cpp method ?

I can probably add some test for `joint_to_string` and `split_string`.

tests/test_tally.py
bam241: Actually, it was an empty test that was just hanging there. (the real `test_tally28()` is defined L268)

I just cleaned this up....

bam241 (commented):
> @gonuke, about the `pyne::join_to_string` we don't have a dedicated test.
> but I am wondering if we have test for utils.cpp method ?
> 
> I can probably add some test for `joint_to_string` and `split_string`.

@gonuke we don't have a direct translation of those in the Python API, so we don't test them individually.

If we want to test them directly we will have expose them in the Python API

gonuke:
> > @gonuke, about the `pyne::join_to_string` we don't have a dedicated test.
> > but I am wondering if we have test for utils.cpp method ?
> > I can probably add some test for `joint_to_string` and `split_string`.
> 
> @gonuke we don't have a direct translation of those in the Python API, so we don't test them individually.
> 
> If we want to test them directly we will have expose them in the Python API

I forgot that we don't test any C++ directly!  I don't think we should add these to the python API.  At some time (not for this PR) we should consider whether we add testing in C++ space.

pyne/tally.pyx
gonuke: I see that they are all that way... I guess I feel like it makes sense to correct it when the opportunity presents itself - and this might be such an opportunity

src/tally.cpp
```diff
@@ -552,7 +575,7 @@ std::string pyne::Tally::form_mcnp_meshtally(
     mtally_stream << dir_name[j] << "MESH=" << pyne::join_to_string(meshes[j]);
     if (ints[j].size() > 0) {
       mtally_stream << " " << dir_name[j]
-                    << "INTS=" << pyne::join_to_string(ints[j]);
+                    << "INTS= " << pyne::join_to_string(ints[j]);
```
gonuke:
```suggestion
                    << "INTS=" << pyne::join_to_string(ints[j]);
```

src/utils.cpp
gonuke: Resolving this because we don't currently support any direct C++ testing and these functions are not exposed through the python API

tests/test_tally.py
gonuke: It doesn't seem completely empty... but I'm not sure if/how it works...

tests/test_tally.py
```diff
@@ -483,14 +483,14 @@ def test_mcnp_mesh_tally_cyl():
...
-        "          ORIGIN= 1.000000 2.000000 3.000000\n"+\
-        "          IMESH= 5.000000 10.000000 20.000000 25.000000 IINTS= 1 2 3 1\n"+\
-        "          JMESH= 2.000000 12.000000 JINTS= 1\n"+\
-        "          KMESH= 45.000000 KINTS= 1\n"+\
-        "          EMESH= 0.000000 10.000000 100.000000\n"+\
-        "          EINTS= 1 1 2";
+        "          AXS=-1.000000 4.000000 -2.000000\n"+\
+        "          VEC=12.000000 -2.000000 5.000000\n"+\
+        "          ORIGIN=1.000000 2.000000 3.000000\n"+\
+        "          IMESH=5.000000 10.000000 20.000000 25.000000 IINTS= 1 2 3 1\n"+\
```
gonuke:
```suggestion
        "          IMESH=5.000000 10.000000 20.000000 25.000000 IINTS=1 2 3 1\n"+\
```

tests/test_tally.py
```diff
@@ -483,14 +483,14 @@ def test_mcnp_mesh_tally_cyl():
...
+        "          JMESH=2.000000 12.000000 JINTS= 1\n"+\
```
gonuke:
```suggestion
        "          JMESH=2.000000 12.000000 JINTS=1\n"+\
```

tests/test_tally.py
```diff
@@ -483,14 +483,14 @@ def test_mcnp_mesh_tally_cyl():
...
+        "          KMESH=45.000000 KINTS= 1\n"+\
```
gonuke:
```suggestion
        "          KMESH=45.000000 KINTS=1\n"+\
```

tests/test_tally.py
bam241: yes you are write... sorry my bad, 2 tests had same labels, and I overlook at them to quickly... I put it back....

bam241: maybe we should rename the second one tho....

gonuke (approved):

pyne/tally.pyx
gonuke: Should I wait for a docstring?

gonuke:
Do you need to rebase against the PR that changed "ill" to "will"?

pyne/tally.pyx
bam241: yes please !

bam241:
I don't think so I did not include the tally one in the "ill" PR so it would not conflict

bam241:
@gonuke I have added a docstring, I was not sure how a `property` should be formatted, let me know if I need to change it!

gonuke:
Thanks @bam241 - Merging now

PR #1197, opened 2019-12-06
CI tests add OpenMC API as optional dependency

zxkjack123:
This PR adds CI tests with OpenMC Python API as an optional dependency.
This PR is part of work for supporting OpenMC and DagOpenMC related functions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,6 +106,7 @@ commands:
       - run: |
           cd ~/repo/tests
           ./travis-run-tests.sh << parameters.flags >>
+
   # Build and push PyNE website
   website_build_push:
     description: "build PyNE website and push it either on the test branch (default) or on the deployed branch"
@@ -129,6 +130,7 @@ commands:
               cd docs/
               make html
               make push-<< parameters.push_option >>
+
 # Jobs part:
 # Define the different job that will be ran this separate building form
 # testing for each configuration allowing to get more information out of the CI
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -103,7 +106,6 @@ commands:
       - run: |
           cd ~/repo/tests
           ./travis-run-tests.sh << parameters.flags >>
-
```
bam241: maybe keep this line break to visually separate different commands

.circleci/config.yml
```diff
@@ -127,7 +129,6 @@ commands:
               cd docs/
               make html
               make push-<< parameters.push_option >>
-
```
bam241: I think we should keep 2 line breaks between different sections.

bam241 (approved):
LGTM. Thx @zxkjack123 !
2 minor things...

Just wondering, does some openMC capabilities get disable if openM is not present in the build ?

kkiesling (approved):
Looks pretty straight forward and good to me! I'll leave it up to @bam241 for the final seal of approval since he is more versed in CI than me.

zxkjack123:
> LGTM. Thx @zxkjack123 !
> 2 minor things...
> 
> Just wondering, does some openMC capabilities get disable if openM is not present in the build ?

Firstly, I should make it clear, the dependency is OpenMC Python API (a Python module), but not OpenMC. OpenMC Python API, as an optional dependency, works similar to other optional dependencies, such as PyMOAB.

It will report an error/warning message when users call a function using OpenMC Python API. But it is there when the uses build it.

gonuke (approved):
LGTM - not sure I understand how this will behave differently, though - I'll let @bam241 merge

zxkjack123:
@bam241 Any further comments about this PR?

bam241:
Looks good thx @zxkjack123 ! Merging!

PR #1199, opened 2019-12-16
Move meshtally to mesh

zxkjack123:
This PR move the `MeshTally` class from `mcnp.py` to `mesh.py`. The `MeshTally` now is regarded as a independent mesh tally container, but not part of  `Meshtal` of MCNP.

The mesh tally data from OpenMC will be supported (in next several PRs) with the modified `MeshTally`. Mesh parameter and flux data will be read from functions in `mcnp.py` or `openmc_utils.py`, and then put to an instance of `MeshTally`.


pyne/mcnp.py
```diff
@@ -1998,202 +1998,223 @@ def _read_tallies(self, f):
...
-        self.tally_number = tally_number
-        self._read_meshtally_head(f)
-        self._read_column_order(f)
 
+        Returns:
+        --------
+        m : MeshTally object
+            The MeshTally object created from MCNP mesh tally of tally_number.
+        """
+        m = MeshTally()
```
gonuke, line 2031: Please add a blank line after docstring for readability - perhaps also in other functions, even though some of the original functions didn't have this extra line.

gonuke (changes requested):
Thanks @zxkjack123 - this LGTM - just a small formatting request

pyne/mcnp.py
zxkjack123: All the docstrings in `mesh.py` and `mcnp.py` are formatted according to PEP257:https://www.python.org/dev/peps/pep-0257/

Major changs:
- Add blank lines after mult-line docstrings
- Shorten some docstrings into one-line docstrings. (No blank line after one-line docstirngs).

gonuke (approved):

gonuke:
LGTM, but I wonder how it interacts with some of the current work by @ngranda ?

pyne/mcnp.py
```diff
@@ -750,8 +750,7 @@ def __init__(self, filename):
         self.read()
 
     def read(self):
-        """Populate the Xsdir object by reading the file.
-        """
+        """Populate the Xsdir object by reading the file."""
```
pshriwise:
```suggestion
        """Populate the xsdir object by reading the file."""
```

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
+        # assign particle and dose_response
+        m.particle, m.dose_response = self.read_meshtally_head(f)
+        # assign tag_names
         if tag_names is None:
-            self.tag_names = ("{0}_result".format(self.particle),
-                              "{0}_result_rel_error".format(self.particle),
-                              "{0}_result_total".format(self.particle),
-                              "{0}_result_total_rel_error".format(self.particle))
+            m.tag_names = ("{0}_result".format(m.particle),
+                              "{0}_result_rel_error".format(m.particle),
```
pshriwise: Indentation is a little off here.

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
+        m.x_bounds, m.y_bounds, m.z_bounds, m.e_bounds = \
+                self.read_xyze_bounds(f)
+        m.num_ves = (len(m.x_bounds)-1) * (len(m.y_bounds)-1)\
+            * (len(m.z_bounds)-1)
+        m.num_e_groups = len(m.e_bounds)-1
+        column_idx = self.read_column_order(f)
+        # create mesh
+        mats = () if mesh_has_mats is True else None
+        super(MeshTally, m).__init__(structured_coords=[m.x_bounds,
+            m.y_bounds, m.z_bounds], structured=True, mats=mats)
```
pshriwise: This looks a little awkward. I'd go with a hanging indent.

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
+
+        # read result, rel_error, res_tot and rel_err_tot
+        result, rel_error, res_tot, rel_err_tot = \
+                self.read_tally_results_rel_error(f, m.num_e_groups, m.num_ves,
```
pshriwise: Same comment on hanging indent here.

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
-                              "{0}_result_total_rel_error".format(self.particle))
+            m.tag_names = ("{0}_result".format(m.particle),
+                              "{0}_result_rel_error".format(m.particle),
+                              "{0}_result_total".format(m.particle),
+                              "{0}_result_total_rel_error".format(m.particle))
         else:
-            self.tag_names = tag_names
-
-        self._create_mesh(f, mesh_has_mats)
+            m.tag_names = tag_names
```
pshriwise, line 2044: Should we do some type-checking to make sure we're getting an Iterable of 4 strs here?

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
+        z_bounds = [float(x) for x in f.readline().split()[2:]]
         # "Energy bin boundaries" contain one more word than "X boundaries"
-        self.e_bounds = [float(x) for x in f.readline().split()[3:]]
+        e_bounds = [float(x) for x in f.readline().split()[3:]]
+        return x_bounds, y_bounds, z_bounds, e_bounds
 
-        self.dims = [0, 0, 0] + [len(self.x_bounds) - 1,
-                                 len(self.y_bounds) - 1,
-                                 len(self.z_bounds) - 1]
+    def read_column_order(self, f):
```
pshriwise, line 2133: This method doesn't use `self` and can be made a static method.

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
+        column_idx = dict(zip(column_names, range(0, len(column_names))))
+        return column_idx
 
-    def _create_mesh(self, f, mesh_has_mats):
-        """Instantiate a Mesh object and tag the PyMOAB core instance
-           with results and relative errors.
+    def read_tally_results_rel_error(self, f, num_e_groups, num_ves,
+            column_idx):
         """
+        Read meshtally results and reletive error data.
```
pshriwise:
```suggestion
        Read meshtally results and relative error data.
```

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
-                                                           self.y_bounds, self.z_bounds],
-                                        structured=True, mats=mats)
-
-        num_vol_elements = (len(self.x_bounds)-1) * (len(self.y_bounds)-1)\
-            * (len(self.z_bounds)-1)
-        num_e_groups = len(self.e_bounds)-1
+        Returns
+        -------
+        results : numpy array
+            Tally results data, shape=(num_ves, num_e_groups)
```
pshriwise: I'd put these shape comments in the line above next to the type.

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
         # get result and relative error data from file
-        result = np.empty(shape=(num_e_groups, num_vol_elements))
-        rel_error = np.empty(shape=(num_e_groups, num_vol_elements))
+        result = np.empty(shape=(num_e_groups, num_ves))
+        rel_error = np.empty(shape=(num_e_groups, num_ves))
         for i in range(0, num_e_groups):
             result_row = []
             rel_error_row = []
-            for j in range(0, num_vol_elements):
+            for j in range(0, num_ves):
```
pshriwise:
```suggestion
            for j in range(num_ves):
```

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
+        rel_err_tot : numpy array
+            Total relative error fo flux.
+        """
 
         # get result and relative error data from file
-        result = np.empty(shape=(num_e_groups, num_vol_elements))
-        rel_error = np.empty(shape=(num_e_groups, num_vol_elements))
+        result = np.empty(shape=(num_e_groups, num_ves))
+        rel_error = np.empty(shape=(num_e_groups, num_ves))
         for i in range(0, num_e_groups):
```
pshriwise:
```suggestion
        for i in range(num_e_groups):
```

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
             result_row = []
             rel_error_row = []
-            for j in range(0, num_vol_elements):
+            for j in range(0, num_ves):
                 line = f.readline().split()
-                result_row.append(float(line[self._column_idx["Result"]]))
+                result_row.append(float(line[column_idx["Result"]]))
```
pshriwise: I thought `column_idx` was being stored on the class? But here it's passed in.

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
-
+        res_tot = []
+        rel_err_tot = []
         # If "total" data exists (i.e. if there is more than
         # 1 energy group) get it and tag it onto the mesh.
         if num_e_groups > 1:
-            result = []
-            rel_error = []
-            for i in range(0, num_vol_elements):
+            for i in range(0, num_ves):
```
pshriwise:
```suggestion
            for i in range(num_ves):
```

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
-            rel_err_tot_tag = self.get_tag(self.tag_names[3])
-
-            res_tot_tag[:] = result
-            rel_err_tot_tag[:] = rel_error
+                res_tot.append(float(line[column_idx["Result"]]))
+                rel_err_tot.append(
+                    float(line[column_idx["Rel_Error"]]))
+        else:
+            res_tot = result.flatten()
+            rel_err_tot = rel_error.flatten()
```
pshriwise, line 2203: It looks like these lines could cause an error if `num_e_groups > 1`.

`res_tot` and `rel_err_tot` are initialized as empty lists. If we get this point we'll get an `AttributeError` when calling `flatten` I think.

pyne/mesh.py
```diff
@@ -1510,6 +1511,102 @@ def _do_op(self, other, tags, op, in_place=True):
 
         return mesh_1
 
+
+class MeshTally(StatMesh):
+    """This class stores all information from all single mesh tally that
```
pshriwise:
```suggestion
    """This class stores all information from a single mesh tally that
```

pyne/mesh.py
```diff
@@ -1510,6 +1511,102 @@ def _do_op(self, other, tags, op, in_place=True):
...
+    exists within some meshtal or state point file. Header information is
+    stored as attributes and the "mesh" attribute is a MOAB mesh with all
```
pshriwise: Technically a PyNE Mesh.
```suggestion
    stored as attributes and the "mesh" attribute is a PyNE mesh object with all
```

pyne/mesh.py
```diff
@@ -1510,6 +1511,102 @@ def _do_op(self, other, tags, op, in_place=True):
...
+    result and relative error data tagged. This class inherits from StatMesh,
+    exposing all statistical mesh manipulation methods.
```
pshriwise: Overall I'd update this docstring to

```
... "mesh" attribute is a PyNE mesh object tagged with all result and relative error data. This class ....
```

pyne/mesh.py
```diff
@@ -1510,6 +1511,102 @@ def _do_op(self, other, tags, op, in_place=True):
...
+    ----------
+    tally_number : int
+        The tally number.
+        For mesh tally from MCNP, it must end with 4 (e.g. 4, 14, 214).
+        For mesh tally from OpenMC, it could be any int.
+    particle : string
+        Either "neutron" for a neutron mesh tally or "photon" for a photon mesh
+        tally.
+    dose_response : bool
+        True is the tally is modified by a dose response function.
```
pshriwise:
```suggestion
        True if the tally is modified by a dose response function.
```

pyne/mesh.py
```diff
@@ -1510,6 +1511,102 @@ def _do_op(self, other, tags, op, in_place=True):
...
+    x_bounds : list of floats
+        The locations of mesh vertices in the x direction.
+    y_bounds : list of floats
+        The locations of mesh vertices in the y direction.
+    z_bounds : list of floats
+        The locations of mesh vertices in the z direction.
+    dims : list
+        Dimensions of the mesh.
```
pshriwise, line 1539: I'm wondering if some of these types should really be `tuple`s as we generally don't expect them to change.

pyne/mesh.py
```diff
@@ -1510,6 +1511,102 @@ def _do_op(self, other, tags, op, in_place=True):
...
+    num_ves : int
+        Number of volume elements.
+    e_bounds : list of floats
```
pshriwise, line 1542: Same comment about `tuple` vs. `list` here.

pyne/mesh.py
```diff
@@ -1510,6 +1511,102 @@ def _do_op(self, other, tags, op, in_place=True):
...
+        The minimum and maximum bounds for energy bins.
+    num_e_groups: int
+        Number of energy groups.
+    mesh :
+        An PyMOAB core instance tagged with all results and
+        relative errors.
+    tag_names : iterable
+        Four strs that specify the tag names for the results, relative errors,
+        total results, and relative errors of the total results.
```
pshriwise:
```suggestion
        total results, and total relative error.
```

pyne/mesh.py
```diff
@@ -1510,6 +1511,102 @@ def _do_op(self, other, tags, op, in_place=True):
...
+    Notes
+    -----
+    All Mesh/StatMesh attributes are also present via a super() call to
+    StatMesh.__init__().
+    """
+
+    def __init__(self):
+        """Create an empty MeshTally object and set default values."""
+        if not HAVE_PYMOAB:
+            raise RuntimeError("PyMOAB is not available, "
```
pshriwise: This should maybe raise `NotImplemented`? Not :100: on that though.

tests/test_r2s.py
```diff
@@ -145,6 +145,7 @@ def test_photon_sampling_setup_structured():
     m = Mesh(structured=True, structured_coords=coords)
     tags = {(10010000, "1 h"): "tag1", ("TOTAL", "shutdown"): "tag2"}
     photon_sampling_setup(m, phtn_src, tags)
+    os.remove(phtn_src+".h5")
```
pshriwise:
```suggestion
    os.remove(phtn_src + ".h5")
```

tests/test_r2s.py
```diff
@@ -300,6 +301,7 @@ def test_photon_sampling_setup_unstructured():
     m.structured = False
     tags = {(10010000, "1 h"): "tag1", ("TOTAL", "shutdown"): "tag2"}
     photon_sampling_setup(m, phtn_src, tags)
+    os.remove(phtn_src+".h5")
```
pshriwise:
```suggestion
    os.remove(phtn_src + ".h5")
```

pshriwise (changes requested):
This looks like a considerable effort @zxkjack123! It's overall a good design change to make this class more generic IMO. 

A couple of thoughts:

  - I wasn't clear on the status of `column_idx`. Docs say it's an attribute, but it's passed into other methods. Let's make the docs and code consistent.

  - ~~I think it would be useful to turn the `Meshtal.create_meshtally` method into a `MesTtally.from_meshtal` or `Meshtally.from_mcnp` classmethod with the logic for parsing the MCNP meshtal file text as part of MeshTally. Then, in the future, support for an OpenMC Meshtally could be added via a `MeshTally.from_openmc` method.~~ Removed after comment from @gonuke below.

gonuke:
> * I think it would be useful to turn the `Meshtal.create_meshtally` method into a `MesTtally.from_meshtal` or `Meshtally.from_mcnp` classmethod with the logic for parsing the MCNP meshtal file text as part of MeshTally. Then, in the future, support for an OpenMC Meshtally could be added via a `MeshTally.from_openmc` method.

The design here intentionally separates these, on my advice... but we can discuss.  I actually think is a good design for many code-specific “decorations” of common functionality.  All MCNP specific stuff can live in `mcnp.py` but referring to common infrastructure - in this case `MeshTally` infrastructure.  As it is, PyNE has a mixture, where the current `MeshTally` is a good example of common infrastructure living in a code-specific file, and `nucname` being the opposite where all code-specific code is in the common file.

pshriwise:
@gonuke ah, yeah, that's true. Kind of a trade-off there. In general, I tend to favor that style of object generation when possible (too much so, maybe). To me, it boils down to a dependency issue as opposed to the location of code-specific logic. While the logic for MCNP doesn't have any external dependencies, adding a `from_openmc` would insert a dependency on OpenMC in `mesh.py`. Not a desirable situation. 

In the end, I think it should say as-is. I'll update my review comment accordingly.

pyne/mcnp.py
```diff
@@ -1990,210 +1987,235 @@ def _read_meshtal_head(self, f):
...
-        f : filestream
-            Open to the neutron/photon line.
+        Parameters:
+        -----------
+        f : str or filestream
+            Filestream of the meshtal file.
         tally_number : int
             The MCNP fmesh4 tally number (e.g. 4, 14, 24).
+        particle : str
+            The particle type, 'neutron' or 'photon'.
```
ngranda: Is this an attribute here or an attribute of the MeshTally class?

ngranda (commented):
LGTM! Thanks @zxkjack123

ngranda:
> LGTM, but I wonder how it interacts with some of the current work by @ngranda ?

I use the Meshtal class in the same way that the r2s script uses it. I do not think these changes would affect this part of the work. For reading the rnucs mesh, I made my own class that is similar to the Meshtal and MeshTally classes but do not depend on either of them.

pyne/mcnp.py
zxkjack123: The type-checking for `tag_names` are added in https://github.com/pyne/pyne/pull/1199/commits/3d21bf4b7ce522fc31dd1f06d217edecb6bf55e9.

pyne/mcnp.py
zxkjack123: Changed in https://github.com/pyne/pyne/pull/1199/commits/86db650ff475201d4cf61d1e8ea1c04015bdaf98.

pyne/mcnp.py
zxkjack123: I kept the `self` here, for one thing, there are two `read_xxx` functions that uses `self`. For another thing, all there readers operates the same `meshtal` file. I think it's better to keep them linked to the same object (`self`).

pyne/mcnp.py
zxkjack123: Yes, `column_idx` should be stored on the class. Modified in https://github.com/pyne/pyne/pull/1199/commits/c2416ec0260a6d04d7a3eb22d696b2160ec8b5ad.

pyne/mcnp.py
zxkjack123: Well, it is the `result` that calls `flatten()`, but not `res_tot`.
The `result` is initialized by `result = np.empty(shape=(num_e_groups, num_ves))` and then filled with data.

pyne/mesh.py
zxkjack123: Updated in https://github.com/pyne/pyne/pull/1199/commits/e50076d1077c74b213305abdacbc9ceb5ff8e393.

pyne/mcnp.py
pshriwise: If we aren’t using `self` in the method to access or set attributes, it doesn’t change how the information is linked to self or how the file handle, `f` is managed. And that method would still maintain it’s context as a reader by nature of being part of the `Meshtal` class definition.

All that being said, `column_idx` is now one of the class attributes (a good update). So, yes, we should keep `self` here. 😃 

We’ll want to update the docstring of this method though as it still indicates that we’re returning a `dict`.

pyne/mesh.py
zxkjack123: The `x/y/z_bounds` may involve unit conversion. As far as I know, MCNP and OpenMC use `cm` as length unit. But I am not sure whether there is a MC code use `mm` or `m` as unit. And the mesh bounds transformation could also be used to support a moving component/mesh.

But `e_bounds` involves unit conversion, as MCNP uses `MeV` and OpenMC uses `eV`. So, `e_bounds` should be `list` rather than `tuple`.

pyne/mcnp.py
pshriwise: Ah, so true. I definitely didn’t read that correctly. Thanks!

Can we ensure that the `res_tot` and `res_err_tot` containers are consistent? In one case they will be lists, and in the other they will be NumPy arrays. Unless I’m missing something else!

pyne/mesh.py
zxkjack123: Thsi `RuntimeError` is replaced with `NotImplementedError` in https://github.com/pyne/pyne/pull/1199/commits/104e9daead65cf68dd2a9fc8f1bf5175bc8f8b90. This link maybe useful for this type of error: https://help.semmle.com/wiki/display/PYTHON/NotImplemented+is+not+an+Exception.

pyne/mcnp.py
zxkjack123: These two lines are deleted in https://github.com/pyne/pyne/pull/1199/commits/bae3bf1d7e03f507e8729edd3197c8daf20853e5, because it is not an input parameter. The `particle` is an attribute of the `MeshTally` class, but not the `Meshtal`. In theory, a `meshtal` file could contain meshtallies of different particles.

zxkjack123:
> > LGTM, but I wonder how it interacts with some of the current work by @ngranda ?
> 
> I use the Meshtal class in the same way that the r2s script uses it. I do not think these changes would affect this part of the work. For reading the rnucs mesh, I made my own class that is similar to the Meshtal and MeshTally classes but do not depend on either of them.

Glad that this PR does not conflicts with your work.

pyne/mcnp.py
zxkjack123: Docstring of this function is update in https://github.com/pyne/pyne/pull/1199/commits/bae3bf1d7e03f507e8729edd3197c8daf20853e5.

pyne/mcnp.py
zxkjack123: They are convert to `np.array` before return in https://github.com/pyne/pyne/pull/1199/commits/c8c6cdc1ece48782d7ec5a096de2bc3c069c97f0.

pyne/mcnp.py
```diff
@@ -2285,3 +2297,24 @@ def _mesh_to_mat_cards(mesh, divs, frac_type):
         mat_cards += mesh.mats[i].mcnp(frac_type=frac_type)
 
     return mat_cards
+
+def _check_tag_names(tag_names):
+    """Make sure tag_names is an iterable of 4 strings."""
+    # check iterable
+    if check_iterable(tag_names):
```
gonuke: This might be simpler as separate if statements that each raise an exception upon failure:

```
if not check_iteraable(tag_names):
    raise ValueError("....")

if len(tag_names) != 4:
    raise ValueError("....")

````
etc

**Review by gonuke — COMMENTED** [review]
One small suggestion, but not worth blocking the PR over this

**Comment by gonuke** [comment]
If @pshriwise is happy with your changes - we can probably merge

File: pyne/mcnp.py (hunk 26 above)
Comment (zxkjack123) [reply pyne/mcnp.py]: Thanks. Suggestion adopted in https://github.com/pyne/pyne/pull/1199/commits/f4b6dccff3290e3d0bb6e6a646e029199c95c759.

**Review by pshriwise — APPROVED** [review]
Looks ready to me. As usual, thanks for the diligence @zxkjack123!

File: pyne/mcnp.py [hunk 27]
```diff
@@ -2285,3 +2297,25 @@ def _mesh_to_mat_cards(mesh, divs, frac_type):
...
+    if len(tag_names) != 4:
+        raise ValueError("The length of tag_names is not 4.")
+
+    # check content strings
+    for item in tag_names:
+        if not isinstance(item, str):
+            raise ValueError("The content of tag_names ",
+                            "should be strings")
+
+    # tag_names should be a string with lenght of 4
```
gonuke:
```suggestion
    # tag_names should be a string with length of 4
```

gonuke:
Not sure why the tests are failing!??!

pshriwise:
Something weird with the config file?

This run has a lot of it commented: https://circleci.com/gh/zxkjack123/pyne/5152#config/containers/0

But the run from the commit before looks fine:
https://circleci.com/gh/zxkjack123/pyne/5151#config/containers/0

Not sure what the root cause is though. Not a circleci guru. @bam241?

gonuke:
Must have been a failure on the CircleCI end - I'll merge once the tests complete...

gonuke:
Thanks for all your effort here @zxkjack123

PR #1200, opened 2019-12-18
revert nucpath to origin value after testing internal material data

bam241:
Depending of the order in which the test runs it this might cause an issue, the other test would not be able to find the `NUC_DATA_PATH`


bam241:
I think it was intended that way as the value was stored at the beginning (but not reverted)

gonuke:
LGTM - thanks @bam241

PR #1205, opened 2020-01-17
Bug class Xsdir(object): def read corrected

Radiation-Transport:
The parsing of XSDIR file was uncorrected when the ZAID XS was defined in more than one row. Indeed, the "+" word was not removed from the list having for instance "+" instead of "p-table" value as in the example reported hereinafter.

	18038.80c 37.6366 xdata/endf71x/Ar/18038.710nc 0 1 4 15536 0 0 2.5301E-08 +
          ptable

### DIFF
--- a/pyne/mcnp.py
+++ b/pyne/mcnp.py
@@ -793,7 +793,7 @@ class Xsdir(object):
             # Handle continuation lines
             while words[-1] == '+':
                 extraWords = self.f.readline().split()
-                words = words + extraWords
+                words = words[:-1] + extraWords  # Correct bug in the XSDIR parsing
             assert len(words) >= 7
 
             # Create XsdirTable object and add to line


pyne/mcnp.py
```diff
@@ -793,7 +793,7 @@ def read(self):
             # Handle continuation lines
             while words[-1] == '+':
                 extraWords = self.f.readline().split()
-                words = words + extraWords
+                words = words[:-1] + extraWords  # Correct bug in the XSDIR parsing
```
pshriwise, line 796:
```suggestion
                words = words[:-1] + extraWords
```

pshriwise (changes requested):
Seems reasonable to me! Thanks for addressing this, @Radiation-Transport.

Just one small line change requested and to satisfy the PR requirements we'll want to add a news file here it seems.

@gonuke do you think an additional test case is in order here? I could take it or leave it.

gonuke (changes requested):
Thanks for finding this long-standing bug!

I do think it would be useful to have a test for this, ideally by adding lines with these characteristics in the `dummy_xsdir` files and adding a test to `test_mcnp.py` to ensure it is handled correctly.

Please also note that we require a brief news file to document the change based on `news/TEMPLATE.rst` to facilitate building change logs.

pshriwise:
@Radiation-Transport I opened a PR into your branch with the addition to our `test_mcnp` file that I think @gonuke was asking for. Once we get that change merged into your branch and take care of that small requested change above I think we'll be good to go.

gonuke:
@Radiation-Transport - we'd like to include this in an impending release.  Are you able to incorporate the requested changes, or should we take over the PR?

Radiation-Transport:
Dears,

Please go ahead as we would not like to be a bottleneck in your process,

Thanks for asking.

Cheers,
Marco

gonuke:
Thanks @Radiation-Transport - we'll do it in a way that preserves attribution of your valuable contribution.

gonuke:
This has been replaced by #1314 to wrap it up with contributions from @Radiation-Transport and @pshriwise

PR #1207, opened 2020-02-05
Material write hdf5 refactor

bam241:
this updates the write_hdf5 method to refactor the structure of material in an hdf5 file as such:

```
/material/
/material/mat_1/
/material/mat_1/nucid
/material/mat_1/composition
/material/mat_1/composition_metadata
```

`/material` and `/datapath` being HDF5 groups
 `/nucid`, `composition` and `composition_metadata` being HDF5 dataset

the value of `datapath` is string variable can/should be user provided.

this will allows a future MaterialLibrary (see #1104) to coexist with simple material inside the same hdf5 file

this is not done yet, still need some cleaning ad missing docstring.
I also need to tweak the `from_hdf5` method to be able to read the new structure

but for some reason `nosetests` are not finishing on my computer for this (or `master`....)

### DIFF
--- a/docs/pyapi/material.rst
+++ b/docs/pyapi/material.rst
@@ -20,7 +20,28 @@ have two main attributes which define them.
 1. **comp**: a normalized composition mapping from nuclides (zzaaam-ints) to mass-weights (floats).
 2. **mass**: the mass of the material.
 
+The :ref:`pyne_material_library` class is available to efficiently manipulate
+collections of materials.
 The material class is presented below.  For more information please refer to :ref:`usersguide_material`.
+
+When using the `write_hdf5` method, the default structure for the `HDF5` file
+is:
+.. verbatim::
+    /material/
+    /material/my_mat/
+    /material/my_mat/composition
+    /material/my_mat/nucid
+    /material/my_mat/composition_metadata
+
+
+Where, `/material` and `/material/my_mat` are HDF5 groups, (`my_mat` value shall provided
+as the datapath by the user).
+
+Older structure are still available to written when providing a `nucpath` to
+the `write_hdf5()` method.
+
+`from_hdf5()` will detect the structure (old or new) of the file (when using `protocol1`).
+
 
 **************
 Material Class
--- a/docs/pyapi/material.rst
+++ b/docs/pyapi/material.rst
@@ -39,9 +39,10 @@ Where, `/material` and `/material/my_mat` are HDF5 groups, (`my_mat` value shall
 be provided as the datapath by the user). IF the `$datapath` or `/material`
 exist as a dataset in the file, then the old writing method will ne used.
 
-Older structure are still available to written when providing a `nucpath` to the
-`write_hdf5()` method.
-Old structure looks like:
+Older data structure are deprecated but still available to written when providing a `nucpath` to the
+`write_hdf5()` method, or when writing a material in a file with the old
+data structure.
+Old data structure looks like:
 .. verbatim::
     /my_mat/
     /------/nucpath
--- a/docs/pyapi/material.rst
+++ b/docs/pyapi/material.rst
@@ -12,17 +12,17 @@ All functionality may be found in the ``material`` package::
 
  from pyne import material
 
-Materials are the primary container for radionuclides. They map nuclides to **mass weights**, 
-though they contain methods for converting to/from atom fractions as well.
-In many ways they take inspiration from numpy arrays and python dictionaries.  Materials
-have two main attributes which define them.
+Materials are the primary container for radionuclides. They map nuclides to
+**mass weights**, though they contain methods for converting to/from atom
+fractions as well.  In many ways they take inspiration from numpy arrays and
+python dictionaries.  Materials have two main attributes which define them.
 
-1. **comp**: a normalized composition mapping from nuclides (zzaaam-ints) to mass-weights (floats).
-2. **mass**: the mass of the material.
+1. **comp**: a normalized composition mapping from nuclides (zzaaam-ints) to
+   mass-weights (floats).  2. **mass**: the mass of the material.
 
 The :ref:`pyne_material_library` class is available to efficiently manipulate
-collections of materials.
-The material class is presented below.  For more information please refer to :ref:`usersguide_material`.
+collections of materials.  The material class is presented below.  For more
+information please refer to :ref:`usersguide_material`.
 
 When using the `write_hdf5` method, the default structure for the `HDF5` file
 is:
@@ -34,13 +34,15 @@ is:
     /material/my_mat/composition_metadata
 
 
-Where, `/material` and `/material/my_mat` are HDF5 groups, (`my_mat` value shall be provided
-as the datapath by the user).
+Where, `/material` and `/material/my_mat` are HDF5 groups, (`my_mat` value shall
+be provided as the datapath by the user). IF the `$datapath` or `/material`
+exist as a dataset in the file, then the old writing method will ne used.
 
-Older structure are still available to written when providing a `nucpath` to
-the `write_hdf5()` method.
+Older structure are still available to written when providing a `nucpath` to the
+`write_hdf5()` method.
 
-`from_hdf5()` will detect the structure (old or new) of the file (when using `protocol1`).
+`from_hdf5()` will detect the structure (old or new) of the file (when using
+`protocol1`).
 
 
 **************
--- a/docs/pyapi/material.rst
+++ b/docs/pyapi/material.rst
@@ -34,7 +34,7 @@ is:
     /material/my_mat/composition_metadata
 
 
-Where, `/material` and `/material/my_mat` are HDF5 groups, (`my_mat` value shall provided
+Where, `/material` and `/material/my_mat` are HDF5 groups, (`my_mat` value shall be provided
 as the datapath by the user).
 
 Older structure are still available to written when providing a `nucpath` to
--- a/docs/pyapi/material.rst
+++ b/docs/pyapi/material.rst
@@ -12,28 +12,28 @@ All functionality may be found in the ``material`` package::
 
  from pyne import material
 
-Materials are the primary container for mixtures of radionuclides. They map nuclides to
-**mass fractions**, though they contain methods for converting to/from atom
-fractions as well.  In many ways they take inspiration from numpy arrays and
-python dictionaries.  Materials have two main attributes which define them.
+Materials are the primary container for mixtures of radionuclides. They map
+nuclides to **mass fractions**, though they contain methods for converting
+to/from atom fractions as well.  In many ways they take inspiration from numpy
+arrays and python dictionaries.  Materials have two main attributes which define
+them.
 
 1. **comp**: a normalized composition mapping from nuclides (zzaaam-ints) to
    mass-fractions (floats).  
-   2. **mass**: the mass of the material.
+2. **mass**: the mass of the material.
 
 The :ref:`pyne_material_library` class is available to efficiently manipulate
 collections of materials.  The material class is presented below.  For more
 information please refer to :ref:`usersguide_material`.
 
-When using the `write_hdf5` method to write a material in a group named `my_mat`, the default structure for the `HDF5` file
-is:
+When using the `write_hdf5` method to write a material in a group named
+`my_mat`, the default structure for the `HDF5` file is:
 .. verbatim::
     /material/
-    /material/my_mat/
-    /material/my_mat/composition
-    /material/my_mat/nucid
-    /material/my_mat/composition_metadata
-
+    /--------/my_mat/
+             /------/composition
+             /------/nuclidelist
+             /------/composition_metadata
 
 Where, `/material` and `/material/my_mat` are HDF5 groups, (`my_mat` value shall
 be provided as the datapath by the user). IF the `$datapath` or `/material`
@@ -41,10 +41,23 @@ exist as a dataset in the file, then the old writing method will ne used.
 
 Older structure are still available to written when providing a `nucpath` to the
 `write_hdf5()` method.
+Old structure looks like:
+.. verbatim::
+    /my_mat/
+    /------/nucpath
+    /my_mat_metadata
+    /nuclidelist
+
+`my_mat` (the `datapath` -- default `material`) is a HDF5 dataset containing the
+material composition, `nucpath` is a attribute containing the path to the
+nuclide list (nested in the `datapath`).
+`my_mat_metadata` is a dataset containing the metatdata of the material.
+`nuclidelist` is a dataset containing the list of nuclides composition the
+material.
+
 
 `from_hdf5()` will detect the structure (old or new) of the file (when using
 `protocol1`).
-
 
 **************
 Material Class
@@ -56,7 +69,8 @@ Material Class
 *****************************
 Material Generation Functions
 *****************************
-The following top-level module functions are used to generate materials from various sources.
+The following top-level module functions are used to generate materials from
+various sources.
 
 .. autofunction:: from_atom_frac
 
--- a/docs/pyapi/material.rst
+++ b/docs/pyapi/material.rst
@@ -12,19 +12,20 @@ All functionality may be found in the ``material`` package::
 
  from pyne import material
 
-Materials are the primary container for radionuclides. They map nuclides to
-**mass weights**, though they contain methods for converting to/from atom
+Materials are the primary container for mixtures of radionuclides. They map nuclides to
+**mass fractions**, though they contain methods for converting to/from atom
 fractions as well.  In many ways they take inspiration from numpy arrays and
 python dictionaries.  Materials have two main attributes which define them.
 
 1. **comp**: a normalized composition mapping from nuclides (zzaaam-ints) to
-   mass-weights (floats).  2. **mass**: the mass of the material.
+   mass-fractions (floats).  
+   2. **mass**: the mass of the material.
 
 The :ref:`pyne_material_library` class is available to efficiently manipulate
 collections of materials.  The material class is presented below.  For more
 information please refer to :ref:`usersguide_material`.
 
-When using the `write_hdf5` method, the default structure for the `HDF5` file
+When using the `write_hdf5` method to write a material in a group named `my_mat`, the default structure for the `HDF5` file
 is:
 .. verbatim::
     /material/
@@ -73,4 +74,3 @@ Material Library
 .. autoclass:: MaterialLibrary(lib=None, datapath="/materials", nucpath="/nucid")
    :members:
    :inherited-members:
-
--- a/news/write_hdf5_structure.rst
+++ b/news/write_hdf5_structure.rst
@@ -6,7 +6,7 @@
     /material/ 
     /material/my_mat/ 
     /material/my_mat/composition
-    /material/my_mat/nucid 
+    /material/my_mat/nuclidelist 
     /material/my_mat/composition_metadata
 where, `/material` and `/material/my_mat` are now HDF% groups
 
--- a/news/write_hdf5_structure.rst
+++ b/news/write_hdf5_structure.rst
@@ -1,0 +1,19 @@
+**Added:** None
+
+**Changed:**
+- change the structure of material when written in a hdf5 file:
+.. verbatim::
+    /material/ 
+    /material/my_mat/ 
+    /material/my_mat/composition
+    /material/my_mat/nucid 
+    /material/my_mat/composition_metadata
+where, `/material` and `/material/my_mat` are now HDF% groups
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/pyne/cpp_material.pxd
+++ b/pyne/cpp_material.pxd
@@ -53,7 +53,7 @@ cdef extern from "material.h" namespace "pyne":
         void from_hdf5(char *, char *, int) except +
         void from_hdf5(char *, char *, int, int) except +
 
-        void write_hdf5(char *, char *, char *, float, int) except +
+        void deprecated_write_hdf5(char *, char *, char *, float, int) except +
         void write_hdf5(char *, char *, float, int) except +
 
         void from_text(char *) except +
--- a/pyne/cpp_material.pxd
+++ b/pyne/cpp_material.pxd
@@ -53,9 +53,8 @@ cdef extern from "material.h" namespace "pyne":
         void from_hdf5(char *, char *, int) except +
         void from_hdf5(char *, char *, int, int) except +
 
-        void write_hdf5(char *, char *, char *) except +
-        void write_hdf5(char *, char *, char *, float) except +
         void write_hdf5(char *, char *, char *, float, int) except +
+        void write_hdf5(char *, char *, float, int) except +
 
         void from_text(char *) except +
 
--- a/src/material.cpp
+++ b/src/material.cpp
@@ -81,26 +81,6 @@ void pyne::Material::_load_comp_protocol0(hid_t db, std::string datapath, int ro
   atoms_per_molecule = -1.0;
 }
 
-bool pyne::Material::detect_nuclidelist(hid_t data_set, std::string& nucpath){
-  hid_t nuc_attr = H5Aopen(data_set, "nucpath", H5P_DEFAULT);
-  
-  // can't find the "nucpath" Attribute
-  if(nuc_attr < 0) 
-    return false;
-  
-  H5A_info_t nuc_info;
-  H5Aget_info(nuc_attr, &nuc_info);
-  hsize_t nuc_attr_len = nuc_info.data_size;
-  hid_t str_attr = H5Tcopy(H5T_C_S1);
-  H5Tset_size(str_attr, nuc_attr_len);
-  char* nucpathbuf = new char[nuc_attr_len];
-  H5Aread(nuc_attr, str_attr, nucpathbuf);
-  nucpath = std::string(nucpathbuf, nuc_attr_len);
-  delete[] nucpathbuf;
-  H5Tclose(str_attr);
-  return true;
-}
-
 
 void pyne::Material::_load_comp_protocol1(hid_t db, std::string datapath,
                                           int row) {
@@ -110,25 +90,21 @@ void pyne::Material::_load_comp_protocol1(hid_t db, std::string datapath,
   // Grab the nucpath
   if (detect_nuclidelist(data_set, nucpath)){
     H5Dclose(data_set);
-  _load_comp_protocol1(db, datapath, nucpath, row);
+    _load_comp_protocol1(db, datapath, nucpath, row);
   } else {
     H5Dclose(data_set);
-    std::cout << "Can't find location of the nuclide list: nucpath attribute not found!" << std::endl;
-    exit(1);
+    throw std::runtime_error("Can't find location of the nuclide list: nucpath attribute not found!");
   }
   
 }
 
 void pyne::Material::_load_comp_protocol1(hid_t db, std::string datapath,
                                           std::string nucpath, int row) {
-  bool datapath_exists = h5wrap::path_exists(db, datapath);
-  bool nucpath_exists = h5wrap::path_exists(db, nucpath);
+  if (!h5wrap::path_exists(db, nucpath))
+    throw std::runtime_error("No path found at the location: " + nucpath);
 
-  if (!nucpath_exists) {
-    std::cout << "nucpath not found "<< std::endl;
-    //nucpath = datapath + "_" + nucpath.substr(1);
-    nucpath_exists = h5wrap::path_exists(db, nucpath);
-  }
+  if (!h5wrap::path_exists(db, datapath))
+    throw std::runtime_error("No path found at the location: " + datapath);
 
   hid_t data_set = H5Dopen2(db, datapath.c_str(), H5P_DEFAULT);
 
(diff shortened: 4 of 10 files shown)


bam241:
@gonuke is it something like that you had in mind ?

gonuke:
Where will material libraries live in this structure?

bam241:
material library will live in:
`/material_library`
so a multiple single materials and multiple material_library could co-exist in the same file


so something like:
```
/material_library/
/material_library/mylib_1/
/material_library/mylib_1/nucid
/material_library/mylib_1/composition
/material_library/mylib_1/composition_metadata
/material_library/mylib_2/
/material_library/mylib_2/nucid
/material_library/mylib_2/composition
/material_library/mylib_2/composition_metadata
```

with `datapath` being a `string` variable

bam241:
@gonuke @pshriwise @makeclean, I think this is ready for review!

I issue this there and not in the material_library PR (#1104 ) to simplify the review process.

Similar structure refactoring will be done for the `MaterialLibrary`


a simple test has been added to ensure, backward compatibility and ability to still use the old structure

bam241:
@ngranda @ypark234 @a-co @ NybergWISC : feel free to review this if you have 5 minutes to spear :)

This is mainly some C++ development, with some hint of `python` binding using `Cython`

a-co (approved):

a-co (commented):
The loop looks good to me. Nice solution to avoid the default false.

a-co (approved):

src/material.cpp
```diff
@@ -469,13 +480,125 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+        if (datapath_exists) {
+          H5Fclose(db);
+          write_hdf5(filename, datapath, "/nucid", row, chunksize);
+          return;
+        }
+      } else if ( object_info.type >= H5O_TYPE_NTYPES || object_info.type < H5O_TYPE_UNKNOWN ) {
+        // create Group
+        H5Gcreate2(db, "/material", H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
+      } else if (object_info.type != H5O_TYPE_GROUP) {
+        std::cout << "Non-group/non-dataset object /Material already exists "
```
gonuke:
```suggestion
        std::cout << "Non-group/non-dataset object /material already exists "
```

src/material.cpp
```diff
@@ -469,13 +480,125 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+    status = H5Oget_info_by_name(db, "/material", &object_info, H5P_DEFAULT);
+
+    if (object_info.type == H5O_TYPE_DATASET)
+       {
+        // Check if path exists
+        bool datapath_exists = h5wrap::path_exists(db, datapath);
+        // no Group but path fallback on old method using default
+        if (datapath_exists) {
+          H5Fclose(db);
+          write_hdf5(filename, datapath, "/nucid", row, chunksize);
```
gonuke: is it always "/nucid" in this case?

src/material.cpp
```diff
@@ -469,13 +480,125 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+    // Check is /material type
+    herr_t status;
+    H5O_info_t object_info;
+    status = H5Eset_auto2(H5E_DEFAULT, NULL, NULL);
+    status = H5Oget_info_by_name(db, "/material", &object_info, H5P_DEFAULT);
+
+    if (object_info.type == H5O_TYPE_DATASET)
+       {
+        // Check if path exists
+        bool datapath_exists = h5wrap::path_exists(db, datapath);
```
gonuke: why not just make the call in the `if` statement below?

src/material.cpp
```diff
@@ -469,13 +480,125 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+        // no Group but path fallback on old method using default
+        if (datapath_exists) {
```
gonuke: what do we do if this is a dataset, but the datapath doesn't exist?

src/material.cpp
```diff
@@ -469,13 +480,125 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+          H5Fclose(db);
+          write_hdf5(filename, datapath, "/nucid", row, chunksize);
+          return;
+        }
+      } else if ( object_info.type >= H5O_TYPE_NTYPES || object_info.type < H5O_TYPE_UNKNOWN ) {
```
gonuke: are these the conditions that define that a group does not already exist?  Is there no simpler way?

gonuke: This may need more comments for which case is which in the if statements

src/material.cpp
gonuke: Could there be a pre-existing nucpath that has a different name?  Maybe it doesn't matter because now it will definitely be this name?

gonuke (changes requested):
This all looks a little messy to me.  I think the logic can be cleaned up and/or documented better to make it easier to follow.  I understand that some of this comes from wanting to preserve the backwards compatibility, but maybe we should do that in a more modular way to better communicate to future developers.

src/material.cpp
bam241: this is true, We should attempt to discover it....

pyne/cpp_material.pxd
```diff
@@ -53,9 +53,8 @@ cdef extern from "material.h" namespace "pyne":
         void from_hdf5(char *, char *, int) except +
         void from_hdf5(char *, char *, int, int) except +
 
-        void write_hdf5(char *, char *, char *) except +
-        void write_hdf5(char *, char *, char *, float) except +
-        void write_hdf5(char *, char *, char *, float, int) except +
+        void deprecated_write_hdf5(char *, char *, char *, float, int) except +
+        void write_hdf5(char *, char *, float, int) except +
```
bam241, line 57: Here I decided to expose only the C++ method we are actually using on the python binding (for `write_hdf5()`)

src/material.cpp
```diff
@@ -469,13 +497,149 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+  if (!h5wrap::path_exists(db, "/material" + datapath)) {
+    data_id = H5Gcreate2(db, ("/material"+datapath).c_str(), H5P_DEFAULT,
+                         H5P_DEFAULT, H5P_DEFAULT);
+  } else {
+    data_id = H5Gopen2(material_grp_id, datapath.c_str(), H5P_DEFAULT);
+  }
+  //std::string nucpath = "/nuclidelist";
+  std::string full_datapath = "/material" + datapath + "/composition";
+  std::string nucpath = "/material" + datapath + "/nuclidelist";
+  std::vector<int> nuclides = write_hdf5_nucpath(data_id, nucpath);
```
bam241: wanted to use relative path (this is why I stored the different group ids.
But for a reason I ignore it does not work...
So I switched back to absolute path....

bam241:
@gonuke I implemented to change we discussed yesterday...

bam241:
@gonuke I cleaned that up based on you comment on the other PR.

Do you have anymore thoughts  ?

src/material.cpp
```diff
@@ -1889,3 +2050,25 @@ pyne::Material pyne::Material::operator/ (double y) {
   // Overloads x / y
   return pyne::Material(comp, mass / y, density );
 }
+
+bool pyne::detect_nuclidelist(hid_t data_set, std::string& nucpath){
```
bam241, line 2102: Got this one out of the `Material` class but kept it in the file as it is related to `Material` but I wanted to be able to access it outside of the `Material` class

src/material.h
```diff
@@ -151,19 +165,53 @@ namespace pyne
     ///            float.  A value of -0.0 indicates that the material should be
     ///            appended to the end of the dataset.
     /// \param chunksize The chunksize for all material data on disk.
-    void write_hdf5(char * filename, char * datapath, char * nucpath, float row=-0.0,
+    void deprecated_write_hdf5(char * filename, char * datapath, char * nucpath, float row=-0.0,
```
gonuke: Might need to change the comments to clarify why this is deprecated?

docs/pyapi/material.rst
```diff
@@ -12,15 +12,38 @@ All functionality may be found in the ``material`` package::
 
  from pyne import material
 
-Materials are the primary container for radionuclides. They map nuclides to **mass weights**, 
-though they contain methods for converting to/from atom fractions as well.
-In many ways they take inspiration from numpy arrays and python dictionaries.  Materials
-have two main attributes which define them.
+Materials are the primary container for radionuclides. They map nuclides to
```
gonuke:
```suggestion
Materials are the primary container for mixtures of radionuclides. They map nuclides to
```
(I know you didn't change this, but its an opportunity to clarify)

docs/pyapi/material.rst
```diff
@@ -12,15 +12,38 @@ All functionality may be found in the ``material`` package::
...
+**mass weights**, though they contain methods for converting to/from atom
```
gonuke:
```suggestion
**mass fractions**, though they contain methods for converting to/from atom
```

docs/pyapi/material.rst
```diff
@@ -12,15 +12,38 @@ All functionality may be found in the ``material`` package::
...
+fractions as well.  In many ways they take inspiration from numpy arrays and
+python dictionaries.  Materials have two main attributes which define them.
 
-1. **comp**: a normalized composition mapping from nuclides (zzaaam-ints) to mass-weights (floats).
-2. **mass**: the mass of the material.
+1. **comp**: a normalized composition mapping from nuclides (zzaaam-ints) to
+   mass-weights (floats).  2. **mass**: the mass of the material.
```
gonuke:
```suggestion
   mass-fractions (floats).  
   2. **mass**: the mass of the material.
```
I'm not sure if this formatting will render correctly

docs/pyapi/material.rst
```diff
@@ -12,15 +12,38 @@ All functionality may be found in the ``material`` package::
...
+
+The :ref:`pyne_material_library` class is available to efficiently manipulate
+collections of materials.  The material class is presented below.  For more
+information please refer to :ref:`usersguide_material`.
+
+When using the `write_hdf5` method, the default structure for the `HDF5` file
```
gonuke:
```suggestion
When using the `write_hdf5` method to write a material in a group named `my_mat`, the default structure for the `HDF5` file
```

docs/pyapi/material.rst
```diff
@@ -12,15 +12,38 @@ All functionality may be found in the ``material`` package::
...
+is:
+.. verbatim::
+    /material/
+    /material/my_mat/
+    /material/my_mat/composition
+    /material/my_mat/nucid
+    /material/my_mat/composition_metadata
+
```
gonuke: How about something like this?
```
/material/
|--- my_mat/
     |--- composition
     |--- nuclidelist
     |--- composition_metadata
```

docs/pyapi/material.rst
```diff
@@ -12,15 +12,38 @@ All functionality may be found in the ``material`` package::
...
+
+Where, `/material` and `/material/my_mat` are HDF5 groups, (`my_mat` value shall
+be provided as the datapath by the user). IF the `$datapath` or `/material`
+exist as a dataset in the file, then the old writing method will ne used.
+
+Older structure are still available to written when providing a `nucpath` to the
```
gonuke: Is this older structure defined somewhere?

src/material.cpp
```diff
@@ -99,59 +116,53 @@ void pyne::Material::_load_comp_protocol1(hid_t db, std::string datapath, int ro
...
   H5Dread(data_set, desc, mem_space, data_hyperslab, H5P_DEFAULT, mat_data);
 
   mass = (*mat_data).mass;
   density = (*mat_data).density;
   atoms_per_molecule = (*mat_data).atoms_per_mol;
-  for (int i = 0; i < nuc_size; i++)
-    comp[nuclides[i]] = (double) (*mat_data).comp[i];
-
+  for (int i = 0; i < nuc_size; i++) {
+    if ((double)(*mat_data).comp[i] != 0) {
```
gonuke, line 161: I thought this changed the other way in the other PR?  If this condition is not met, what do we get for `comp[nuclides[i]]`?

src/material.cpp
```diff
@@ -163,43 +174,39 @@ void pyne::Material::_load_comp_protocol1(hid_t db, std::string datapath, int ro
...
+               metadata, false);
 
   // close attr data objects
   H5Fflush(db, H5F_SCOPE_GLOBAL);
   H5Dclose(metadataet);
   H5Sclose(metadatapace);
   H5Tclose(attrtype);
-
-  // Close out the HDF5 file
-  H5Fclose(db);
```
gonuke, line 188: why don't we want to close the file/db here?  We open it in this method, right?

src/material.cpp
```diff
@@ -286,29 +285,53 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
 
+  // if nucpath exist: get it and check against the one we have!
   if (nucpath_exists) {
-    nuclides = h5wrap::h5_array_to_cpp_vector_1d<int>(db, nucpath, H5T_NATIVE_INT);
+    nuclides =
+        h5wrap::h5_array_to_cpp_vector_1d<int>(db, nucpath, H5T_NATIVE_INT);
     nuc_size = nuclides.size();
     nuc_dims[0] = nuc_size;
+    bool missing_nucs = false;
+    for ( pyne::comp_iter i = comp.begin(); i != comp.end(); i++ ) {
```
gonuke, line 329: Can we assume that either `comp` or `nuclides` is sorted?  Might be able to do a comparison that is more efficient and/or simpler code

src/material.cpp
```diff
@@ -469,13 +485,163 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+    // Create the file
+    db = H5Fcreate(filename.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, fapl);
+
+  } else {
+    bool ish5 = H5Fis_hdf5(filename.c_str());
+    if (!ish5) throw h5wrap::FileNotHDF5(filename);
+    db = H5Fopen(filename.c_str(), H5F_ACC_RDWR, fapl);
+
+    // if datapath != /material and exist -> old format
+    if (h5wrap::path_exists(db, datapath) && datapath != "/material") {
```
gonuke: I think this might need an additional comment.

If the user gives `datapath = "my_mat"` this `path_exists()` will definitely fail, right?  But if the user gives `datapath = "/my_mat"` then it may succeed as an instance of the old format.  Furthermore, the latter example is never valid in our new structure.  I think this leading slash is a very subtle difference that is worth explaining.

src/material.cpp
```diff
@@ -469,13 +485,163 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+  H5O_info_t object_info;
+  status = H5Eset_auto2(H5E_DEFAULT, NULL, NULL);
+  status = H5Oget_info_by_name(db, "/material", &object_info, H5P_DEFAULT);
+  // check if "/material" exists
+  if (status == 0) {
+    if (object_info.type ==
+        H5O_TYPE_DATASET) {  // "/material" is a dataset -> old format use
+                             // deprecated write_hdf5
+      H5Fclose(db);
+      deprecated_write_hdf5(filename, datapath, "/nucid", row, chunksize);
```
gonuke: Is it worth it to overload `deprecated_write_hdf5` to have a version that takes a `db` and one that takes a `filename`?

It looks awkward to close this `db` and then open it again in the method

news/write_hdf5_structure.rst
```diff
@@ -0,0 +1,19 @@
+**Added:** None
+
+**Changed:**
+- change the structure of material when written in a hdf5 file:
+.. verbatim::
+    /material/ 
+    /material/my_mat/ 
+    /material/my_mat/composition
+    /material/my_mat/nucid
```
gonuke:
```suggestion
    /material/my_mat/nuclidelist 
```

gonuke (changes requested):
Most of this is documentation, either in comments or in code

src/material.cpp
bam241: this is the composition, I can't find it in the original `material.cpp` file.
I think I introduced this in the MaterialLibrary PR. 
Then removed it into this one, so it disappear from the MaterialLibrary PR and you were highlighted from the merge/rebase commit.

I kinda like it, it prevent to have empty nuclide in a composition. 
`comp` is a `map<nuc, double>` (`nuc` is a `int`) containing the material composition.

src/material.cpp
bam241: we don't here. 
This method takea a `hid_t db` as a parameter so we don't want to close here but in the method that calls it.

src/material.cpp
bam241: I don't know enough hdf5 reading/writing mechanism to guaranty it.

they are sorted when we issue the write. so if they remain sorted in the hdf5 dataset and when we read them back, then they are sorted

src/material.cpp
bam241: I forgot to add the `/` into the `datapath` is not already present...


 it should resolve this potential issue

src/material.cpp
bam241: as I rename it, I can probably change it to directly take a `db` instead of a filename

src/material.cpp
gonuke: I see...if the value is zero, we never populate that entry of the dictionary 👍

src/material.cpp
gonuke: Yes... I misread this above.

src/material.cpp
gonuke: I think it’s probably correct, but it took me a long time of thinking carefully about the difference between the cases I mentioned to convince myself of the behavior

src/material.cpp
gonuke: You might need/want a version that takes a filename as well - at least looking at the use cases above

src/material.h
bam241: shall we move that into `private` to avoid external call ?

src/material.cpp
```diff
@@ -469,13 +485,168 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+  if (!pyne::file_exists(filename)) {
+    // Create the file
+    db = H5Fcreate(filename.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, fapl);
+
+  } else {
+    bool ish5 = H5Fis_hdf5(filename.c_str());
+    if (!ish5) throw h5wrap::FileNotHDF5(filename);
+    db = H5Fopen(filename.c_str(), H5F_ACC_RDWR, fapl);
+
+    // if datapath != /material and exist -> old format
```
gonuke:
```suggestion
    // there are a few things that may be true here:
    // datapath does not exist - there are many possible reasons for this, including
    //         datapath = "non_default_path" without a leading slash (/) --> assume this is a modifier for a new format
    // datapath does exist:
    //         datapath = "/non_default_path" --> old format
    //         datapath = "/material"   --> new format
```

src/material.h
gonuke: Sure - if it's not part of the public interface anymore

gonuke:
Do you know why this is failing test? I think I'm pretty much ready to merge this

bam241:
> Do you know why this is failing test? I think I'm pretty much ready to merge this

it seems to be the issue #1210  with `LANL` website

bam241:
I also added a quick check to verify that the `datapath` starts a `/`. and had it if not.

gonuke:
do we have enough tests of the new behavior?  and how we now treat the old behavior?  etc?

bam241:
> do we have enough tests of the new behavior? and how we now treat the old behavior? etc?

probably not...
I added a test to specifically test old format writing, but don't have test to check all the possible cases when attempting to write in a file that is already in the old format.
I'll add this and ping you.

bam241:
@gonuke this should be ok now.

I added tests to verify write and read using old format on the same `datapath` (default and non default path) and on 2 different different paths.
 -- and fixed some part of the from_hdf5()... 

The only thing that we don't (and cannot) check is detecting old format on a non-default `datapath` (if the second write is on a different one) as we don't know where to look for...

src/material.cpp
```diff
@@ -221,19 +227,41 @@ void pyne::Material::from_hdf5(std::string filename, std::string datapath, int r
...
+    // Check /material type
+    herr_t status;
+    H5O_info_t object_info;
+    status = H5Eset_auto2(H5E_DEFAULT, NULL, NULL);
+    status = H5Oget_info_by_name(db, "/material" , &object_info, H5P_DEFAULT);
+    herr_t status2;
+    H5O_info_t object_info2;
+    status2 = H5Eset_auto2(H5E_DEFAULT, NULL, NULL);
+    status2 = H5Oget_info_by_name(db, datapath.c_str(), &object_info, H5P_DEFAULT);
+    // Group "/material" does not exist
```
bam241: Had to add a test for H5_DATASET for in the custom `datapath` to the one in the default `datapath` (`/material`)

src/material.cpp
```diff
@@ -243,8 +243,13 @@ void pyne::Material::from_hdf5(std::string filename, std::string datapath, int r
     H5O_info_t object_info;
     status = H5Eset_auto2(H5E_DEFAULT, NULL, NULL);
     status = H5Oget_info_by_name(db, "/material" , &object_info, H5P_DEFAULT);
+    herr_t status2;
```
gonuke: Can we add something like this to help document what we are doing?  i.e. I don't actually know that values of `status` or the logic, but this will make it easier.
```suggestion
    bool material_path_exists = (status ==0 && object_info.type != H5O_TYPE_DATASET)
    herr_t status2;
```
We can then use it in the `if` statement in line 251

gonuke: Similar for status2....

src/material.cpp
```diff
@@ -221,19 +227,45 @@ void pyne::Material::from_hdf5(std::string filename, std::string datapath, int r
...
+    status = H5Oget_info_by_name(db, "/material" , &object_info, H5P_DEFAULT);
+    bool material_notdataset_exists = (status == 0 && object_info.type != H5O_TYPE_DATASET);
+   
+    // Reset status and test for datapath
+    status = H5Eset_auto2(H5E_DEFAULT, NULL, NULL);
+    status = H5Oget_info_by_name(db, datapath.c_str(), &object_info, H5P_DEFAULT);
+    bool datapath_notdataset_exists = (status == 0 && object_info.type != H5O_TYPE_DATASET);
+    
+    // Group "/material" does not exist as a non-dataset
+    if (!material_notdataset_exists || !datapath_notdataset_exists) {
```
bam241: @gonuke does it make more sense ?
I used your suggestion, renamed the `bool` variable so it contains all the info and added some comments

gonuke: I think the logic is clearer, but there is some lingering confusion (maybe it's just me) because we are mixing a few cases here.

It is old format if either "/material" or `datapath` is a dataset  - this is clear
It is new format if "/material" is a groupset   - this is clear
It is irrelevant if datapath is a non-dataset   - this is implied but not clear

So maybe some new variable names/logic (pseudocode):

```
enum data_layout_types = {unknown, old_format, new_format};

data_layout_types this_data_layout = unknown

// get info for `datapath`
if (status == 0 && object_info.type == DATASET) this_data_layout = old_format

// get info for "/material"
if (status == 0 && object_info.type == DATASET) { 
    this_data_layout = old_format
} else if (status == 0 && object_info.type == GROUP) {
    this_data_layout = new_format
}

```

In particular, the current code will try the old format if both "/material" and `datapath` do NOT exist.  It will probably fail later, but why not fail here if that is true.

gonuke: instead of the `enum` you could have two `bool`: `format_detected` and `new_format`

src/material.cpp
```diff
@@ -469,13 +499,176 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+    } else {  // "/material" is a group get his hid
+      material_grp_id = H5Gopen2(db, "/material", H5P_DEFAULT);
+    }
+  } else {  // "/material" do not exist -> create it !
+    material_grp_id =
+        H5Gcreate2(db, "/material", H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);
+  }
+  // Check if /material exist as a Group
+  // Group "/material/datapath" does not exist create it
+  if (!h5wrap::path_exists(db, "/material" + datapath)) {
```
gonuke, line 616: Isn't your default `datapath` still "/material"?  So doesn't this mean we get "/material/material" as the default group?

bam241: do you think we should use an other default ? like "mat" ?

bam241:
@gonuke I updated the layout detection algorithm.

what do you think ?

src/material.cpp
```diff
@@ -221,19 +260,39 @@ void pyne::Material::from_hdf5(std::string filename, std::string datapath, int r
...
+  if (protocol == 0) {
+    bool datapath_exists = h5wrap::path_exists(db, datapath);
+    if (!datapath_exists)
+      throw h5wrap::PathNotFound(filename, datapath);
     _load_comp_protocol0(db, datapath, row);
-  else if (protocol == 1)
-    _load_comp_protocol1(db, datapath, row);
-  else
+  } else if (protocol == 1) {
+    enum prot1_layout {path_donotexists=-1, unknown, old_layout, new_layout};
```
gonuke: Define this as a private member of the `Material` class and then use them as the return types of the function?

src/material.cpp
```diff
@@ -469,13 +521,175 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+    // Create the file
+    db = H5Fcreate(filename.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, fapl);
+
+  } else {
+    bool ish5 = H5Fis_hdf5(filename.c_str());
+    if (!ish5) throw h5wrap::FileNotHDF5(filename);
+    db = H5Fopen(filename.c_str(), H5F_ACC_RDWR, fapl);
+
+    // there are a few things that may be true here:
+    // datapath does not exist - there are many possible reasons for this, including
```
gonuke: Can we use the new method to determine the layout here as well?

src/material.cpp
gonuke: or "mat_name" or something to make sure it's clear, at least to the software developers...

gonuke (changes requested):
This is a lot better... wonder if we can use it more places - see my comments?

bam241:
I change in most places `datapath` for `mat_name` except in the deprecated_write_hdf5

I also make use of the `detect_hdf5_layout` method in `write_hdf5`

bam241:
might have discover issues, please restrain from merging yet....

src/material.h
```diff
@@ -123,48 +130,123 @@ namespace pyne
...
     /// Writes this material out to an HDF5 file.
     /// This happens according to protocol 1.
     /// \param filename Path on disk to the HDF5 file.
+    /// \param mat_name Path to the the material in the file.
+    /// \param row The index to read out, may be negative. Also note that this is a
+    ///            float.  A value of -0.0 indicates that the material should be
+    ///            appended to the end of the dataset.
+    /// \param chunksize The chunksize for all material data on disk.
+    /// New write_hdf5 which fallback on the old one when required
+    void write_hdf5(std::string filename, std::string mat_name="/material",
```
gonuke: I think it was this string "/material" that I was recommending changing to "/mat_name", not the variable name `datapath`.

With this default, you get a group "/material/material" by default which can be a little confusing.

bam241: I changed it in most of the places... missed that one... :)

bam241: ok I did not change the default values.... but all the variables name.....

src/material.cpp
```diff
@@ -469,13 +519,173 @@ void pyne::Material::write_hdf5(std::string filename, std::string datapath,
...
+    // /material as a group -> new layout: open the "/material" group
+    case prot1_layout::new_layout: {
+      material_grp_id = H5Gopen2(db, "/material", H5P_DEFAULT);
+      break;
+    }
+  }
+  
+
+  //mat_name is proided with a "/" so need to open with fullpath
+  // Group "/material/mat_name" does not exist create it
```
gonuke: You still have "/material" as the default value of the variable `mat_name`, so this comment should refer to "/material/material" (or change the default value?)

bam241:
> might have discover issues, please restrain from merging yet....

I fixed it... but I now am scared to merge without have the test working on CI....

src/material.h
gonuke: I actually think the variable names make more sense as `datapath`....  And we still call it `datapath` in the python layer

gonuke:
I'm ready to go forward with this one we get passing tests...

bam241:
I updated `mesh.py` to use the new material layout.

 the only tests failing on my machine are the `endf` ones.

gonuke:
I think some of these failures are real....

bam241:
yes there are,

I started looking at them Friday, did had the time to solve them....

bam241:
@gonuke I fixed the tests.

I add to modify a little bit how `mesh.py` was looking for and loading materials.
see change in [[`mesh.py`](https://github.com/pyne/pyne/pull/1207/files#diff-4be2851467674dbfe9961bc16279fd1b)]

gonuke:
Thanks @bam241 !! - this was a long and interesting development and I think we're ready to merge it into the mainline.

gonuke (approved):

PR #1211, opened 2020-03-04
Add a flag for MCNP/PHITS material cards to write mass fractions instead of mass densities

ypark234:
This PR adds a boolean argument `mult_den` to methods `get_density_frac` and therefore `mcnp`, `phits`, so that users can choose to write MCNP/PHITS material cards with either (1) mass density (default) or (2) mass fraction.

### DIFF
--- a/pyne/material.pyx
+++ b/pyne/material.pyx
@@ -335,18 +335,18 @@ cdef class _Material:
         c_nucpath = nucpath_bytes
         self.mat_pointer.write_hdf5(c_filename, c_datapath, c_nucpath, row, chunksize)
 
-    def phits(self, frac_type='mass'):
+    def phits(self, frac_type='mass', mult_den=True):
         """phits(frac_type)
         Return an phits card
         Parameters
         ----------
- 	   int 0 means use "mass" as the frac_type
+        int 0 means use "mass" as the frac_type
         """
         cdef std_string card
-        card = self.mat_pointer.phits(frac_type.encode())
+        card = self.mat_pointer.phits(frac_type.encode(), mult_den)
         return card.decode()
 
-    def mcnp(self, frac_type='mass'):
+    def mcnp(self, frac_type='mass', mult_den=True):
         """mcnp(frac_type)
         Return an mcnp card
         Parameters
@@ -354,9 +354,9 @@ cdef class _Material:
         int 0 means use "mass" as the frac_type
         """
         cdef std_string card
-        card = self.mat_pointer.mcnp(frac_type.encode())
+        card = self.mat_pointer.mcnp(frac_type.encode(), mult_den)
         return card.decode()
-    
+
     def get_uwuw_name(self):
         """get_uwuw_name()
         Return a uwuw material name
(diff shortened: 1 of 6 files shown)


ypark234:
Thanks @bam241 for helping me with making this PR.
I am seeing some tests failing less likely related to these changes?

bam241:
news file is missing, 

the other tests failing are not related to this PR...

pyne/material.pyx
```diff
@@ -335,28 +335,28 @@ cdef class _Material:
...
         """
         cdef std_string card
-        card = self.mat_pointer.phits(frac_type.encode())
+        card = self.mat_pointer.phits(frac_type.encode(), mult_den)
         return card.decode()
 
-    def mcnp(self, frac_type='mass'):
+    def mcnp(self, frac_type='mass', mult_den=True):
         """mcnp(frac_type)
         Return an mcnp card
```
bam241, line 356: Could you update the docstring ?

pyne/material.pyx
```diff
@@ -335,28 +335,28 @@ cdef class _Material:
         c_nucpath = nucpath_bytes
         self.mat_pointer.write_hdf5(c_filename, c_datapath, c_nucpath, row, chunksize)
 
-    def phits(self, frac_type='mass'):
+    def phits(self, frac_type='mass', mult_den=True):
         """phits(frac_type)
         Return an phits card
```
bam241, line 340: here as well

bam241 (commented):
Looking good.
Just few docstrings update on the python.

src/material.cpp
```diff
@@ -1483,7 +1483,9 @@ std::map<int, double> pyne::Material::get_density_frac(std::string frac_type){
     fracs = comp;
     if (density != -1.0) {
       for (comp_iter ci = fracs.begin(); ci != fracs.end(); ci++){
-        ci->second *= density;
+        if (mult_den) {
```
gonuke: Move this conditional up 2 lines.... no point in entering the loop or checking the value of `density` if this is not true

gonuke (changes requested):
Thanks @ypark234 - one small suggestion in the coding.

Also - why not make this same flag useful for atom densities/fractions?  Right now the only logic that determines that is whether or not there is a value in `density`.  You are now adding more logic to that case for mass densities, but could easily do the same for atom densities.

I think in both cases we change
```
if (density != -1.0):
```
to
```
if (density != -1.0 && !mult_den):
```

gonuke:
I'm also wondering if it makes sense to write a small (inline?) `to_mass_dens` function to be consistent with the `to_atom_dens` and `to_atom_frac` functions.

src/material.cpp
ypark234: Don't know why I put the conditional inside for loop...
The above comment makes sense except that I assume you meant 
`&& mult_den):` instead of `&& !mult_den):`

gonuke:
I've merged @bam241's PR to skip the tests that require the download from LANL.  Rebase to see if you now pass all tests, and then I'm happy to merge.

ypark234:
Rebased and all tests passed. Thanks for feedbacks, @gonuke and @bam241.

gonuke (approved):

gonuke:
Thanks @ypark234

PR #1212, opened 2020-03-05
Attempt to add try/catch/skip to endf

bam241:
Add try/catch/skip to endf tests that require some file from remote location.

Also force the endf test to run without skip to inform if those test have been skipped.


gonuke:
Thanks for implementing this @bam241 

I'll merge this now, but we should consider a longer-term solution, e.g. store our own copy of these files but in a limited-access location and only for CI

Issue #1209, opened 2020-02-10
PyNE ENDF reader fails to load SiO2 and benzene files from ENDF/VIII.0


tkittel:
Hi there,

First of all, thanks to everyone creating/maintaining such a nice project!

While using the pyne.endf module to extract some thermal scattering laws, I noticed that 3 files from the recent ENDF release could not be read. They are the two SiO2 files and the benzene files in the Thermal Neutron Scattering Sublibrary at https://www.nndc.bnl.gov/endf/b8.0/download.html .

After some investigation, I think I understood the problem. Reading carefully section 7.4.1 of the endf6 manual https://www-nds.iaea.org/public/endf/endf-manual.pdf I noticed that the optional teff tables for non-principal atoms at the end were never read in the `_read_thermal_inelastic` method of pyne.endf's `Evaluation` class. This, despite the fact that the manual mandates their presence when B[7], B[13] or B[19] are 0.0:

> <continue with TAB1 records for Teff1, Teff2 and/ or Teff3 if the
> corresponding value of a1 , a2 , or a3 in B(7), B(13), or B(19) is zero>

The unread records cause the pyne endf reading code to think it needs to invoke the `_read_thermal_inelastic` once again, but the second time it of course encounters unexpected values and crashes with a ValueError.

I am not an ENDF expert, but I have confirmed that I can read the file successfully If I insert the following lines at the very end of the `_read_thermal_inelastic` method:
```python
    for i in range(len(B)//6-1):
        if B[6*(i+1)]==0.0:
            params, teff = _self._get_tab1_record()
            inel['teff_%i'%(i+1)] = teff#Store as teff_1, teff_2, etc. for now.
```
For users like me who would like the reading to work for them already now, the method can be fixed via monkey-patching:
```python
import pyne.endf
_orig_rti = pyne.endf.Evaluation._read_thermal_inelastic
def _fix_rti(_self):
    _orig_rti(_self)
    inel=_self.thermal_inelastic
    B=inel['B']
    for i in range(len(B)//6-1):
        if B[6*(i+1)]==0.0:
            params, teff = _self._get_tab1_record()
            inel['teff_%i'%(i+1)] = teff#Store as teff_1, teff_2, etc. for now.
pyne.endf.Evaluation._read_thermal_inelastic=_fix_rti
```
Cheers,
Thomas

gonuke:
Thanks for this update @tkittel and the detective work to sort it out.

[...]

PR #1214, opened 2020-03-11
Fix ENDF parsing of TSL files using SCT for non-principal atoms

tkittel:
Opening pull request as request on issue #1209 .

### DIFF
--- a/news/endf_read_sct.rst
+++ b/news/endf_read_sct.rst
@@ -1,0 +1,12 @@
+**Added:** None
+
+**Changed:**
+  - Fix ENDF parsing of TSL files with short collision time approximation for non-principal atoms.
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


gonuke:
Thanks for this contribution @tkittel !  Are you able to add a test for this? Ideally a case that would have failed without this addition.

tkittel:
I don't really have a test case - apart from the SiO2 and benzene files from ENDF8. I don't think I am allowed to just commit those. If you have access to the relevant files on some hidden test server, it should be simple to simply try to parse those files.

gonuke:
Hmmm.... that is a problem.  Some of our tests download the files they will use from the internet, but these seem to only be available as part of larger zip/tarballs, which is not suitable for testing.

I guess the only other option would be to construct a minimally sufficient file that includes the features in question, but that sounds like too much work!

tkittel:
Yeah, I definitely don't have the skills to start learning how to construct ENDF files. I barely know how to read them, which is why I turned to PyNE in the first place :-)

gonuke:
Ha!  I don't really want to do that either.

PR #1218, opened 2020-04-06
Point Source

bam241:
This is an attempt to add Point Source in PyNE.

DocString are still missing, but most of the basic are there (C++/Python API, and python tests)

On the C++ point of view, the idea was to create a Source class that would be mainly empty.
This Source class would be inherited by other Source classes list the PointSource.

Ultimately the Source class could be allowed to include multiple sources (and have a dedicated method to generated input card for example:`mcnp()` that would allow to build a `sdf` card combining all of them.)

@bohmt If you have any thought please feel free to share.

### DIFF
--- a/pyne/cpp_source.pxd
+++ b/pyne/cpp_source.pxd
@@ -18,7 +18,17 @@ cdef extern from "source.h" namespace "pyne":
         PointSource(double, double, double, double, double, double, double, cstr, double) except +
 
         # attributes
+        double x
+        double y
+        double z
+        double i
+        double j
+        double k
+        double E
+        cstr particle
+        double weight
 
         # methods
         cstr mcnp() except +
+        cstr mcnp(int) except +
         pass
--- a/pyne/cpp_source.pxd
+++ b/pyne/cpp_source.pxd
@@ -1,0 +1,24 @@
+from libcpp.map cimport map as cpp_map
+from libcpp.string cimport string as cstr
+from libcpp.vector cimport vector
+
+cdef extern from "source.h" namespace "pyne":
+
+    cdef cppclass PointSource:
+        # constructors
+        PointSource() except +
+        PointSource(double) except +
+        PointSource(double, double) except +
+        PointSource(double, double, double) except +
+        PointSource(double, double, double, double) except +
+        PointSource(double, double, double, double, double) except +
+        PointSource(double, double, double, double, double, double) except +
+        PointSource(double, double, double, double, double, double, double) except +
+        PointSource(double, double, double, double, double, double, double, cstr) except +
+        PointSource(double, double, double, double, double, double, double, cstr, double) except +
+
+        # attributes
+
+        # methods
+        cstr mcnp() except +
+        pass
--- a/pyne/cpp_source.pxd
+++ b/pyne/cpp_source.pxd
@@ -21,9 +21,9 @@ cdef extern from "source.h" namespace "pyne":
         double x
         double y
         double z
-        double i
-        double j
-        double k
+        double u
+        double v
+        double w
         double E
         cstr particle
         double weight
--- a/src/source.cpp
+++ b/src/source.cpp
@@ -35,42 +35,45 @@ std::ostream& operator<<(std::ostream& os, pyne::Source tal) {
   return os;
 }
 
-pyne::SourcePoint::SourcePoint(double x, double y, double z, double u, double v,
+pyne::PointSource::PointSource(double x, double y, double z, double u, double v,
                                double w, double E, std::string p, double weight)
-    : x(x), y(y), z(z), u(u), v(v), w(w), E(E), p(p), weight(weight) {}
+    : x(x), y(y), z(z), i(i), j(j), w(w), E(E), p(p), weight(weight) {}
 
     
 // Destructor
-pyne::SourcePoint::~SourcePoint() {}
+pyne::PointSource::~PointSource() {}
 
-std::string pyne::SourcePoint::mcnp() {
+std::string pyne::PointSource::mcnp() {
   std::stringstream output;  // output stream
-  std::string separator = "     "
+  std::string indent = "     ";
 
-                          output
-                          << "SDEF "
+  output << "SDEF ";
 
-                          if (x != 0 && y != 0 && z != 0) {
+  if (x != 0 && y != 0 && z != 0) {
     output << "POS=" << x << " " << y << " " << z << std::endl;
+  }
+  
+  if (i != 0 && j != 0 && k != 0) {
+    output << indent << "VEC=" << i << " " << j << " " << k << "DIR=1"<< std::endl;
   }
 
   if (E != 14) {
-    output << std::endl << separator << "ERG=" << E;
+    output << std::endl << indent << "ERG=" << E;
   }
 
   if (weight != 1) {
-    output << std::endl << separator << "WGT=" << weight;
+    output << std::endl << indent << "WGT=" << weight;
   }
 
   if (p != "n") {
-    output << std::endl << separator << "PAR=" << pyne::particle::mcnp(p);
+    output << std::endl << indent << "PAR=" << pyne::particle::mcnp(p);
   }
 
   return output.str();
 }
 
 // Produces valid fluka source
-std::string pyne::SourcePoint::fluka() {
+std::string pyne::PointSource::fluka() {
   std::stringstream output;  // output stream
 
   return output.str();
--- a/src/source.cpp
+++ b/src/source.cpp
@@ -37,10 +37,10 @@ pyne::PointSource::~PointSource() {}
 
 std::string pyne::PointSource::mcnp(int version) const {
   std::stringstream output;  // output stream
-  
+
   std::stringstream newline;
   newline << std::endl << "     ";
-  
+
   output << "SDEF ";
   output << "POS=" << x << " " << y << " " << z;
   if (u != 0 || v != 0 || w != 0) {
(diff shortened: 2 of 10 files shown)


pyne/cpp_source.pxd
```diff
@@ -0,0 +1,34 @@
...
+        PointSource(double, double, double, double, double, double, double, cstr) except +
+        PointSource(double, double, double, double, double, double, double, cstr, double) except +
+
+        # attributes
+        double x
+        double y
+        double z
+        double i
+        double j
+        double k
```
kkiesling: I assume x, y, z are the position. What is i, j, k? If those are direction, we typically use u, v, and w.

tests/test_source.py
```diff
@@ -0,0 +1,56 @@
+"""Source tests"""
+from pyne.source import PointSource
+import os
+import warnings
+
+from unittest import TestCase
+import nose
```
kkiesling, line 7: nose is deprecated (see note [here](https://nose.readthedocs.io/en/latest/#note-to-users)). You can use just unittest or pytest.

tests/test_source.py
```diff
@@ -0,0 +1,56 @@
...
+    assert_equal(pt_src.particle, "p")
+    assert_equal(pt_src.weight, 0.5)
+
+
+################################################################################
+# test write particle for mcnp
+def test_point_source_mcnp():
+    pt_src = PointSource(1, 2, 3, 4, 5, 6, 12, "Proton", 0.5)
+    assert_equal(
+        "SDEF POS=1 2 3\n     VEC=4 5 6 DIR=1\n     ERG=12\n     WGT=0.5\n     PAR=h", pt_src.mcnp(6))
```
kkiesling: for easy of readability, you can store the expected string as a variable (eg `exp_str`)

pyne/cpp_source.pxd
```diff
@@ -0,0 +1,34 @@
...
+        PointSource() except +
+        PointSource(double) except +
+        PointSource(double, double) except +
+        PointSource(double, double, double) except +
+        PointSource(double, double, double, double) except +
+        PointSource(double, double, double, double, double) except +
+        PointSource(double, double, double, double, double, double) except +
+        PointSource(double, double, double, double, double, double, double) except +
+        PointSource(double, double, double, double, double, double, double, cstr) except +
+        PointSource(double, double, double, double, double, double, double, cstr, double) except +
```
kkiesling, line 18: I know you said docstrings are missing, but also be sure to include what the difference is with all these constructors.

tests/test_source.py
```diff
@@ -0,0 +1,56 @@
...
+    assert_equal(pt_src.i, 4)
+    assert_equal(pt_src.j, 5)
+    assert_equal(pt_src.k, 6)
+    assert_equal(pt_src.particle, "p")
+    assert_equal(pt_src.weight, 0.5)
+
+
+################################################################################
+# test write particle for mcnp
+def test_point_source_mcnp():
```
kkiesling: This test should be split into 5 different tests since it is testing 5 different cases.

tests/test_source.py
```diff
@@ -0,0 +1,56 @@
...
+    assert_almost_equal, assert_true, assert_false, assert_in
+
+from pyne.utils import QAWarning
+warnings.simplefilter("ignore", QAWarning)
+
+
+###############################################################################
+# tests the constructor
+def test_point_source():
+    pt_src = PointSource(1, 2, 3, 4, 5, 6, 12, "p", 0.5)
```
kkiesling, line 19: should this test also assert that E is 12?

kkiesling (commented):
I know it's still WIP, but I have a few questions and comments.

pyne/cpp_source.pxd
bam241: I know but there is weight, I wanted to avoid confusion between `w` and `weight`

kkiesling: But you're using `weight` as a variable instead anyway. i, j, k are typically used for an index notation.

src/source.cpp
```diff
@@ -0,0 +1,84 @@
...
+/************************/
+
+/*--- Constructors ---*/
+
+pyne::Source::Source() {}
+
+// Destructor
+pyne::Source::~Source() {}
+
+std::ostream& operator<<(std::ostream& os, pyne::Source tal) {
```
gonuke: This doesn't seem to do anything useful and is not overloaded in `PointSource`

src/source.cpp
```diff
@@ -0,0 +1,84 @@
...
+  // print the Source to ostream
+  os << "\t---------\n";
+  return os;
+}
+
+pyne::PointSource::PointSource(double x, double y, double z, double i, double j,
+                               double k, double E, std::string p, double weight)
+    : x(x), y(y), z(z), i(i), j(j), k(k), E(E), particle(p), weight(weight) {}
```
gonuke: I'm used to seeing different parameters from the variables, either `_x(x)` or `x(_x)` or something like that, to be clear about scope.  I don't have a strong reference for that, though - is there guidance in the style guide?

src/source.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+std::string pyne::PointSource::mcnp(int version) {
+  std::stringstream output;  // output stream
+  
+  std::stringstream newline;
+  newline << std::endl << "     ";
+  
+  
+  output << "SDEF ";
+  if (x != 0 || y != 0 || z != 0) {
```
gonuke: I think we might as well write the information explicitly, even if it is the defaults.  No reason not to. The defaults are a convenience to the user in terms of keystrokes.   Writing out the explicit defaults will be a convenience for readers.

src/source.h
```diff
@@ -0,0 +1,74 @@
...
+
+  // fluka tally
+  virtual std::string fluka() {}
+};
+
+class PointSource : public Source {
+ public:
+  /// Point Source Constructors
+  PointSource(double x = 0, double y = 0, double z = 0, double i = 0,
+              double j = 0, double k = 0, double E = 14, std::string p = "n",
```
gonuke: I think the MCNP default is actually 14.1 MeV isn't it???  Do we want to allow defaults?

tests/test_source.py
gonuke: Can we mix these if the rest of PyNE is using nose still?

gonuke (changes requested):
Mostly questions about default values for variables

tests/test_source.py
kkiesling: Good question, I didn't think of that. I do not know the answer to that.

kkiesling: Really we should switch PyNE away from nose.... buuuuuuut.. that sounds like too much work

gonuke: and certainly a different PR.  I don't think it will be that bad as much of the syntax for pytest is the same as nose

pyne/cpp_source.pxd
bam241: we don't usually put docstring in the cpp_XXX.pxd.

those are defined in the `.cpp/.hpp`

src/source.h
bam241: Screenshot 2020-04-13 10 13 08
User manual says 14MeV.


I don't know about all defaults, but some would be nice, such as the isotropy of the source...

tests/test_source.py
bam241: I wanted to keep them together as this should grow, with other kind of source, and other MC-softwares.

I fear if I split this to much, we will have billion of tests....
So I planned was to regroup the test per tested  method.
Don't have a. strong opinion there. so if you feel strongly about it, I'll fix it.

src/source.cpp
bam241: I keep the explicit information for all except for the direction.
I felt like it make more sense to not indicate the default (0 0 0) for an isotropic source.
let me know if you disagree

tests/test_source.py
bam241: changing from nose to purest does not look too complicated...
we have a lot of test so it will be a lot of work, but probably not that complicated.

src/source.cpp
```diff
@@ -0,0 +1,78 @@
...
+// Destructor
+pyne::PointSource::~PointSource() {}
+
+std::string pyne::PointSource::mcnp(int version) {
+  std::stringstream output;  // output stream
+  
+  std::stringstream newline;
+  newline << std::endl << "     ";
+  
+  std::cout << __LINE__ << std::endl;
```
gonuke: I don't understand what `__LINE__` is?

Do we really want to write to `std::cout` always?

gonuke (commented):
Just one question/comment....

src/source.cpp
bam241: sorry forgot tho clean that up :(

gonuke (approved):

gonuke:
LGTM

tests/test_source.py
kkiesling: I feel strongly that they should be split up into separate tests. It is not uncommon to test the same method/function in many different ways. Making separate tests will make it more clear what is failing.

If there is a concern about trying to keep track of all the tests when more source methods are added, then I think that can be resolved by separating them into different classes.

kkiesling (changes requested):
I think the tests should be separated still. Other than that, looks okay to me.

tests/test_source.py
bam241: will tackle this asap :)

gonuke: Maybe parametrize some of these? http://doc.pytest.org/en/latest/example/parametrize.html

gonuke:
In case there's any confusion, I'm waiting for action on splitting up the tests and then I think I can merge

bam241:
@gonuke @kkiesling this should be good to go :)

kkiesling (approved):
Thanks @bam241 for updating the tests. It looks good to me!

PR #1220, opened 2020-04-16
Resolved compatibility issue with HDF5 version 1.12.0

yunhuangzhang:
The latest HDF5 release (version 1.12.0) changed interfaces for several functions, including but not limited to: H5Oget_info_by_name(), H5Oget_info_by_idx(), and H5O_info_t(). The purpose of this fix is to maintain compatibility with legacy HDF5, by making use of HDF5 internal compatibility flags defined as compiler macros in the CMakelists.txt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,7 +80,15 @@ set(LIBS ${LIBS} ${HDF5_C_LIBRARIES})
 MESSAGE("--    HDF5 Include directory: ${HDF5_INCLUDE_DIRS}")
 MESSAGE("--    HDF5 Library directories: ${HDF5_LIBRARY_DIRS}")
 MESSAGE("--    HDF5 Libraries: ${HDF5_C_LIBRARIES}")
+MESSAGE("--    HDF5 library version: ${HDF5_VERSION}")
 
+# Version 1.12 of HDF5 deprecates the H5Oget_info_by_idx(), 
+# H5O_info_t() and H5Oget_info_by_name() interfaces.
+# Thus, we give these flags to allow usage of the old interface in newer
+# versions of HDF5.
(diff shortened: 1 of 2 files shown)


gonuke:
Thanks @yunhuangzhang for this contribution.  I think we'll need to enhance our automated testing to include HDF5 1.12.0, perhaps both demonstrating the prior failure and the success of this patch.

In the meantime, please add a news file to this PR.  You can find a template and some examples in the `news` directory.  These can be very brief, but are valuable in assembling release notes.

yunhuangzhang:
News file added following the template. Hope it helps! @gonuke

gonuke:
@bam241 - wondering if you have thoughts on how to test this....  here are my thoughts:

In a different PR
1. create a CI docker image with HDF 1.12
2. add a test to CI that uses that image
3. set that failing test only causes warn not error (since it will probably fail)

Merge that PR and then see if this PR passes on HDF 1.12, with an update here that changes that new test to be an error.

bam241:
yes I think that's a good set of action.
I can do it, just need to think where this new docker image/Dockerfile should live....

gonuke:
Can't it live in the pyne org for dockerhub?

bam241:
yes but we need to host the dockefile somewhere.
As we have (because of me) a single file to generate all the docker images we need in PyNE CI, its makes it a bit complicated if we want to be consistant...

gonuke:
Should we extend your dockerfile system to include the newer HDF5 version?

gonuke:
Hello @yunhuangzhang - if you rebase this branch against `pyne/develop`, it should add a test that will prove that is currently failing without this change.

gonuke:
This was merged in #1247

PR #1221, opened 2020-04-20
allowing Material almagamation without decay

bam241:
- Defines a PYNE_DECAY variable to detect if `decay.ccp` is part of amalgamation

- Regroups Material methods in `materials.cpp` & `material.h` to "disable" them is decay.h is not part of the amalgamated `PyNE`

- additional clean fix (clearing the comp when loading a new composition) -- I can do a separate PR for this, if you'd like

### DIFF
--- a/src/material.cpp
+++ b/src/material.cpp
@@ -2089,3 +2089,4 @@ bool pyne::detect_nuclidelist(hid_t data_set, std::string& nucpath){
   H5Tclose(nuc_attr);
   return true;
 }
+
--- a/src/material.cpp
+++ b/src/material.cpp
@@ -175,7 +175,8 @@ void pyne::Material::_load_comp_protocol1(hid_t db, std::string datapath,
   //
   std::string attrpath = datapath + "_metadata";
   bool attrpath_exists = h5wrap::path_exists(db, attrpath);
-  if (!attrpath_exists) return;
+  if (!attrpath_exists)
+    return;
 
   hid_t metadatapace, attrtype, metadataet, metadatalab, attrmemspace;
   int attrrank;
--- a/src/material.cpp
+++ b/src/material.cpp
@@ -45,9 +45,6 @@ void pyne::Material::norm_comp() {
 
 
 void pyne::Material::_load_comp_protocol0(hid_t db, std::string datapath, int row) {
-  // Clear current content
-  comp.clear();
-  
   hid_t matgroup = H5Gopen2(db, datapath.c_str(), H5P_DEFAULT);
   hid_t nucset;
   double nucvalue;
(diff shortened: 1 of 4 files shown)


gonuke:
If you rebase this against changes in `material.cpp` I can review

bam241:
I thought I did... but my local upstream/master was not up-to-date.... Sorry about that

bam241:
@gonuke merge done.

src/material.cpp
```diff
@@ -87,7 +90,7 @@ void pyne::Material::_load_comp_protocol1(hid_t db, std::string datapath,
   hid_t data_set = H5Dopen2(db, datapath.c_str(), H5P_DEFAULT);
 
   // Grab the nucpath
-  if (detect_nuclidelist(data_set, nucpath)){
+  If (detect_nuclidelist(data_set, nucpath)){
```
gonuke: oops!
```suggestion
  if (detect_nuclidelist(data_set, nucpath)){
```

src/material.cpp
```diff
@@ -2027,6 +1973,65 @@ pyne::Material pyne::Material::cram(std::vector<double> A,
...
+                       decay_const(i->first) * inhale_dose(i->first, source) / \
+                       atomic_mass(i->first);
+    }
+  } else {
+    throw std::invalid_argument("Dose type must be one of: ext_air, ext_soil, ingest, inhale.");
+  }
+  return dose;
+}
+
+#endif
```
gonuke: I think this syntax is OK, and it helps the reader find which `#if` statement this is ending.

```suggestion
#endif  PYNE_DECAY
```

src/material.h
```diff
@@ -422,7 +406,24 @@ namespace pyne
     /// Returns a mapping of the nuclides in this material to their atom densities.
     /// This calculation is based off of the material's density.
     std::map<int, double> to_atom_dens();
-
+#ifdef PYNE_DECAY
+    /// Calculates the activity of a material based on the composition and each
+    /// nuclide's mass, decay_const, and atmoic_mass.
+    comp_map activity();
```
gonuke: None of these rely on decay, do they?  Just on decay data which is found elsewhere.  Can't we keep these even if decay is not included?

src/material.cpp
```diff
@@ -45,6 +45,9 @@ void pyne::Material::norm_comp() {
 
 
 void pyne::Material::_load_comp_protocol0(hid_t db, std::string datapath, int row) {
+  // Clear current content
+  comp.clear();
```
gonuke: Is this related to the other changes in this PR????

gonuke (changes requested):
Oops - didn't include my changes in a review.

I have a few suggestions/requests.

src/material.h
bam241: we surely can

bam241: this is correct now only include `Material::decay()` in the `IFDEF`

src/material.cpp
bam241: some merge left over.

gonuke:
Moving those methods in & out of the DECAY option has made this look like a much bigger change than it is because a bunch of code is now just moved around in the file with no changes.  Can you streamline this to only include the changes that matter?

bam241:
> Moving those methods in & out of the DECAY option has made this look like a much bigger change than it is because a bunch of code is now just moved around in the file with no changes. Can you streamline this to only include the changes that matter?

This should be clean now :)

gonuke:
Thanks @bam241

gonuke (approved):

PR #1225, opened 2020-05-16
Fix issue #1224: error in serpent.parse_dep

### LINKED ISSUE #1224: Error in serpent.parse_dep's executable
In serpent.py's function parse_dep, the function runs into an error when trying to execute the file to create the parsed depletion file.  After finding this error trying to run my own dep.m file, I ran the [HGTR single-pebble burnup example](http://serpent.vtt.fi/mediawiki/index.php/HTGR_fuel_pebble_burnup_example), to check that it wasn't simply my error.

>  ---------------------------------------------------------------------------
> ValueError                                Traceback (most recent call last)
> <ipython-input-13-a61baff232d9> in <module>
> ----> 1 serpent.parse_dep('serpentexinput.inp_dep.m')
> 
> ~/.local/lib/python3.7/site-packages/pyne/serpent.py in parse_dep(depfile, write_py, make_mats)
>     286     # Execute the adjusted file
>     287     dep = {}
> --> 288     exec(f, dep, dep)
>     289     if '__builtins__' in dep:
>     290         del dep['__builtins__']
> 
> <string> in <module>
> 
> <string> in <listcomp>(.0)
> 
> ValueError: operands could not be broadcast together with shapes (44,) (16,) 

And still get the above error.

ZoeRichter:
This is a PR to (hopefully) fix an issue in parse_dep.  I ran the test_serpent script, but it would probably be wise to make a new test to double check the math.

Fixes #1224


katyhuff:
@ZoeRichter You can pass the news check by adding a news file to this PR. You can find a template and some examples in the news directory. These can be very brief, but are valuable in assembling release notes.

Your Py2 tests are failing as well. 


```bash

Testing test_rxname.py:

................................................................
Ran 64 tests in 0.004s

OK


Testing test_serpent.py:

..E..
ERROR: test_serpent.test_parse_dep1
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/root/repo/tests/test_serpent.py", line 39, in test_parse_dep1
    dep = serpent.parse_dep('sample_dep.m')
  File "/root/.local/lib/python2.7/site-packages/pyne/serpent.py", line 292, in parse_dep
    exec(f, dep, dep)
  File "<string>", line 11006, in <module>
TypeError: 'float' object has no attribute '__getitem__'

Ran 5 tests in 0.245s

FAILED (errors=1)
Testing failed on test_serpent.py

Exited with code exit status 1
CircleCI received exit code 1
```

ZoeRichter:
Does this error mean that this issue with __getitem__ is in the executable in parse_dep?  I can't find anywhere that uses or references it.

ZoeRichter:
Is it possible to see the input file that made sample_dep.m?  it has a few portions the _dep.m file I tested with doesn't, and I'm trying to figure out why.

ZoeRichter:
@robfairh
@andrewryh
@smpark7

ZoeRichter:
I've made the first attempt at detecting variable type in the parse_dep executable; still failing

@andrewryh

ZoeRichter:
Alright, I noticed an issue that I'm not really sure as to the cause.

When you let make_mats be true (default), parse_dep is meant to create an array for each mat in the `_dep.m` file ('fuel', 'matrix', etc) whose length is the number of burnup steps, and each entry is a pyne.Materials call to give the isotopic composition.  Then, it ends with a `TOT_ matrix`, whose length is the number of burnup steps, that gives the isotopic composition of all the materials combined together.

Here's the problem.  With my change to correct for the _VOLUME variable being an array (lines 268-272 in serpent.py), it correctly creates the `TOT_` matrix, but the matrices for the individual materials aren't creating pyne.Material calls, each entry is just the isotopic concentrations, in order, but not attached to the nucname.

Here's the code (lines 268-272 in serpent.py) that make these material matrices:
```
vol_name = re.search('.*_VOLUME = .*\]',f)
        if type(vol_name) is type(None):
            mat_gen_line = "{name}MATERIAL = [{name}VOLUME * Material(dict(zip(zai[:-2], {name}MDENS[:-2, col]))) for col in cols]\n"
        else:
            mat_gen_line = "{name}MATERIAL = [{name}VOLUME[col] * Material(dict(zip(zai[:-2], {name}MDENS[:-2, col]))) for col in cols]\n"
```
I'm really confused about what the issue could be.  In theory, {name}VOLUME[col] should just be a scalar, and the rest of the line is unchanged.

andrewryh:
> Alright, I noticed an issue that I'm not really sure as to the cause.
> 
> When you let make_mats be true (default), parse_dep is meant to create an array for each mat in the _dep.m file ('fuel', 'matrix', etc) whose length is the number of burnup steps, and each entry is a pyne.Materials call to give the isotopic composition. Then, it ends with a TOT_ matrix, whose length is the number of burnup steps, that gives the isotopic composition of all the materials combined together.
> 
> Here's the problem. With my change to correct for the _VOLUME variable being an array (lines 268-272 in serpent.py), it correctly creates the TOT_ matrix, but the matrices for the individual materials aren't creating pyne.Material calls, each entry is just the isotopic concentrations, in order, but not attached to the nucname.
> 
> Here's the code (lines 268-272 in serpent.py) that make these material matrices:
> 
> ```
> vol_name = re.search('.*_VOLUME = .*\]',f)
>         if type(vol_name) is type(None):
>             mat_gen_line = "{name}MATERIAL = [{name}VOLUME * Material(dict(zip(zai[:-2], {name}MDENS[:-2, col]))) for col in cols]\n"
>         else:
>             mat_gen_line = "{name}MATERIAL = [{name}VOLUME[col] * Material(dict(zip(zai[:-2], {name}MDENS[:-2, col]))) for col in cols]\n"
> ```
> 
> I'm really confused about what the issue could be. In theory, {name}VOLUME[col] should just be a scalar, and the rest of the line is unchanged.

Try to use the same solution, PyNE uses to parse `MAT_{name}_BURNUP`. Because it parses burnup correctly, just make it the same for volume.

gonuke:
Perhaps my question is a distraction from resolving this issue, am I correct in understanding that this method creates a string containing python code and then executes that code?  Is there a reason not to just execute the code directly?

ZoeRichter:
I didn't write parse_dep originally, but I believe it's because you need to go through the `_dep.m` file with regex to find the different parameters (`MAT_???_VOLUME`, etc) based on pattern matching, and regex returns what it finds as a string.  Since you can't pull the variables and their values out directly, it generates the script and then runs it.

andrewryh:
@ZoeRichter Looks like 68ed45a works well. Just try to figure out why tests are failed.

andrewryh:
@ZoeRichter Small tip: you can run tests on your own machine (no need to push it to Github every time):
`pytest /path/to/pyne/`

bam241:
those tests are designed to work with `nosetests` not `pytest` (last time I check `pytest` some were failing :)

Nosetests is deprecated, but we did not had the time yet to update them for `pytest`

andrewryh:
> those tests are designed to work with `nosetests` not `pytest` (last time I check `pytest` some were failing :)
> 
> Nosetests is deprecated, but we did not had the time yet to update them for `pytest`

Thanks, Baptiste!
Same for me, some tests failed when trying with pytest.
Then, probably pushing commits here is the only way to see tests output...

bam241:
if you have `docker` you can only run them in it. 
you can pull the exact same system that is running in the test here.

ZoeRichter:
Alright. provided these tests all pass, I just need to do some clean-up to get rid of files I used to debug that don't need to be in the final version, then I think it should be ready to go.

ZoeRichter:
@andrewryh do you have permission to pull prs?

pyne/serpent.py
```diff
@@ -177,6 +181,8 @@ def parse_res(resfile, write_py=False):
         with open(new_filename, 'w') as pyfile:
             pyfile.write(f)
 
+    print(f)
```
bam241: do we want to keep that print here ?

pyne/serpent.py
```diff
@@ -2,12 +2,16 @@
 import sys
 from warnings import warn
 from pyne.utils import QAWarning
- 
+
 import numpy as np
 
+import pdb
+
```
bam241: PEP8 recommends not blank lines between import
also absolute import should be before partial imports.

local import (`pyne`) should be last

pyne/serpent.py
```diff
@@ -365,8 +383,46 @@ def parse_det(detfile, write_py=False):
...
+    Returns
+    -------
+    corrected_f : str
+        The new version of the f executable that has the corrected
+        mat_gen_line variable, and all subsequent additions to f.
+
+    """
+
+    mat_gen_line = "{name}MATERIAL = [{name}vol[col] * Material(dict(zip(zai[:-2], {name}MDENS[:-2, col]))) for col in cols]\n"
```
bam241: the line here is a bit long, PEP8 suggest less that 80 chars.

pyne/serpent.py
```diff
@@ -365,8 +383,46 @@ def parse_det(detfile, write_py=False):
...
+    construct_string = ('\n\n# Construct materials\n'
+                        'zai = list(map(int, ZAI))\n'
+                        'cols = list(range(len(DAYS)))\n'
+                        '{name}vol = list(map(float, {name}VOLUME))\n')
+    base_names = re.findall('(MAT_\w*_)MDENS = ', quicksave_f)
+    for base_name in base_names:
+        footer += construct_string.format(name=base_name)
+    for base_name in base_names:
+        footer += mat_gen_line.format(name=base_name)
+    footer += "TOT_MATERIAL = [Material(dict(zip(zai[:-2], TOT_MASS[:-2, col]))) for col in cols]\n"
```
bam241: Line is too long there.

pyne/serpent.py
```diff
@@ -365,8 +383,46 @@ def parse_det(detfile, write_py=False):
             new_filename = detfile.name.rpartition('.')[0] + '.py'
         with open(new_filename, 'w') as pyfile:
             pyfile.write(f)
-    # Execute the adjusted file
+    # Execute the adjusted fileopen overwrite python w
```
bam241: some typos there.

pyne/serpent.py
```diff
@@ -365,8 +383,46 @@ def parse_det(detfile, write_py=False):
...
     det = {}
     exec(f, {}, det)
 
     return det
+
+def VOL_serp2_fix(quicksave_f, header):
```
bam241: PEP8 suggests to avoid missing CamelCase with "_".
our convention is lower case with underscore

pyne/serpent.py
```diff
@@ -365,8 +383,46 @@ def parse_det(detfile, write_py=False):
...
+    """Corrects for the _VOLUME arrays in a *_dep.m file being an array.
```
bam241: could you add more detail about what this correction brings, and when it is usefull.

pyne/serpent.py
```diff
@@ -261,6 +267,7 @@ def parse_dep(depfile, write_py=False, make_mats=True):
     # Add materials
     footer = ""
     if make_mats:
+        quicksave_f = copy.deepcopy(f)
```
bam241: maybe quicksave_f would be required if instead of overwriting f (line 282) you set a new variable ?

tests/test_serpent.py
```diff
@@ -56,6 +56,31 @@ def test_parse_dep1():
     assert_equal(dep['i952421'], 123)
     assert_array_equal(dep['MAT_fuelp1r2_H'][3], [0.00000E+00, 5.56191E-11, 3.22483E-10])
 
+def test_parse_dep2():
```
bam241, line 63: could you detail what the dep2 is testing, i.e. what does it bring vs the deps1

pyne/serpent.py
```diff
@@ -285,7 +292,18 @@ def parse_dep(depfile, write_py=False, make_mats=True):
...
+        exec(f, dep, dep)
+    except ValueError:
+        corrected_f = VOL_serp2_fix(quicksave_f, header)
+
+        # Overwrite the file made before to reflect changes
+        if write_py:
+            with open(new_filename, 'w') as pyfile:
+                pyfile.write(corrected_f)
+
+        exec(corrected_f,dep,dep)
```
bam241, line 305: Mainly PEP8 comments,

Just thinking, to avoid multiplying the code to form the footer, maybe we could have a single method to do it, that you call first without enablign the `vol` addition, and if the exec fails you call it back adding the `vol` into it ?
something like:
```
f_out = header + f + form_footer(add_vol=False)
try():
    exec...
catch(...):
    f_out = header + f + form_footer(add_vol=True)
    exec...
```

bam241 (changes requested):
some PEP8 comments,

I was thinking, to avoid multiplying the code to form the footer, maybe we could have a single method to do it, that you call first without enabling the `vol` addition, and if the exec fails you call it back adding the `vol` into it ?
something like:
```
f_out = header + f + form_footer(add_vol=False)
try():
    exec...
catch(...):
    f_out = header + f + form_footer(add_vol=True)
    exec...
```

pyne/serpent.py
ZoeRichter: I'm sorry, I don't think I understand what you mean.  I made a deep copy of f at that line to save a version that does not have any added materials.  Then, if the original mat_gen_line doesn't work, instead of finding and inserting the correct characters, I start with quicksave_f.  What would be the new variable?

bam241: see my example below, 

basically you can keep the `f` variable untouched and just make one with the header and the footer: `f_full = header + f + footer`

bam241: instead of `f = header + f + footer`

ZoeRichter:
> some PEP8 comments,
> 
> I was thinking, to avoid multiplying the code to form the footer, maybe we could have a single method to do it, that you call first without enabling the `vol` addition, and if the exec fails you call it back adding the `vol` into it ?
> something like:
> 
> ```
> f_out = header + f + form_footer(add_vol=False)
> try():
>     exec...
> catch(...):
>     f_out = header + f + form_footer(add_vol=True)
>     exec...
> ```

I thought this footer is only made if make_mats is true - isn't an extra argument unneeded? I also only intended this PR to fix issue #1224, which is targeted at a specific bug.  Maybe we could make a separate issue for this sort of cleanup/optimization?

pyne/serpent.py
ZoeRichter: Ah, because the mat_gen_line is only in the footer, and f_full is a new variable, so you don't have to worry about one changing the other.  That makes sense, thanks!

pyne/serpent.py
ZoeRichter: Nope, thanks for catching that!

pyne/serpent.py
ZoeRichter:
```
import re
import sys
import numpy as np
import pdb
import copy
from warnings import warn
from pyne.utils import QAWarning
```
Like this?

pyne/serpent.py
ZoeRichter: Got it

pyne/serpent.py
ZoeRichter: split into multiple lines

bam241:
> > some PEP8 comments,
> > I was thinking, to avoid multiplying the code to form the footer, maybe we could have a single method to do it, that you call first without enabling the `vol` addition, and if the exec fails you call it back adding the `vol` into it ?
> > something like:
> > ```
> > f_out = header + f + form_footer(add_vol=False)
> > try():
> >     exec...
> > catch(...):
> >     f_out = header + f + form_footer(add_vol=True)
> >     exec...
> > ```
> 
> I thought this footer is only made if make_mats is true - isn't an extra argument unneeded? I also only intended this PR to fix issue #1224, which is targeted at a specific bug. Maybe we could make a separate issue for this sort of cleanup/optimization?

Sadly if this is not cleaned up in this, it will probably never be done :)

yes you'd probably need more arguments.
you could change your method into:
```
def from_footer(quicksave_f, header, add_vol =False):
mat_gen_line = "{name}MATERIAL = [{name}vol[col] * Material(dict(zip(zai[:-2], {name}MDENS[:-2, col]))) for col in cols]\n"
    footer = ""
    construct_string = ('\n\n# Construct materials\n'
                        'zai = list(map(int, ZAI))\n'
                        'cols = list(range(len(DAYS)))\n'
    if(add_vol):
                        construct_string += '{name}vol = list(map(float, {name}VOLUME))\n')
    base_names = re.findall('(MAT_\w*_)MDENS = ', quicksave_f)
    if(add_vol):
        for base_name in base_names:
            footer += construct_string.format(name=base_name)
    for base_name in base_names:
        footer += mat_gen_line.format(name=base_name)
    footer += "TOT_MATERIAL = [Material(dict(zip(zai[:-2], TOT_MASS[:-2, col]))) for col in cols]\n"
    footer += "del zai, cols\n"

    # Add header & footer to file
    corrected_f = header + quicksave_f + footer

    return corrected_f
```
and then call it twice, one with `form_footer(...)`
and then `form_footer(..., add_vol = True)` if the first exec failed

this would simplify the implementation and avoid redundancy...

pyne/serpent.py
ZoeRichter: Fixed!

pyne/serpent.py
bam241: yes ! :)

pyne/serpent.py
ZoeRichter: changed to 'vol_serp2_fix'

pyne/serpent.py
ZoeRichter: How is
"""Corrects for the _VOLUME arrays in a *_dep.m file being an array.
    This happens when a Serpent 2 _dep.m file is parsed and make_mats
    is True.

    Parameters
    ----------
    f : str
        f from parse_dep without the header or footer added in
        the make_mats if statement
    header : str
        header is the variable of the same name in parse_dep

    Returns
    -------
    corrected_f : str
        The new version of the f executable that has changed
        {name}VOLUME to {name}vol[col].  Also adds the vol
        variable, which is the VOLUME matrix mapped to an int list.

    """

tests/test_serpent.py
ZoeRichter: I added this to test_parse_dep1
    '''
    Tests the parse_dep function with a serpent 1 _dep.m
    file.  In this, the _VOLUME variable is a float.
    '''
And this to test_parse_dep2:
    '''
    Tests the parse_dep function with a simple _dep.m file,
    sample2_dep.m, which was generated using the 2d pin burnup
    example from the serpent wiki, changed to have fewer depsteps.
    In this, the _VOLUME variable is a numpy array.
    http://serpent.vtt.fi/mediawiki/index.php/2D_PWR_pin-cell_burnup_example
    '''

ZoeRichter:
I would add the from_footer call into vol_serp2_fix?  adding the line that maps _VOLUME to an integer list named _vol is required for it to work, just the same as 'zai' and 'cols'.  If the exception happened and vol_serp2_fix needs to be used, there will never be a scenario where you wouldn't have to add it.

Unless you mean somewhere else!

bam241:
> I would add the from_footer call into vol_serp2_fix? adding the line that maps _VOLUME to an integer list named _vol is required for it to work, just the same as 'zai' and 'cols'. If the exception happened and vol_serp2_fix needs to be used, there will never be a scenario where you wouldn't have to add it.
> 
> Unless you mean somewhere else!

the form_footer() would replace the vol_serp2_fix()

the implementation would look like:
```
def parse_dep(depfile, write_py=False, make_mats=True):
    [...]
    if make_mats:
       footer = form_footer(f)
    file = header + f + footer
    [...]
    try:
    [...]
    except ValueError:
        footer = form_footer(f, add_vol = True)
        file = header + f + footer
    [...]
```
with the `form_footer` method:
```
def from_footer(f,, add_vol =False):
    mat_gen_line = "{name}MATERIAL = [{name}vol[col] * Material(dict(zip(zai[:-2], {name}MDENS[:-2, col]))) for col in cols]\n"
    footer = ""
    construct_string = ('\n\n# Construct materials\n'
                        'zai = list(map(int, ZAI))\n'
                        'cols = list(range(len(DAYS)))\n'
    if(add_vol):
                        construct_string += '{name}vol = list(map(float, {name}VOLUME))\n')
    base_names = re.findall('(MAT_\w*_)MDENS = ', quicksave_f)
    if(add_vol):
        for base_name in base_names:
            footer += construct_string.format(name=base_name)
    for base_name in base_names:
        footer += mat_gen_line.format(name=base_name)
    footer += "TOT_MATERIAL = [Material(dict(zip(zai[:-2], TOT_MASS[:-2, col]))) for col in cols]\n"
    footer += "del zai, cols\n"

    return footer
```

ZoeRichter:
Oh!  I think I see what you mean.  Make one function that can make either version of the footer, with an extra flag for whether or not you need to?

So you'd have:

full_f = add_footer(f, header, add_vol = False)

This makes the original version of f.

At the bottom, you'd have

`Try:
exec(full_f, dep, dep)

except ValueError
corrected_f = add_footer(f, header, add_vol = True)
[lines for write_py]
exec(corrected_f, dep, dep)`

bam241:
don't even need to pass the header, just pass the f and return the footer

ZoeRichter:
You would have to also change mat_gen_line based on whether or not you need add_vol to be true, but that would just be an extra line.

Edit: maybe a better name for add_vol is serp2, since the changes are based on the version of serpent the dep file comes from

bam241:
> Edit: maybe a better name for add_vol is serp2, since the changes are based on the version of serpent the dep file comes from

I let you decide which name are the best, I actually don't know a lot about Serpent/Serpent2

pyne/serpent.py
```diff
@@ -45,12 +45,57 @@
...
+    '''
+
+    if serp2:
+        mat_gen_line = ('{name}MATERIAL = [{name}vol[col] *'
+                        'Material(dict(zip(zai[:-2], {name}MDENS[:-2, col])))'
+                        ' for col in cols]\n')
+    else:
+        mat_gen_line = ('{name}MATERIAL = [{name}VOLUME *'
+                        'Material(dict(zip(zai[:-2], {name}MDENS[:-2, col])))'
+                        ' for col in cols]\n')
```
bam241:
```suggestion
    vol_name = ""
    if serp2:
        vol_name = "vol[vol]"
    else:
        vol_name = "VOLUME"

    mat_gen_line = ('{name}MATERIAL = [{name}' + vol_name + ' *'
                   'Material(dict(zip(zai[:-2], {name}MDENS[:-2, col])))'
                   ' for col in cols]\n')
```

pyne/serpent.py
```diff
@@ -321,7 +371,17 @@ def parse_det(detfile, write_py=False):
     # Replace matlab Arrays
     f = _replace_arrays(f)
 
-    # Find detector variable names
+    # Find detector variable namesmat_gen_line = "{name}MATERIAL =
+    # [{name}VOLUME * Material(dict(zip(zai[:-2], {name}MDENS[:-2, col]))) for
+    # col in cols]\n"
```
bam241: if this is not needed any more maybe just remove it.

pyne/serpent.py
bam241: I think this would simplify implentation and readability....

pyne/serpent.py
```diff
@@ -281,11 +319,23 @@ def parse_dep(depfile, write_py=False, make_mats=True):
...
+            pyfile.write(full_f)
 
     # Execute the adjusted file
     dep = {}
-    exec(f, dep, dep)
+    try:
+        exec(full_f, dep, dep)
+    except ValueError:
+        correct_footer = form_footer(f, serp2=True)
+        corrected_f = header + f + correct_footer
```
bam241: here you can probably overwrite the `full_f` to reduce the memory footprint.

bam241 (commented):
Looking good.
2 additional minor comments it this should be good for me.

Great job !

pyne/serpent.py
ZoeRichter: Fixed the second comment, what change are you suggesting for the first?

ZoeRichter:
@katyhuff  this should be good to go after this round of changes!

pyne/serpent.py
bam241: you should be able to see it [here](https://github.com/pyne/pyne/pull/1225/files#r436215045)

pyne/serpent.py
```diff
@@ -281,11 +319,23 @@ def parse_dep(depfile, write_py=False, make_mats=True):
...
-            pyfile.write(f)
+            pyfile.write(full_f)
 
     # Execute the adjusted file
     dep = {}
-    exec(f, dep, dep)
+    try:
+        exec(full_f, dep, dep)
+    except ValueError:
+        correct_footer = form_footer(f, serp2=True)
```
bam241, line 340: maybe add a comment to explain what is attempted if the first `exec` fails

bam241 (approved):
Looking great !

thx a lot for fixing this!!

pyne/serpent.py
katyhuff: looks handled.

katyhuff (approved):

katyhuff:
Any objections to merging @bam241 ? Looks good to me!

gonuke:
Thanks @ZoeRichter - nice work on a tricky bit of code.

gonuke:
I think we just needed an ARFC approval since we don't really use this part of the code.

PR #1228, opened 2020-06-05
Add flags information on install from sources docs

bam241:
update the list of optional dependencies (I might be missing some...)

and add information about the different available flags.

### DIFF
--- a/docs/install/linux_source.rst
+++ b/docs/install/linux_source.rst
@@ -56,39 +56,39 @@ Most commun install flags:
 The list of the possible flags can be retrieve using:
 
 unzipped directory::
-      python setup.py --help
+   python setup.py --help
 
 
-   #. optional arguments:
-      -h, --help            show this help message and exit
-      --clean [CLEAN]       removes the build directory before continuing.
-      --user [USER]         Installs into ~/.local
+#. optional arguments:
+   -h, --help            show this help message and exit
+   --clean [CLEAN]       removes the build directory before continuing.
+   --user [USER]         Installs into ~/.local
 
-   #. cmake:  CMake arguments.
+#. cmake:  CMake arguments.
 
-      -D VAR                Set environment variable.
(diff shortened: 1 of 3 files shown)


docs/install/linux_source.rst
```diff
@@ -46,6 +47,46 @@ do its best to find relevant nuclear data elsewhere on your machine
 or from public sources on the internet.
 
 
+Most commun install flags:
```
ljacobson64: typo

docs/install/linux_source.rst
```diff
@@ -23,7 +23,8 @@ PyNE has the following dependencies:
 
 Optional Depenendencies:
    #. `MOAB <https://press3.mcs.anl.gov/sigma/moab-library>`_
-   #. `PyTAPS <https://pythonhosted.org/PyTAPS/index.html>`_
+   #. `pyMOAB <https://press3.mcs.anl.gov/sigma/moab-library>`_
```
gonuke, line 27: I wonder if this should be listed as a part of MOAB rather than its own line.  It is not really a separate dependency, and you need pyMOAB to use MOAB

docs/install/linux_source.rst
```diff
@@ -46,6 +47,46 @@ do its best to find relevant nuclear data elsewhere on your machine
 or from public sources on the internet.
 
 
+Most common install flags:
+**************************
+The list of the possible flags can be retrieve using:
```
gonuke:
```suggestion
The list of the possible flags can be retrieved using:
```

gonuke (changes requested):
Thanks for this addition @bam241

docs/install/linux_source.rst
gonuke: ping @bam241

gonuke:
This should be coordinated with #1238 by @nsryan2 - perhaps only the install flags section needs to be retained from this, given his work

bam241:
this has been combined with #1238.
new combined PR is #1315.
Closing.

PR #1230, opened 2020-06-15
bash script to build and test in docker

bam241:
Might not be the best implementation.
I am opened to suggestion to make this better

### DIFF
--- a/docker_build.bash
+++ b/docker_build.bash
@@ -1,0 +1,35 @@
+#! /bin/bash
+
+
+function test_img_name()
+{
+    if [[    "$1" != *"pyne/ubuntu_18.04_py2_pyne-deps"* \
+          && "$1" != *"pyne/ubuntu_18.04_py2_dagmc_pyne-deps"* \
+          && "$1" != *"pyne/ubuntu_18.04_py2_pymoab_pyne-deps"* \
+          && "$1" != *"pyne/ubuntu_18.04_py2_dagmc_pymoab_pyne-deps"* \
+          && "$1" != *"pyne/ubuntu_18.04_py3_pyne-deps"* \
+          && "$1" != *"pyne/ubuntu_18.04_py3_dagmc_pyne-deps"* \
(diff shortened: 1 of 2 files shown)


gonuke:
Might be good to discuss if this is better than just documentation?

docker_build.bash
```diff
@@ -0,0 +1,35 @@
...
+    echo "pyne/ubuntu_18.04_py3_pymoab_pyne-deps"
+    echo "pyne/ubuntu_18.04_py3_dagmc_pymoab_pyne-deps"
+    echo "pyne/ubuntu_18.04_py3_dagmc_pymoab_openmc_pyne-deps"
+  fi
+}
+
+test_img_name $1
+
+docker pull $1
+docker run -v $PWD:/pyne -i -w "/pyne" -t $1 /bin/bash  -c "python setup.py install --prefix=~/.local -j18 --clean"
```
gonuke, line 32: `-j18` is a lot for some people's machines!

docker_build.bash
```diff
@@ -0,0 +1,35 @@
...
+commit_id=$(docker ps -lq | tail -n1)
+docker commit ${commit_id} $1
+docker run -v $PWD:/pyne -i -w "/pyne" -t $1 /bin/bash
```
gonuke, line 35: Is this necessary?

bam241: I usually use this to run the `nosetests` interactively.
I usually prefer to be able to run them interactively, it might not be necessary....

docker_build.bash
bam241: fair point, just copy/pasted the command I usually run....

bam241:
> Might be good to discuss if this is better than just documentation?

Yes, 
 I guess the goal of this is to move forward, and decide what we want to do :)

gonuke:
I think this is best in documentation.  Basically we need to provide documentation that describes how to:
1. pull & run the appropriate docker image opening a simple bash session
2. what command to run in the bash session of that image to do the build
3. what command to run in the bash session of that image to do the testing

Does this fit into @nsryan2's documentation overhaul #1238?

Issue #1232, opened 2020-06-22
Pyne is deprecated as of python 3.7


ZoeRichter:
I have python version 3.7.4 installed on my machine running Ubuntu 18.04.  When I import pyne, I get the following block of warnings:
```
/home/zoe/anaconda3/lib/python3.7/site-packages/ipykernel_launcher.py:5: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  """
/home/zoe/anaconda3/lib/python3.7/importlib/_bootstrap.py:219: QAWarning: pyne.data is not yet QA compliant.
  return f(*args, **kwds)
/home/zoe/anaconda3/lib/python3.7/importlib/_bootstrap.py:219: QAWarning: pyne.material is not yet QA compliant.
  return f(*args, **kwds)
/home/zoe/.local/lib/python3.7/site-packages/pyne/serpent.py:11: QAWarning: pyne.serpent is not yet QA compliant.
  warn(__name__ + " is not yet QA compliant.", QAWarning)

```

If it will stop working as of 3.8, I'm thinking that there should be something to change the ABC implementation for users with python 3.7+

gonuke:
Thanks for pointing this out @ZoeRichter - I have a fix in the works in PR #1233

ZoeRichter:
It looks like #1233 has been merged, so I'll close this issue.

PR #1233, opened 2020-06-23
Fix import of collections

gonuke:
Beginning with python 3.8, the `collections` module needs to be imported differently.  This was identified in #1232 .

This should replace all imports to `collections` to be safe for the changes in python3.8.

It also removes some places where `collections` was imported and never used.


pyne/openmc_utils.py
```diff
@@ -5,7 +5,10 @@
 import io
 import sys
 from warnings import warn
-from collections import namedtuple
+try:
+    from collections.abc import namedtuple
+except ImportError:
+    from collections import namedtuple
```
bam241, line 11: I think this one did't need to be changed. as it does not rely on `abc` in the first place.

pyne/rxdata.py
```diff
@@ -1,4 +1,3 @@
-import collections
```
bam241, line 1: needed L19

bam241 (approved):
LGTM!
thx @gonuke

bam241:
@gonuke could you verify you agree with my conflict resolution https://github.com/pyne/pyne/pull/1233/commits/d7ea237bfc8bf6c8cebe3d369979109358c4a12b

(basically removed the outdated `MaterialLibrary`  implementation in `material.pyx`)

if you agree, I am ready to merge.

gonuke:
I had to add the fix to your new `material_library` code

bam241:
LGTM!,
thx @gonuke !
merging

PR #1242, opened 2020-07-21
Enable docker build with custom version of HDF5.

gonuke:
Also clean up/streamline some of the build process.

I'm not sure we need this for CI in the long term(??), but it will definitely help test #1220

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -346,7 +346,7 @@ workflows:
       - py3_dagmc_pymoab_hdf5_12_build:
           requires:
             - py3_dagmc_pymoab_build
-      - py3_dagmc_pymoab_test:
+      - py3_dagmc_pymoab_hdf5_12_test:
           requires:
             - py3_dagmc_pymoab_hdf5_12_build
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,9 @@ executors:
     py3_dagmc_pymoab:
         docker:
             - image: pyne/ubuntu_18.04_py3_dagmc_pymoab_pyne-deps:latest
+    py3_dagmc_pymoab_hdf5_12:
+        docker:
+            - image: pyne/ubuntu_18.04_py3_hdf5-1_12_0_dagmc_pymoab_pyne-deps:latest
     py3_dagmc_pymoab_openmc:
         docker:
             - image: pyne/ubuntu_18.04_py3_dagmc_pymoab_openmc_pyne-deps:latest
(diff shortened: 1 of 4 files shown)


docker/ubuntu_18.04-dev.dockerfile
```diff
@@ -58,50 +58,67 @@ RUN if [ "${py_version%.?}" -eq 3 ] ; \
...
+        && cd .. \
+        && rm -rf hdf5; \
+    fi
+# put MOAB on the path
+ENV LD_LIBRARY_PATH $HDF5_INSTALL_PATH/lib:$LD_LIBRARY_PATH
+ENV LIBRARY_PATH $HDF5_INSTALL_PATH/lib:$LIBRARY_PATH
+
+ARG build_moab=NO
+ARG enable_pymoab=NO
+ENV INSTALL_PATH=$HOME/opt/moab
```
bam241, line 88: I believe than ARG command must be placed just before line they are used, 
I am unsure if having a ENV in-between could be a problem.

you might need to redeclare `build_hdf5` as an argument there along with `build_moab` and `enable_moab`

gonuke: This is not a problem - everything works as it is....

.circleci/config.yml
```diff
@@ -323,6 +343,14 @@ workflows:
           requires:
             - py3_dagmc_pymoab_build
 
+      - py3_dagmc_pymoab_hdf5_12_build:
```
gonuke, line 346: @bam241 How do I make it so that this only produces a warning - and not a failure?

gonuke:
This is supposed to fail on HDF5 v1.12,  working on amending it to produce a warning instead of a failure

.circleci/config.yml
gonuke: Or is that a Travis thing?

bam241: I don't think this exists in CircleCI.

but we can decide to block or not merge based on the results of some tests.

docker/ubuntu_18.04-dev.dockerfile
bam241: I had to make sure...

I was convinced that during one of my previous experience with docker `ARGS` I add to re-add the `ARG MY_VAR=XX` before each command that was using the variable....

I tested and apparently not...

bam241:
SO apparently allow failure is not a feature in CircleCI, 
but we can do the opposite: block merge if one of the other tests is failing
in the GitHub repo: `settings/Branches/"Branch protection rules"`

gonuke:
The current test failure is expected and will be resolved by #1220.  I think we should merge this and then see how #1220 performs in CI.

bam241:
sounds about right! merging !

bam241:
thx @gonuke

PR #1248, opened 2020-08-10
simplify throw return

bam241:
Someone of the conda gitter suggested that this might be the source or the problem on MacOS tests....


gonuke:
Thanks @bam241 - Let's see if this fixes conda for OSX?

PR #1257, opened 2020-08-11
Issue1164

### LINKED ISSUE #1164: Half life time should be nan if nuclide isn't found in database
When you try to retrieve the half life time of an unknown isotope 'inf' is returned. For example
`nucdat.half_life("Cm-253")` produces the output 'inf', while this isotope cannot be found in the nudat2 data https://www.nndc.bnl.gov/nudat2/chartNuc.jsp . 
Wouldn't it make more sense if pyne would return 'nan' in such cases, like it does for known isotopes with unknown half life time (e.g. U-241)?

ZoeRichter:
Answers issue #1164 and changes half_life in data.pyx to return nan instead of inf if the isotope cannot be found.  Previous behavior assumed nuclides not in data libraries are stable.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -134,7 +134,7 @@ v0.7.0 RC2
    * ENSDF database link to 2019 Oct 4th database
    * Update the C012-n.ace file link.
    * Missing elements name_to_zz dictionary
-   * Updated half_life in data.pyx to return nan if isotope not found
+   * Updated half_life in data.pyx to return nan if isotope not found (#1257)
 
 * Improvements in Mesh capabilities
    * added mesh tally definitions to tallies
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,7 @@ v0.7.0 RC2
 
 **New Capabilities**
 
-* Subvoxel R2S allows activation and photon source sampling by 
+* Subvoxel R2S allows activation and photon source sampling by
   material within a mesh voxel
   * SubMode parameter in source_sampling
   * Add code for submode: SUBVOXEL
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -134,6 +134,7 @@ v0.7.0 RC2
    * ENSDF database link to 2019 Oct 4th database
    * Update the C012-n.ace file link.
    * Missing elements name_to_zz dictionary
+   * Updated half_life in data.pyx to return nan if isotope not found
```
nsryan2:
```suggestion
   * Updated half_life in data.pyx to return nan if isotope not found (#1257)
```

nsryan2 (commented):
Other than adding the #PR to the changelog, LGTM!

gonuke:
So... because of our rebasing workflow - you'll need to rebase this on the current version of the `0.7.0-rc` branch.  Is that something you're familiar with?  If not, we can walk you through it.

ZoeRichter:
I'm not familiar with rebasing, no.

ZoeRichter:
Based a conversation with @gonuke , there is a discussion about how to handle isotopes not in the library.
`from pyne import data
import numpy as np

#F19 is stable
print(data.half_life('F19'))
print(data.half_life('F19', use_metastable=False))
print(data.half_life('F19', use_metastable=True))
print()
#U235 is a known radioisotope
print(data.half_life('U235'))
print(data.half_life('U235', use_metastable=False))
print(data.half_life('U235', use_metastable=True))
print()
#Tb136 is a radioisotope with an unknown half-life
print(data.half_life('Tb136'))
print(data.half_life('Tb136', use_metastable=False))
print(data.half_life('Tb136', use_metastable=True))`

Is a quick example showing the three isotope cases we might see - known and radioactive, unknown and radioactive, and unknown and stable.  U235 outputs what you would expect, but calling data.half_life on F19 and Tb136 both return inf.  For all three isotopes. I saw no difference in output based on the use_metastable flag.

gonuke:
I think the difference occurs if you pass in a metastable nuclide label, e.g. `Cf252m`

gonuke:
Thanks for the contribution on this @ZoeRichter

Issue #810, opened 2016-01-21
Ipython not actually needed for setup


makeclean:
Right now we have an unofficial dependency which is IPython, not listed at www.pyne.io. My guess is is that its actually needed for tests or building documentation? We should therefore move the check in https://github.com/pyne/pyne/blob/develop/setup.py#L153 to a more appropriate place?

skc222:
@makeclean @katyhuff   I tested all sample tests without installing IPython. Does this mean that this issue is no longer open? or can you tell me why IPython is an unofficial dependency then?

makeclean:
Guess so, its possible something change between then and now?

skc222:
Ohh yes. You are correct. We don't need ipython for unit testing but we need to install it for example testing. Many other packages are also required for example testing. I installed all but still got following error:
```
FAILURE: ace.ipynb:
CAT: ipython nbconvert --to=python --stdout ace.ipynb
----------
Traceback (most recent call last):
  File "../execer.py", line 65, in main
    execer(f)
  File "../execer.py", line 39, in execipynb
    exec(compile(src, filename, "exec"), glb, loc)
  File "ace.ipynb", line 30, in <module>
    "    urlretrieve(\"https://www-nds.iaea.org/wolfram/w180/beta3/W180.ace\", \"W180.ace\")"
  File "pyne/ace.pyx", line 181, in pyne.ace.Library.read (/home/ubuntu/Downloads/pyne-0.5.3/build/pyne/ace.cxx:5440)
  File "pyne/ace.pyx", line 288, in pyne.ace.Library._read_ascii (/home/ubuntu/Downloads/pyne-0.5.3/build/pyne/ace.cxx:7416)
ValueError: could not convert string to float: html>

FAILURE: discretized_teapot.ipynb:
CAT: ipython nbconvert --to=python --stdout discretized_teapot.ipynb
----------
Traceback (most recent call last):
  File "../execer.py", line 65, in main
    execer(f)
  File "../execer.py", line 39, in execipynb
    exec(compile(src, filename, "exec"), glb, loc)
  File "discretized_teapot.ipynb", line 11, in <module>
    {
ImportError: No module named dagmc

FAILURE: endf_reader.ipynb:
CAT: ipython nbconvert --to=python --stdout endf_reader.ipynb
----------
Traceback (most recent call last):
  File "../execer.py", line 65, in main
    execer(f)
  File "../execer.py", line 39, in execipynb
[...]

PR #1262, opened 2020-08-11
iPython --> Jupyter (Issue859)

### LINKED ISSUE #859: IPython notebooks now called Jupyter notebooks
A few places in the docs should be updated to reflect this.

Should the hosted examples themselves still be called IPython notebooks since they're .ipynb files?

opotowsky:
Addresses Issue #859 

Note: 
This PR only changes "external" facing mentions, not "internal": it ignores `.. code-block:: ipython` in .rst files because I don't think there's a simple way to solve that (embedded ipython code is in `docs/sphinxext/ipython_directive.py)`

Question for reviewer: 
Should I change the [code](https://github.com/opotowsky/pyne/blob/issue859/setup.py#L104) in `setup.py` to check for jupyter v4.0+ instead of ipython?
Edit: Or, referring to Issue #810, remove it completely?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -179,6 +179,7 @@ v0.7.0 RC2
     instead of "C++ & Fortran API Documentation"
   * Added publications to bibliography (PR #1256)
   * Adding contributing guide and code of conduct (#1258)
+  * Changed Doc and Tutorial mentions of iPython notebooks to Jupyter notebooks (PR #1262)
     
 * Improvements in building and testing
   * require contributor to change CHANGELOG
--- a/tutorial/readme.rst
+++ b/tutorial/readme.rst
@@ -5,21 +5,21 @@ The Great PyNE Tutorial
 
 **Contents:**
 
-1. `Intro <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/00-intro.ipynb>`_
-2. `Nuclide Naming <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/01-nuclide-naming.ipynb>`_
-3. `Materials <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/02-materials.ipynb>`_
-4. `Nuclear Data <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/03-nuclear-data.ipynb>`_
-5. `Reaction Names <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/04-reaction_names.ipynb>`_
-6. `Cross Sections <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/05-cross-sections.ipynb>`_
-7. `Transmutation <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/06-transmutation.ipynb>`_
-8. `Mesh Basics <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/07-mesh-basics.ipynb>`_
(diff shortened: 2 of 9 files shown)


gonuke:
How did this pass the changelog test?

gonuke:
Let's leave the fix for #810 to a different discussion

tutorial/readme.rst
```diff
@@ -5,21 +5,21 @@ The Great PyNE Tutorial
...
-2. `Nuclide Naming <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/01-nuclide-naming.ipynb>`_
-3. `Materials <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/02-materials.ipynb>`_
-4. `Nuclear Data <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/03-nuclear-data.ipynb>`_
-5. `Reaction Names <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/04-reaction_names.ipynb>`_
-6. `Cross Sections <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/05-cross-sections.ipynb>`_
-7. `Transmutation <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/06-transmutation.ipynb>`_
-8. `Mesh Basics <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/07-mesh-basics.ipynb>`_
-9. `Diffusion <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/08-diffusion.ipynb>`_
-10. `Outro <http://nbviewer.ipython.org/urls/raw.github.com/pyne/tutorial/master/09-outro.ipynb>`_
+1. `Intro <http://nbviewer.jupyter.org/urls/raw.github.com/pyne/tutorial/master/00-intro.ipynb>`_
```
opotowsky: link that actually opens is:
https://nbviewer.jupyter.org/github/pyne/tutorial/blob/master/*

Should I change these to that?

gonuke: If that works to open the notebooks, then I think it's good to change to that.  It's possible that this file was originally written at a time when Github didn't render notebooks correctly/automatically

CHANGELOG.rst
```diff
@@ -167,56 +167,57 @@ v0.7.0 RC2
 **Maintenance**
 
 * Documentation Changes
-  * Credit Rochman for allowing redistribute TENDL file
-  * Fix various typos
-  * automatic deployment of a updated version of the website on tags
-  * automatic creation of a new version of the website (not deployed) for
+   * Credit Rochman for allowing redistribute TENDL file
```
gonuke: This whitespace change may be "right" for RST, but I'm not sure I want to put it in now as it will wreak havoc on all our other PRs.  Not sure if you have a way to produce this CHANGELOG without the whitespace changes?

gonuke (changes requested):
Would be great if we can avoid the whitespace changes as part of this PR - maybe we can make it a different PR by itself?

tutorial/readme.rst
opotowsky: Great, will do

CHANGELOG.rst
opotowsky: Actually those were intentional because the formatting was messed up. I can submit that as a separate PR, though

opotowsky: I actually noticed even more formatting errors after this too -- but it definitely makes more sense to have it separate

gonuke (approved):

gonuke:
Thanks @opotowsky

Issue #1144, opened 2019-06-12
CI-docker python package


bam241:
Maybe we should install all the python dependencies with `pip` as suggested in pyne/install_scripts#32 
As we can see [[see](https://github.com/pyne/pyne/blob/develop/docker/ubuntu_18.04-dev.dockerfile)], 10 python package are installed with `pip`:
- `sphinx`
- `cloud_sptheme`
- `prettytable`
- `sphinxcontrib_bibtex`
- `numpydoc`
- `nbconvert`
- `numpy`
- `nose`
- `cython`
- `future`

and 4+1 are still installed with `apt-get`:
- `nose` (yes installed with `apt-get` then overwritten by `pip`)
- `matplotlib`
- `tables`
- `scipy`
- `jinja2`

gonuke:
closed by #1263

PR #1263, opened 2020-08-11
(Docker) Install all Python packages with Pip

epassaro:
Following https://github.com/pyne/pyne/issues/1144.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -203,6 +203,8 @@ v0.7.0 RC2
    * test file for ENDF was wrong
   * Add functions to do file, file block, line, and string almost the same
     compare functions in pyne/utils.py
+  * Install Python dependencies with Pip instead of APT in Dockerfile
+  * Remove if block in travis-run-tests.sh (just run nosetests)
 
 * Code cleanup
    * Formatting improvements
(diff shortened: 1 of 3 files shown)


bam241:
looks good to me, just an update of the changeling is missing
Have you try to build a docker image with this ?

epassaro:
I do not have so much experience with Docker. I made an image using the DockerHub web interface using the dockerfile from my fork: https://hub.docker.com/r/epassaro/pyne.

Also I should modify the changelog manually?

opotowsky:
Yep you modify the changelog manually!

gonuke:
Either @bam241 or I can test your dockerfile, but it looks like the one on Dockerhub used the old file: https://hub.docker.com/r/epassaro/pyne/dockerfile

gonuke:
Oops - I was wrong - it is using a new Dockerfile!

gonuke:
It looks like when we pip install `nose` with python3, the executable is not called `nosetests3`.  So this breaks `travis-run-tests.sh` and we can't quite run tests with the current CI setup.  @bam241  - we should discuss the remedy to this....

bam241:
Have you tried the image @epassaro linked ? this image is built with python2.7.  so their will be no `nosetests3` in it

have you tried it on an other docker image ?

bam241:
I have the same issue on Mac, nose tests installed through pip come as nosetests-3.x not nosetests3...

bam241:
I am note sure that the travis_tests.sh is really useful... it basically only run the nosetests....

bam241:
working ! thx @epassaro for all the work.

Merging !

PR #1264, opened 2020-08-11
An other attempt to fix nosetests on OSX-Conda

bam241:
@gonuke, I need to confess that I am throwing darts in the dark here.

I am trying to fix a warning that I don't understand, and don't observe on my own machine....
here is the warning:
`/Users/runner/miniforge3/conda-bld/pyne_1597167143946/work/src/utils.h:218:14: warning: address of stack memory associated with local variable 'msgstr' returned [-Wreturn-stack-address]
`
and I used [this](https://flaviocopes.com/c-return-string/) as an inspiration

### DIFF
--- a/src/nucname.h
+++ b/src/nucname.h
@@ -138,7 +138,7 @@ namespace nucname
         NaNEstr += " --> ";
         NaNEstr += nucnow;
       }
-      return (const char *) NaNEstr.c_str();
+      return NaNEstr.c_str();
     };
 
   private:
@@ -206,7 +206,7 @@ namespace nucname
         INFEstr += " --> ";
         INFEstr += nucnow;
       }
-      return (const char *) INFEstr.c_str();
+      return INFEstr.c_str();
     }
 
   private:
--- a/src/nucname.h
+++ b/src/nucname.h
@@ -138,7 +138,8 @@ namespace nucname
         NaNEstr += " --> ";
         NaNEstr += nucnow;
       }
-      return NaNEstr.c_str();
+      const char* NaNEstr_rtn = NaNEstr.c_str();
+      return NaNEstr_rtn;
     };
 
   private:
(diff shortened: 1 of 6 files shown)


bam241:
and btw we are unsure if this is actually the source of the problem....

src/nucname.h
```diff
@@ -138,7 +138,8 @@ namespace nucname
         NaNEstr += " --> ";
         NaNEstr += nucnow;
       }
-      return (const char *) NaNEstr.c_str();
+      const char* NaNEstr_rtn = NaNEstr.c_str();
```
ngranda, line 141: This may be obvious, but why are you keeping `const char* ` in the changes to this file (`nucname.h`) and not in the one for the above file (`h5wrap.h`). The changelog says these may cause some problems.

bam241: this was not intended.
Let me push a fix...

thx for catching this

src/rxname.h
```diff
@@ -344,10 +345,11 @@ namespace rxname
         INFEstr += " --> ";
         INFEstr += rxnow;
       }
-      return (const char *) INFEstr.c_str();
+      const char* INFEstr_rtn = INFEstr.c_str();
+      return INFEstr_rtn;
     }
 
-  private:
+   private:
```
ngranda: Is this extra space on purpose?

ngranda (commented):
I just have one question about the extra blank space, but otherwise LGTM! Thanks @bam241

gonuke:
Thanks @bam241 - let's hope this makes a difference

bam241:
@gonuke I PRed this one vs develop :(

gonuke:
Oops!  Does that mean we need another PR to test this in conda?

bam241:
No I updated my PR on Pune-feedstock...
I fixed the warning, but not the unexpected failure of the test...
Not sure what to do with OSX-conda....

PR #1267, opened 2020-08-11
Fix changelog

opotowsky:
I noticed that some spacing issues was preventing some of the lists rendering as such (and so they were paragraphs with asterisks separating entries). I have done my best to fix this formatting problem. In most cases, it is because there needs to be a minimum of 3 spaces to create a new nested list. 

Also changed: 
- Single ` are now double `` uniformly
- Lines 64 and 185 have what I believe to be corrected typos -- please check that these make sense
- Tried to create uniform blank spacing

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,18 +15,18 @@ v0.7.0 RC2
 
 * Subvoxel R2S allows activation and photon source sampling by
   material within a mesh voxel
-  * SubMode parameter in source_sampling
-  * Add code for submode: SUBVOXEL
-  * Test functions to test SUBVOXEL codes in test_source_sampling.py
-  * Add subvoxel r2s workflow support in pyne/scripts/r2s.py
-  * Use new sampler for both voxel and sub-voxel R2S.
-  * Define cell_mats for subvoxel equals False
-  * Function Sampler::mesh_tag_data to be compatible for both DEFAULT and SUBVOXEL mode
+   * SubMode parameter in source_sampling
+   * Add code for submode: SUBVOXEL
+   * Test functions to test SUBVOXEL codes in test_source_sampling.py
+   * Add subvoxel r2s workflow support in pyne/scripts/r2s.py
+   * Use new sampler for both voxel and sub-voxel R2S.
+   * Define cell_mats for subvoxel equals False
+   * Function Sampler::mesh_tag_data to be compatible for both DEFAULT and SUBVOXEL mode
 
 * OpenMC support
-  * Add CI test for OpenMC as an optional dependency
-  * Add options to install OpenMC Python API
-  * Functions to read OpenMC state point file to get meshtally data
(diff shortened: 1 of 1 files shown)


opotowsky:
Let me know if you want me to include this PR in the changelog ;-)

CHANGELOG.rst
```diff
@@ -15,18 +15,18 @@ v0.7.0 RC2
 
 * Subvoxel R2S allows activation and photon source sampling by
   material within a mesh voxel
```
gonuke: I think that if the above is on a single line, the formatting might work better.

CHANGELOG.rst
```diff
@@ -15,18 +15,18 @@ v0.7.0 RC2
...
 * OpenMC support
-  * Add CI test for OpenMC as an optional dependency
-  * Add options to install OpenMC Python API
-  * Functions to read OpenMC state point file to get meshtally data
+   * Add CI test for OpenMC as an optional dependency
+   * Add options to install OpenMC Python API
+   * Functions to read OpenMC state point file to get meshtally data
 
 * File scripts/activation_responses.py to get activation responses data for
   ALARA output file for better visualization. Responses include:
```
gonuke: Also here with keeping these on a single line for formatting

CHANGELOG.rst
```diff
@@ -147,80 +145,80 @@ v0.7.0 RC2
...
+   * new check won't now be triggered after a merge only on PRs
+   * utils.py: updated the download timeout time to 120sec (from 30sec)
+   * updated CI to use CircleCI 2.1 workflows: now build separately from tests with state saved between runs
+   * test_fluka:
+      * added test to check the data tag name of the different tally part and
+        error.
+   * revert internal nuc_data_path to origin value after internal data test
+   * added DEFINE variable to allow material.cpp amalgamation without decay.cpp
+   * now skips endf test when website is not reachable to allow completeness of
+     the other tests.
```
gonuke: Keep the above on the same line

gonuke (changes requested):
Thanks for breaking the seal on this tedious formatting tasks....
I have a few more thoughts on formatting improvements.....

CHANGELOG.rst
opotowsky: on it!

opotowsky:
@gonuke from those changes, at least now all list entries with something nested under it are formatted the same (but they are the same no matter where they fall in the nesting, don't know if that's a bother or not)

gonuke:
I think this is good.  Will need a rebase on other recent merges that affect CHANGELOG

opotowsky:
rebased!

gonuke (approved):

gonuke:
Thanks @opotowsky - I'm not waiting for tests since I know that the only file that changed is the CHANGELOG and I don't want you rebasing forever :)

PR #1268, opened 2020-08-12
Flip QA paradigm

gonuke:
PyNE issues QA warnings excessively which may impact the reception by some audiences.    This PR aims to "flip the bit" on QA and only issue warnings when an environment variable has been set.

The first step is to move all QA warnings to a single imported function call.

Now also check an ENV variable to decide whether or not we issue QA warning

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -43,6 +43,7 @@ v0.7.0 RC2
 
 **Enhancements to Previous Capabilities**
 
+* Turn off QA warnings by default and enable with env variable 'PYNE_QA_WARN' (#1268)
 * Enhancements of Material and MaterialLibrary capabilities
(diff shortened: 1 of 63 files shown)


opotowsky:
This looks good to me. I don't have all the components that this would affect (e.g. dagmc), so should I wait on another reviewer to test on their machine?

gonuke:
Note: this will change the output that appears in tutorials and/or examples.  Probably good to merge it before too much documentation work

opotowsky:
> Note: this will change the output that appears in tutorials and/or examples. Probably good to merge it before too much documentation work

Yes agreed. Hopefully @bam241 can get to it soon :-)

bam241 (approved):
I scanned through this.

Looks good.

thanks you @gonuke for providing a way to get rid of those annoying msg :)

PR #1271, opened 2020-08-12
Python 2 to 3 Syntax for Tutorial and Example Jupyter Notebooks

### LINKED ISSUE #1036: Tutorial print function in python 2
Hey, 
I was going through the tutorials and I noticed that all the tutorials' print functions are written as:

print variable (python 2)
instead of: 

print(variable) (python 3)
I was wondering if the print functions should be updated to the python 3 way. I have been working through the tutorials and updating them so they are concurrent with python 3. Should I pull request these as an update(s)?

Thank you

opotowsky:
Addresses Issue #1036 

In looking into the use of jupyter notebooks (for both tutorials and examples), it is clear many were failing because of simple python 2 v 3 syntax. This PR fixes that, which is mostly print statements (and some import statements), but it does not address the other errors (some are expected import errors, some are ipython magic command related). 

The notebooks were updated with the syntax changes but not re-run on my machine. They were tested with the `execer.py` script.


opotowsky:
Since @ZoeRichter @ngranda and I are going to be working on fixing the tutorials, I propose we first merge this since it addresses an issue (this PR doesn't fix all the other problems).

opotowsky:
@bam can you understand why I'm getting these two errors? They are different and the py3 one should be completely unrelated to the changes I made

bam241 (approved):
LGTM!

PR #1273, opened 2020-08-12
Jupyter (formerly iPython) not needed in setup.py

### LINKED ISSUE #810: Ipython not actually needed for setup

opotowsky:
This PR resolves issue #810 by deleting the ipython check from setup.py. Since it is an optional dependency, it has been added as such to the appropriate documentation.

### DIFF
--- a/docs/devsguide/testing.rst
+++ b/docs/devsguide/testing.rst
@@ -32,12 +32,13 @@ passing.
 Example Testing
 ---------------
 The examples directory should also be kept up-to-date as much as possible.
-PyNE examples are either in Python files or Jupyter notebooks. This means that
-to test the examples requires a recent version of Jupyter.  Furthermore, the 
-examples themseleves may have many other optional dependencies.  Don't be alarmed
-if testing the examples fails do to a lack of having a dependency installed.
-For this reason, testing the examples is not as important as unit tests, but still
-should be done occassionally.
+PyNE examples are either in Python files or `Jupyter notebooks
+<https://jupyter.org/>`_. This means that to test the examples requires a
+recent version of Jupyter.  Furthermore, the examples themseleves may have many
+other optional dependencies.  Don't be alarmed if testing the examples fails do
+to a lack of having a dependency installed.  For this reason, testing the
+examples is not as important as unit tests, but still should be done
+occassionally.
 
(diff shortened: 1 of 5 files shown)


docs/devsguide/testing.rst
```diff
@@ -32,12 +32,13 @@ passing.
...
-PyNE examples are either in Python files or Jupyter notebooks. This means that
-to test the examples requires a recent version of Jupyter.  Furthermore, the 
-examples themseleves may have many other optional dependencies.  Don't be alarmed
-if testing the examples fails do to a lack of having a dependency installed.
-For this reason, testing the examples is not as important as unit tests, but still
-should be done occassionally.
+PyNE examples are either in Python files or `Jupyter notebooks
+<https://jupyter.org/>`_. This means that to test the examples requires a
+recent version of Jupyter.  Furthermore, the examples themseleves may have many
+other optional dependencies.  Don't be alarmed if testing the examples fails do
```
nsryan2:
```suggestion
other optional dependencies.  Don't be alarmed if testing the examples fails due
```

opotowsky: Wow I wonder how long that typo has lingered. Thanks for catching!

docs/install/linux_source.rst
```diff
@@ -20,11 +20,15 @@ PyNE has the following dependencies:
    #. `Python 2.7 <http://www.python.org/>`_
```
nsryan2:
```suggestion
   #. `Python <http://www.python.org/>`_
```

nsryan2:
It looks good to me, though I can't figure out why it isn't passing two of the tests. Are they even important for this PR, because it only has to do with the documentation and examples and not pyne itself?

opotowsky:
It's not passing because of changes to `setup.py`...I'm looking into it

opotowsky:
I think they are red x's b/c they were cancelled builds as I pushed new commits -- looking good so far this time!

gonuke (approved):
Thanks @opotowsky - I'll wait for testing and then merge

PR #1274, opened 2020-08-12
Deprecating py2 test

bam241:
this remove most of the python2 tests.
Only keeping python2 PyNE tests with all (pyMOAB and DAGMC) dependencies.


gonuke:
Thanks @bam241  for streamlining the tests as we move away from python 2

PR #1275, opened 2020-08-13
Typo in format string in pyne/fortranformat/_parser.py

### LINKED ISSUE #894: Question about format string in pyne/fortranformat/_parser.py
Near the end of `parse_tokens()`, if the edit descriptor isn't matched an exception is thrown with:

```
            raise InvalidFormat('Could not identify edit descriptor '
                                'in sequence $s' % str(token_set))

```
Is `$s` a legitimate format string for `%`? It looks like a typo but I've been wrong before.

I'd prefer to refactor this to 

```
            raise InvalidFormat('Could not identify edit descriptor '
                                'in sequence {0:s}'.format(str(token_set)))

```
but I want to make sure I understand the code's intent before making that change.

opotowsky:
The person with a keen eye who started issue #894 noticed a typo, and this PR fixes that.

### DIFF
--- a/pyne/fortranformat/_parser.py
+++ b/pyne/fortranformat/_parser.py
@@ -79,7 +79,7 @@ def _parse_tokens(tokens, reversion=False, version=None):
         elif ed_type == 'ED10':
             ed = _read_ed10(token_set)
         else:
-            raise InvalidFormat('Could not identify edit descriptor in sequence $s' % str(token_set))
+            raise InvalidFormat('Could not identify edit descriptor in sequence %s' % str(token_set))
         # If there is a repeat number cached, then apply
         if repeat is not None:
             ed.repeat = repeat
--- a/pyne/fortranformat/_parser.py
+++ b/pyne/fortranformat/_parser.py
@@ -79,7 +79,7 @@ def _parse_tokens(tokens, reversion=False, version=None):
         elif ed_type == 'ED10':
             ed = _read_ed10(token_set)
         else:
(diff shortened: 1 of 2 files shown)


pyne/fortranformat/_parser.py
```diff
@@ -79,7 +79,7 @@ def _parse_tokens(tokens, reversion=False, version=None):
         elif ed_type == 'ED10':
             ed = _read_ed10(token_set)
         else:
-            raise InvalidFormat('Could not identify edit descriptor in sequence $s' % str(token_set))
+            raise InvalidFormat('Could not identify edit descriptor in sequence %s' % str(token_set))
```
gonuke: I think the OP's suggestion of switching from the `%`-formatting to `string.format()` is probably a good idea here.

gonuke (changes requested):
Thanks for picking up this pinoli @opotowsky

pyne/fortranformat/_parser.py
opotowsky: I can do that throughout the file. Should I scrape other files for that or will that take me weeks?

opotowsky: > Thanks for picking up this pinoli @opotowsky

I drink piñon coffee every day, so I'm devoted to all low-hanging pinoli!

opotowsky: (No idea how I responded to that here)

gonuke (approved):

gonuke:
Thanks @opotowsky - I'm pretty sure we can merge this without the rebasing to remove tests.  If you pass more tests than necessary, it should be fine :)

Issue #1276, opened 2020-08-13
Tutorials and Examples Update


ngranda:
Created to help keep track of what is being updated  and not replicate efforts. 

Tutorials
- [x]  Introduction
- [x] Nuclide Naming
- [x]  Materials
- [x]  Nuclear Data
- [x]  Reaction Names
- [x]  Cross Sections
- [x]  Transmutation 
- [x]  Mesh Basics
- [x]  Diffusion
- [x]  Outroduction
- [ ]  Spatial Solver
- [x]  ENSDF Processing (This is not linked on the site so that needs to happen too)

Examples
- [x]  Ace Module
- [x]  Chainsolve Transmutation
- [x]  Discretized Teapot
- [x]  Half Life Plot
- [x]  Data Sources
- [x]  Enrichment: M* vs Flowrate
- [x]  FNG Model
- [x]  Materials
- [x]  Material Library
- [x]  Mesh Tagging
- [x]  Open ORIGEN Data - No issue running as is.
- [x]  ORIGEN 2.2: H1 Cross Section - Cannot reproduce without user-supplied input file.
- [x]  Reaction Names
- [x]  ENDF Files

opotowsky:
As per our discussion, the `AttributeError: 'get_ipython' object has no attribute 'run_line_magic'` problem when running all notebooks using `execer.py` will not factor into these fixes unless we find a solution. This magic command error (in our case, from using `%matplotlib inline`) is a [known issue](https://github.com/jupyter/nbconvert/issues/503) when using `jupyter nbconvert`.

opotowsky:
@gonuke should we submit each notebook fix as a separate PR or is it OK to clump them? The former is probably better practice but wanted to check you don't prefer the latter.

gonuke:
I think it's best to be flexible.  I don't think we need separate PRs for each one, but I also wouldn't want to hold up review to include them all.  If there's a natural break point to PR a group of them together and move on, go for it.

kkiesling:
There's a one-line fix for the intro notebook that can be found in PR #1241 if anyone gets the chance to grab that one liner (could close #1241 then in favor of a different PR).

opotowsky:
> There's a one-line fix for the intro notebook that can be found in PR #1241 if anyone gets the chance to grab that one liner (could close #1241 then in favor of a different PR).

Left a comment in the PR

ZoeRichter:
[...]

PR #1241, opened 2020-07-21
update tutorial to remove pytaps references

kkiesling:
This fixes one line in the intro tutorial to properly call the pyMOAB function rather than the pytaps function.

PR #1282, opened 2020-08-13
Low-hanging pinoli tutorial updates

### LINKED ISSUE #1240: pytaps reference in the tutorials
The tutorials still use the old `pyne.Mesh` implementation based on `pytaps`

for example the usage of `m.mesh.getVtxCoords(ve)` in the `00-intro` tutorial is outdated.

One ned to update the tutorials to the last `pyne.Mesh` implementation based on `pymoab`

opotowsky:
This is part of the larger effort in Issue #1276 

It reruns the cells in most of the tutorial notebooks included here, to check they still work and to get rid of QA wqarnings.

The only exception is 00-intro, which takes the change (update to pymoab call) originally introduced in PR #1241 (Which addressed issue #1240).


gonuke:
Can you please run all these NBs again after the mass data update, please?

kkiesling (approved):

PR #1293, opened 2020-08-14
Fix ace example

ngranda:
This PR fixes the ace reader example. Main change is in how we get the data. urllib's urlretrieve not longer valid, now using requests


opotowsky (approved):
LGTM, ping me when CI passes if I don't come back to merge soon enough

gonuke:
Thanks @ngranda

PR #1296, opened 2020-08-14
Remove example Origen 2.2:H1 Cross Section

ypark234:
This PR removes an example [Origen 2.2:H1 Cross Section] as it requires `BASE_TAPE9.INP` to reproduce the example, but PyNE repo does not have the file in it.
#1276


gonuke:
Thanks @ypark234

opotowsky:
@ypark234 @gonuke Right now this file is in the current RC branch. How did it get re-added?

PR #1298, opened 2020-08-14
Remove deprecated U mode in file open

gonuke:
This mode is now default and the input is deprecated.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -250,6 +250,7 @@ v0.7.0 RC2
    * cleanup throws return from ``return (const char *)`` to simple ``return``
      (it was suggested that those complicated return might cause seg fault, on some system -- OsX+conda)
    * Fixes string formatting typo in ``pyne/fortranformat/_parser.py``. Entire file is now ``str.format()`` (#1275)
+   * Remove deprecated file open mode 'U' (#1298)
 
 **Fixed:**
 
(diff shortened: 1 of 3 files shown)


gonuke:
This PR breaks python2 support.  In Python2, you have to explicitly choose to open in universal newline mode, while in python 3 you do not.  We have three choices:
1. merge this and say goodbye to Python 2
2. leave this unmerged and live with warnings about deprecation (we can 
3. add some (complicated?) logic to handle both cases...

gonuke:
I'm taking this out of the 0.7 release, we can live with warnings there

gonuke:
Now that we're officially done with Python 2 - we can merge this....

gonuke:
but this will officially break python 2 for ever... it is the point of no return.

kkiesling:
> but this will officially break python 2 for ever... it is the point of no return.

we made the community decision back in 2017 and python 2 is officially deprecated- time to go past the point of no return.

gonuke:
> > but this will officially break python 2 for ever... it is the point of no return.
> 
> we made the community decision back in 2017 and python 2 is officially deprecated- time to go past the point of no return.

Yup - then we should stop testing on python 2 in CI (post-release)

kkiesling:
I am fine with merge failed python 2 CI since this is expected but we should immediately remove python 2 testing in CI

PR #1304, opened 2020-08-14
Updates to some of the last tutorials

opotowsky:
A part of issue #1276 , this just reruns import statements to remove QA warnings, adds yt warnings, updates some syntax, and adds another urllib --> requests update. None of the cells in these notebooks are rerun because of either yt or there are errors that need more digging into outside of the release timeframe. 

00-intro is just yt warn
05-cross-sections is QA removal, urllib update
08-diffusion is yt warn, syntax update
11-ensdf-processing is QA warn removal


ngranda (approved):
This looks good, waiting on CI tests

Issue #1302, opened 2020-08-14
Tutorial 11: ENSDF processing's alphad not found 


opotowsky:
@ZoeRichter and I got the same error when trying to run tutorial 11, ENSDF Processing.

Code:
```
from pyne import ensdf_processing
input_dict = {}
input_dict['input_file'] = 'ensdf_processing/alphad/ref_a228.ens'
input_dict['report_file'] = '/alphad.rpt'
input_dict['rewrite_input_with_hinderance_factor'] = 1
input_dict['output_file'] = '/alphad.out'
output_dict = ensdf_processing.alphad(input_dict)
```
And the error:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/opotowsky/.local/lib/python3.6/site-packages/pyne/alphad': '/home/opotowsky/.local/lib/python3.6/site-packages/pyne/alphad'
```

gonuke:
PR #1307 should fix this.  The tests still fail - not sure why - but maybe there will be more info available through this tutorial?

opotowsky:
It doesn't fix it for me (did a clean reinstall). Can anyone else reproduce the error?

opotowsky:
> It doesn't fix it for me (did a clean reinstall). Can anyone else reproduce the error?

DUH! It's not merged yet.

PR #1307, opened 2020-08-16
Remove extra flag that prevented linking FORTRAN executables.

gonuke:
Final linking ENSDF fortran-based helper apps was prevented by a spurious `-c` in the linking line.

That has been removed...


gonuke:
The test still basically fail, but magically turn into `SkipTest`

opotowsky:
`alphad` is still missing for me after a clean reinstall

opotowsky (approved):
This looks good, but I want to note that it doesn't close issue #1302, for me at least.

gonuke:
It is not expected to resolve #1302 but does move us forward.  The processing tools are now at least built/installed, but they still aren't launched properly by the python methods that try to do so.

katyhuff:
Seems right, tests pass.

katyhuff (approved):

PR #1308, opened 2020-08-17
Ahot option

### LINKED ISSUE #1239: Add build option for AHOT
The AHOT module in PyNE adds dependencies (FORTRAN/BLAS/LAPACK) that can be non-trivial on many environments.  It would be useful to have a build option that turns off building this module, possibly by default.

gonuke:
add an option in CMake and setup.py to turn on/off spatial solvers.  Default is ON in both places.

Closes #1239


bam241 (approved):
LGTM!
thx @gonuke

PR #1310, opened 2020-08-17
Fix integer division and z typo in tutorial/01-nuclide-naming.ipynb

alexmalins:
This is a fix for the nuclide naming tutorial notebook. It address an issue created by the switch to [true division](https://www.python.org/dev/peps/pep-0238/) for the `/` operator between Python 2 and 3. Using `//` instead, the fixed version works both on Python 2 and 3.

The PR also fixes a typo in the example it concerned, as the atomic number was Pu (94) not Am (95).


gonuke:
Thanks @alexmalins

PR #1314, opened 2020-08-17
Bug fix class Xsdir

kkiesling:
Updated PR that has been rebased and can replace #1205


pyne/mcnp.py
```diff
@@ -793,7 +793,7 @@ def read(self):
             # Handle continuation lines
             while words[-1] == '+':
                 extraWords = self.f.readline().split()
-                words = words + extraWords
+                words = words[:-1] + extraWords  # Correct bug in the XSDIR parsing
```
gonuke:
```suggestion
                words = words[:-1] + extraWords
```
This was originally suggested by @pshriwise - but I'll add it here to expidite.

gonuke:
Thanks @kkiesling for pulling together this contribution from @Radiation-Transport and @pshriwise - Looks good to me.

PR #1322, opened 2020-08-20
[WIP] updating docs for VM

kkiesling:
This updates the docs for the VM for 0.7.0 release.

This is not quite ready yet- finishing the VM creation and then need to update name and links.

### DIFF
--- a/docs/install/vb.rst
+++ b/docs/install/vb.rst
@@ -4,7 +4,7 @@
 VirtualBox
 =================================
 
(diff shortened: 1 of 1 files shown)


gonuke:
Maybe this should just be changed to remove reference to the VM (at least for now?)

kkiesling:
do you suggest removing this page entirely or just unlinking it from the install page?

kkiesling:
addressed by #1348

PR #1324, opened 2020-08-20
Update conda install instruction

AI-Pranto:
Currently, `pyne` has several build options on `conda-forge` channel. 
```
conda install -c conda-forge pkg_name=version=build_string
```
will help users to install `pyne` with the specific build option.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -204,6 +204,7 @@ IMPORTANT: Python 2 is no longer supported as of this release.  Some packages ma
    * Fix broken links (#1311)
    * Update install instructions (#1238 & #1315)
    * Update New Developers Guide
+   * Update conda install instruction
 
 * Improvements in building and testing
    * require contributor to change CHANGELOG
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,10 @@ pyne Change Log
 Next Version
 ====================
 
+**Maintenance**
+
+* Documentation Changes
+   * Update conda install instruction. (#1324)
 
 v0.7.0
 ====================
@@ -204,7 +208,6 @@ IMPORTANT: Python 2 is no longer supported as of this release.  Some packages ma
    * Fix broken links (#1311)
    * Update install instructions (#1238 & #1315)
    * Update New Developers Guide
-   * Update conda install instruction
 
 * Improvements in building and testing
    * require contributor to change CHANGELOG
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -204,6 +204,7 @@ IMPORTANT: Python 2 is no longer supported as of this release.  Some packages ma
    * Fix broken links (#1311)
    * Update install instructions (#1238 & #1315)
    * Update New Developers Guide
+   * Update conda install instruction
```
gonuke: We have made the 0.7.0 release, so this changelog entry should now come near the top of the file under the "Next Version" heading.

Also, please include a reference to this pull request number in parentheses.

docs/install/conda.rst
```diff
@@ -14,9 +14,24 @@ option when installing conda.
 --------------------------
 Binary Package (For Users)
 --------------------------
+
+To list all of the versions of PyNE available on `conda-forge
+<https://conda-forge.github.io/>`_ channel, in your terminal window run::
+
+    conda search -c conda-forge pyne
+
 Binary distributions of the latest release for mac and linux (64-bit)
```
gonuke: We will not have mac versions available immediately.  This should probably be corrected.

docs/install/conda.rst
```diff
@@ -14,9 +14,24 @@ option when installing conda.
...
+
 Binary distributions of the latest release for mac and linux (64-bit) 
 using the conda package manager can be installed by running the command::
 
     conda install -c conda-forge pyne
 
+If you want to install PyNE with the correct package specification, try
+``pkg_name=version=build_string``.
+
+For example, if you want to install ``pyne version=0.7.0`` with build option ``moab_openmc_py36he21c9c4_1``, you would enter::
```
gonuke: It may be more interesting to users to use wild card on specifying the build options.  That is, a user may know that they want a version with moab and openmc, but don't care about the hash or python version.  Therefore, they might install as:
` conda install -c conda-forge pyne=0.7.0=moab_openmc*`

gonuke (changes requested):
Thanks for this contribution @AI-Pranto.  We are in the process of finalizing the conda-forge builds for the recent release and will need to update the documentation.  This is a good start, but we are not entirely sure how the documentation will change.

CHANGELOG.rst
AI-Pranto: What will be the name of the next version @gonuke? `0.7.1`?

gonuke: That's not clear, but we can leave the header as "Next Version" until that is decided.  I expect it will be 0.7.1, but until we have a release candidate, we'll continue to refer to is as "Next Version"

gonuke (approved):

gonuke:
Thanks for this contribution @AI-Pranto

PR #1326, opened 2020-08-23
Update version numbers for release 0.7.1

alexmalins:
It looks like the version numbers didn't get updated from 0.5.11 to 0.7.0 with the latest release. This means `pyne.__version__` reports the old number. This PR changes the version numbers in the code to 0.7.0. Not sure if this is how you want to catch this or if you just want to update to an appropriate number at the time of the next release.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -24,7 +24,7 @@ Note, please do not include screenshots or snippets of code that is export contr
 
 **Please complete the following information regarding your system:**<br/>
  - OS: [e.g. ubuntu Linux]
- - PyNE Version: [e.g. 0.7.0]
+ - PyNE Version: [e.g. 0.7.1]
  - Versions of dependencies installed with: [e.g. HDF5 1.10]
 
 **Additional Context**<br/>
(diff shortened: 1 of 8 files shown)


gonuke:
thanks @alexmalins - I thought I had looked for this, but I guess I failed!

Now I'm trying to figure out how to wrap my head around which version this should be...  kind of like time travel

gonuke:
Maybe we update this to `0.7.1` and do a microrelease immediately

alexmalins:
Yes, possible options are i) delete the old release, merge the correct numbers and re-release as `0.7.0`, ii) go with a `0.7.1` microrelease now, or iii) just catch this at the time of the next planned release. I'm happy to do whatever the project decides is best and will update/change this PR as and when is required.

gonuke:
We'll be doing a microrelease soon because of changes that were needed for downstream purposes, so let's just go with that. Thanks!

gonuke:
In case my previous message wasn't clear, let's update the version string to 0.7.1 in advance of that microrelease.

alexmalins:
@gonuke Ok, it should be good to go now!

gonuke:
Thanks for catching all these @alexmalins - will hopefully have the microrelease soon

PR #1327, opened 2020-08-28
move join_to_string declaration in the header

bam241:
Clang didn't like to the way the declaration of `join_to_string` was...

Moving everything int header fixed it....

this is required to amalgamated PyNE and compile it with `clang`. see svalinn/DAGMC#700

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -131,7 +131,24 @@ namespace pyne {
 
   // join the vector element into a string, each values will be delimited ny the delimiter
   template<typename T>
-  std::string join_to_string(std::vector<T> vect, std::string delimiter = " ");
+  std::string join_to_string(std::vector<T> vect, std::string delimiter = " "){
+    std::stringstream out;
+    out << std::setiosflags(std::ios::fixed) << std::setprecision(6);
+    
+    // ensure there is at least 1 element in the vector
+    if (vect.size() == 0)
+      return out.str();
+    // no delimiter needed before the first element
+    out << vect[0];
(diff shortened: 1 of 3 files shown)


src/utils.h
```diff
@@ -131,7 +131,24 @@ namespace pyne {
...
+    // no delimiter needed before the first element
+    out << vect[0];
+    for( int i = 1; i < vect.size(); i++)
+      out << delimiter << vect[i];
+    return out.str();
+  };
+  template std::string join_to_string(std::vector<int> vect,
+                                      std::string delimiter);
+  template std::string join_to_string(std::vector<double> vect,
+                                  std::string delimiter);
```
pshriwise:
```suggestion
  template std::string join_to_string(std::vector<double> vect,
                                      std::string delimiter);
```

pshriwise (changes requested):
Very small style change, but then I think this good to go.

gonuke (approved):
Seems OK to me, but I wish I understood why....

gonuke:
Thanks @bam241

bam241:
@gonuke

I did not read it yet, but here is what @pshriwise told me on slack:
```
This has a pretty good explanation of why it's best to put 
templates in the header file. 
http://www.cs.technion.ac.il/users/yechiel/c++-faq/templates-defn-vs-decl.html


I think gcc is smart enough to add that template definition to 
the header file's translation unit (as if it were an inline'd 
function) but clang apparently isn't

```

PR #1330, opened 2020-09-01
Allowing MaterialLibrary to behave like an unordered_map

bam241:
this add most the the `unodered_map` API to the MaterialLibrary front-end.

this API link directly to the nested unordered_map (the material_library) to simplify calls.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,9 @@ pyne Change Log
 Next Version
 ====================
 
+**New Capabilities**
+   * Adding unordered_map like API to MaterialLibrary pointing directly to the nested unordered_map
+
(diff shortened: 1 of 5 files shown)


bam241:
FYI svalinn/dagmc#700 is using this new API successfully

gonuke:
Are there ways to expose the new methods to Python and test that way???

bam241:
maybe partially, I'll try something

bam241:
@gonuke this is not a perfect test, but now Python uses the frontend iterator from MaterialLibrary instead of getting the `unordered_map` object first

gonuke:
thanks @bam241 - I think this is best we can do for now with testing.

PR #1334, opened 2020-09-10
Now adds number material when added to a MaterialLibrary

bam241:
this change the behavior of  the `add_material()` of the `MaterialLibrary`, if no number were associated to the material :
- before: a number was found and used to name the material if it had no name (the number was not attributed
to the material)
- now: a number is found and attributed to the material if it has none...

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,8 @@ pyne Change Log
 Next Version
 ====================
 
-
+**Change**
+   * Now attribute a number to material when adding them into a material library (#1334)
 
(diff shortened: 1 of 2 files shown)


bam241:
this should solve all the material number problem experienced with uwuw preproc.
we don't need to update the `nuc_data.h5` as numbers would be attributed on library loading.

We could update the `nuc_data.h5` to ensure compatibility of uwuw preprocessing with `0.7.1`

gonuke (approved):
Thanks @bam241 - LGTM

PR #1336, opened 2020-09-16
Mcnp6 ptrac

MigMagan:
Compatibility fix: small changes to `mcnp.py` to make it compatible with MCNP6-generated ptrac files. Also added an MCNP6 ptrac file in tests,  generated running MCNP6 with the same `mcnp_ptrac_inp.txt` .

PR #1337, opened 2020-09-16
update the address of the materials group for the hash checking

bam241:
New API of the MaterialLibrary write (in HDF5) the `materials` group in the `/maerial_library` dataset.

this reflect that change for hash checking.

### DIFF
--- a/pyne/dbgen/hashtools.py
+++ b/pyne/dbgen/hashtools.py
@@ -12,7 +12,7 @@ from .. import data
 QA_warn(__name__)
 
 # list of nodes from distinct data sets
-nodelist = ['/atomic_mass', '/material_library',
+nodelist = ['/atomic_mass', '/material_library/materials',
             '/neutron/eaf_xs', '/neutron/scattering_lengths',
             '/neutron/simple_xs', '/decay', '/dose_factors']
 
--- a/pyne/dbgen/materials_library.py
+++ b/pyne/dbgen/materials_library.py
@@ -128,7 +128,7 @@ def parse_materials(mats, lines):
 # Writes to file
 def make_materials_compendium(nuc_data, matslib):
     """Adds materials compendium to nuc_data.h5."""
-    matslib.write_hdf5(nuc_data, datapath="/materials")
+    matslib.write_hdf5(nuc_data)
 
 
 def make_matslib(fname):
--- a/pyne/dbgen/materials_library.py
+++ b/pyne/dbgen/materials_library.py
@@ -128,8 +128,7 @@ def parse_materials(mats, lines):
 # Writes to file
 def make_materials_compendium(nuc_data, matslib):
     """Adds materials compendium to nuc_data.h5."""
-    matslib.write_hdf5(nuc_data, datapath="/material_library/materials",
-                       nucpath="/material_library/nucid")
+    matslib.write_hdf5(nuc_data, datapath="/materials")
 
 
 def make_matslib(fname):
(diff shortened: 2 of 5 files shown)


bam241:
This should address the bug seen in #1336

bam241:
@gonuke @pshriwise @kkiesling sorry to be pressing but...

With the update of the `nuc_data.h5`, the CI-test are not working. One need to review and merge this to have CI running again, allowing us to review/merge other PR.

pshriwise:
Sounds good. I'll take a look at it today.

pyne/dbgen/hashtools.py
```diff
@@ -12,7 +12,7 @@
 QA_warn(__name__)
 
 # list of nodes from distinct data sets
-nodelist = ['/atomic_mass', '/material_library',
+nodelist = ['/atomic_mass', '/material_library/materials',
```
pshriwise, line 15: Confused here. In other locations the path is updated from `/material_library` to `/materials`, but here `materials` is appended to the original path?

pyne/dbgen/materials_library.py
```diff
@@ -128,8 +128,7 @@ def parse_materials(mats, lines):
 # Writes to file
 def make_materials_compendium(nuc_data, matslib):
     """Adds materials compendium to nuc_data.h5."""
-    matslib.write_hdf5(nuc_data, datapath="/material_library/materials",
-                       nucpath="/material_library/nucid")
+    matslib.write_hdf5(nuc_data, datapath="/materials")
```
pshriwise: Given that `/materials` is the default path in the C++ API for this object, we shouldn't need to provide it here any longer. True here and in a few other locations.

pshriwise: I'm noting now that this is only true for `write_hdf5`. A separate issue, but it seems like it would make sense for `from_hdf5` to have the same default value for the datapath.

pshriwise (commented):
Just a few small questions.

pyne/dbgen/hashtools.py
bam241: because for a material library the name of the groups is nested inside the `material_library` dataset

pyne/dbgen/materials_library.py
bam241: this is the name of the material library that will contain the written library.
it will be nested inside the `material_library` dataset

pyne/dbgen/hashtools.py
pshriwise: Ah, I see -- the base is always `/material_library`

https://github.com/pyne/pyne/blob/c61feaca4748299b08aeec7f069c221684da8e45/src/material_library.cpp#L245

Sorry, it's been a while since I've looked at this.

bam241: it is kinda new... been working on it for a year, but only been merged a month or 2 ago...

pyne/dbgen/materials_library.py
pshriwise: Agreed, based on your comment above, but we're still explicitly providing the default value of the `datapath` argument here. 

https://github.com/pyne/pyne/blob/c61feaca4748299b08aeec7f069c221684da8e45/pyne/material_library.pyx#L122

It seems like this could be shortened this to:

```python
matslib.write_hdf5(nuc_data)
```

pshriwise (approved):
Looks good to me now! Thanks @bam241. Will merge today on :heavy_check_mark:

PR #1341, opened 2020-09-25
Tag energy boundaries as pymoab sparse

zxkjack123:
##  Description
Add energy boundaries data (`e_bins`) as a PyMOAB SPARSE tag. This PR is the replacement of PR #1157 .

## Motivation and Context
This PR aims to tag energy boundaries data (`e_bins`) as a PyMOAB NativeMeshTag for a mesh in the r2s workflow. The `e_bins` tag will replace the file "e_bounds" in the photon transport step of PyNE R2S. 

## Behaviour
Before this PR, a "e_bounds" text file is required in photon transport of r2s.
After this PR, the "e_bounds" is no longer required, the `e_bounds` information can be passed in by `source.h5m` file. The `e_bounds` are stored in the `source.h5m` as a sparse tag.
For compability, it is still supported to use text file with the old `source.h5m` generated before this PR.

## Todo
- [x] : read `e_bounds` from h5m file during the source sampling
- [x] : add new sampler without e_bounds argument
- [x] : compatible of two ways to pass in `e_bounds`:
    - [x] : use 'e_bounds' in `source.h5m` if it exists
    - [x] : use 'e_bounds' provided by txt file if 'source.h5m' does not contain it 
- [x] : update docs


zxkjack123:
@pshriwise I am trying to tag a sparse tag on mesh with `NativeMeshTag` with `MB_TAG_SPARSE`. The data is tagged, but it works like `MB_TAG_DNESE`. Do I use the wrong parameters here?

pshriwise:
This means it behaves like a dense tag in that each element has a tag value after the tag is created?

pshriwise:
When you say it works like `MB_TAG_DENSE` this means that all of the elements have a value when you're only trying to tag a few?

zxkjack123:
> When you say it works like `MB_TAG_DENSE` this means that all of the elements have a value when you're only trying to tag a few?

Yes. I want the tag data appears only once, but it appears `num_ves` times, just like `MB_TAG_DENSE` tag. For example, the data I want to tag in the test function is `[0.0, 0.1, 20]`, the targe mesh has 4 elements. However, the output tag data is: 
```
[[0.0, 0.1, 20],
[0.0, 0.1, 20],
[0.0, 0.1, 20],
[0.0, 0.1, 20]]
```

pshriwise:
Sounds good. I'll take a look shortly.

pyne/mesh.py
```diff
@@ -974,6 +996,22 @@ def tag(self, name, value=None, tagtype=None, doc=None, size=None,
         dtype : numpy dtype, optional
             The data type of the tag. This only applies to NativeMeshTags. See PyMOAB
             for more details.
+        storage_type: str, optional
```
pshriwise, line 997: I think the storage types listed below should be `'sparse'` and `'dense'` based on the type description here and the code below.

pyne/mesh.py
```diff
@@ -1008,9 +1046,19 @@ def tag(self, name, value=None, tagtype=None, doc=None, size=None,
                                  'or dtype'.format(name))
             else:
                 tagtype = MetadataTag
+
         if tagtype is NativeMeshTag or tagtype.lower() == 'nat_mesh':
+            # storage_type works for only NativeMeshTag
+            if storage_type is None or storage_type.lower() == 'dense':
+                storage_type = types.MB_TAG_DENSE
+            elif storage_type == types.MB_TAG_SPARSE or \
```
pshriwise: I think we only want to check if the `storage_type` is equal to `'sparse'` here.

pyne/mesh.py
```diff
@@ -1452,6 +1501,22 @@ def tag_cell_fracs(self, cell_fracs):
                  size=1, dtype=float)
 
 
+    def tag_e_bins(self, e_bins):
```
pshriwise: This is a nice convenience function, but it seems a little out of place, a little too specific, for this generic `Mesh` class. It might be better off in the location of the code being used?

tests/test_mesh.py
```diff
@@ -1002,6 +1002,13 @@ def test_tag_cell_fracs():
         assert_equal(m.cell_largest_frac[i],
                      exp_cell_largest_frac[i])
 
+def test_tag_e_bins():
+    m = gen_mesh()
+    e_bins = np.array([0.0, 0.1, 20])
+
+    m.tag_e_bins(e_bins)
```
pshriwise: This test looks good to me. Based on my comment above we may want to create the tag manually here instead.

pshriwise (changes requested):
Thanks for the addition @zxkjack123! You did a nice job keeping the internals of the Mesh class' reliance on PyMOAB mostly separate from the external API. I just have a few comments on how to complete that and on the `tag_e_ bins` convenience method.

tests/test_mesh.py
zxkjack123: @pshriwise I have moved the `tag_e_bins` to `r2s` part as it is used in `r2s step2`. The test function fails for me (with pymoab version '5.1.0'), here is the error message:
```
FAIL: test_r2s.test_tag_e_bins
  File "/home/gw/.local/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/gw/opt/pyne/tests/test_r2s.py", line 378, in test_tag_e_bins
    assert_array_equal(m.e_bins[:], e_bins)
  File "/home/gw/.local/lib/python3.6/site-packages/numpy/testing/_private/utils.py", line 931, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "/home/gw/.local/lib/python3.6/site-packages/numpy/testing/_private/utils.py", line 759, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not equal

(shapes (4, 3), (3,) mismatch)
 x: array([[ 0. ,  0.1, 20. ],
       [ 0. ,  0.1, 20. ],
       [ 0. ,  0.1, 20. ],
       [ 0. ,  0.1, 20. ]])
 y: array([ 0. ,  0.1, 20. ])
```
How should we do it if we want to create the tag `manually`?

pshriwise: Sorry for taking to long to get back to you @zxkjack123. It looks like there are a couple of things preventing the behavior you want. First, there's this behavior at the bottom of `Mesh.tag`:

https://github.com/pyne/pyne/blob/dc5fccf9b1624fc6e70e976f6c433e5a2e7f074b/pyne/mesh.py#L1020-L1021

If a tag is added to the mesh with the `value` keyword set. The method will set that value for all elements on the mesh. This is happening in the `t[:] = value` line. So even though it is being stored in a sparse manner in PyMOAB, all elements are getting data and it behaves like a dense tag.

Second, there's a similar line in the `NativeMeshTag` constructor:

https://github.com/pyne/pyne/blob/dc5fccf9b1624fc6e70e976f6c433e5a2e7f074b/pyne/mesh.py#L398-L399

which will set a default value if provided. Changing the default value to `None` should fix this problem, though if a tag is dense it's considered best practice to provide a default value. I'll leave it up to you as to how you'd like to address that.

With that fix to the second problem in place, the best way I can see to set a single tag value for the mesh in `tag_e_bins` would be

```python
def tag(m, e_bins, tag_name='e_bins'):
    m.tag(name=tag_name, doc='energy boundaries of the photon source',
              tagtype=NativeMeshTag, size=len(e_bins), dtype=float,          
              storage_type='sparse')                                                                                                                                                                                                                    
    m.get_tag(tag_name)[m] = e_bxins
    return m
```

This syntax tags the energy bins on the mesh instance.

The `test_tag_e_bins` function will need to be updated as well to only check for the energy bins on the mesh instance:

```python
assert_array_equal(m.e_bins[m], e_bins)
```

I could see that becoming a one-liner function in `r2s.py` for convenience:

```python
def get_e_bins(m):
    return m.e_bins[m]
```

pshriwise:
Not sure what’s going on with the `changelog_update` failure. Looks like it’s getting to this line

https://github.com/pyne/pyne/blob/dc5fccf9b1624fc6e70e976f6c433e5a2e7f074b/.circleci/changelog_test.sh#L20

and failing due the value of `default_branch` being `null`. Maybe I’m misinterpreting that though.

Any thoughts @gonuke, @bam241?

zxkjack123:
Thanks for the help of @pshriwise , this PR is ready to review.

pyne/mesh.py
```diff
@@ -554,6 +574,8 @@ def expand(self):
 
         if self.size < 2:
             raise TypeError("Cannot expand a tag that is already a scalar.")
+        if self.storage_type == types.MB_TAG_SPARSE:
+            raise TypeError("Expand of a sparse tag is not implemented.")
```
pshriwise:
```suggestion
            raise TypeError("Expansion of sparse tags is not implemented.")
```

pyne/mesh.py
```diff
@@ -1008,9 +1046,20 @@ def tag(self, name, value=None, tagtype=None, doc=None, size=None,
                                  'or dtype'.format(name))
             else:
                 tagtype = MetadataTag
+
         if tagtype is NativeMeshTag or tagtype.lower() == 'nat_mesh':
+            # storage_type works for only NativeMeshTag
```
pshriwise: Along the lines of this comment. I'm wondering if we would raise a warning for a non-default value of `storage_type` if the specified tag type isn't `NativeMeshTag`?

src/source_sampling.cpp
```diff
@@ -56,7 +57,8 @@ std::vector<double> pyne::read_e_bounds(std::string e_bounds_file){
       e_bounds.push_back(value);
   }
   else {
-    throw std::runtime_error("File " + e_bounds_file + " not found or no read permission");
+    //throw std::runtime_error("File " + e_bounds_file + " not found or no read permission");
```
pshriwise: Let's remove this commented line.

src/source_sampling.cpp
```diff
@@ -125,6 +127,65 @@ pyne::Sampler::Sampler(std::string filename,
...
+    if (e_bounds.size() > 0) {
+      // there is also an user defined e_bounds
+      std::cout<<"warning: e_bounds in 'source.h5m' will be used."<<std::endl;
+    }
+  }
+  setup();
+}
+
+pyne::Sampler::Sampler(std::string filename,
+                 std::map<std::string, std::string> tag_names,
```
pshriwise: Indentation updates needed here.

src/source_sampling.cpp
```diff
@@ -125,6 +127,65 @@ pyne::Sampler::Sampler(std::string filename,
...
+                 int mode)
+  : filename(filename),
```
pshriwise: These data member initializers can probably do on the same line.

src/source_sampling.h
```diff
@@ -180,6 +180,14 @@ namespace pyne {
             std::map<std::string, std::string> tag_names,
             std::vector<double> e_bounds,
             int mode);
+    /// Constuctor for simplified overall sampler
```
pshriwise, line 184:
```suggestion

    /// Constuctor for simplified overall sampler
```

src/source_sampling.cpp
```diff
@@ -224,6 +285,25 @@ void pyne::Sampler::setup() {
...
+  // get e_bounds tag data
+  moab::ErrorCode rval;
+  moab::Tag e_bounds_tag;
+  rval = mesh->tag_get_handle(e_bounds_tag_name.c_str(),
+                              e_bounds_tag);
+  if (rval == moab::MB_SUCCESS) {
+    // use e_bounds in 'source.h5m'
+    e_bounds.resize(num_e_groups + 1);
+    moab::EntityHandle root_set = mesh->get_root_set();
+    rval = mesh->tag_get_data(e_bounds_tag, &root_set, 1, &e_bounds[0]);
```
pshriwise:
```suggestion
    rval = mesh->tag_get_data(e_bounds_tag, &root_set, 1, e_bounds.data());
```

pyne/source_sampling.pyx
```diff
@@ -382,6 +383,38 @@ cdef class Sampler:
...
+    def _sampler_sampler_3(self, filename, tag_names, mode):
+        """Sampler(self, filename, tag_names, mode)
+        
+        Constuctor for overall Sampler
+        
+        Parameters
+        ----------
+        filename : std::string
+        
+        tag_names : std::map<std::string, std::string>
```
pshriwise, line 397: Missing parameter description for the `mode` argument here.

pyne/mesh.py
```diff
@@ -974,6 +996,22 @@ def tag(self, name, value=None, tagtype=None, doc=None, size=None,
...
+                entity handle
+            dense - MB_TAG_DENSE tag, values are stored in arrays which match
+                arrays of contiguous entity handles. Dense tags are more
+                efficient in both storage and memory if large numbers of
+                entities are assigned the same tag. Storage for a given dense
+                tag is not allocated until a tag value is set on an entity;
+                memory for a given dense tag is allocated for all entities in a
+                given sequence at the same time.  Sparse: Sparse tags are stored
+                as a list of (entity handle, tag value) tuples, one list per
+                sparse tag, sorted by entity handle.
```
pshriwise:
```suggestion
            dense - MB_TAG_DENSE tag, values are stored in arrays which match
                arrays of contiguous entity handles. Dense tags are more
                efficient in both storage and memory if large numbers of
                entities are assigned the same tag. Storage for a given dense
                tag is not allocated until a tag value is set on an entity, at which point
                memory allocation for the dense tag occurs for all entities.
```

pshriwise: The sparse description was present twice here. I also made some updates to the description of dense tags.

pyne/mesh.py
```diff
@@ -368,6 +368,22 @@ def __init__(self, size=1, dtype='f8', default=0.0, mesh=None, name=None,
...
+                entity handle
+            MB_TYPE_DENSE - Dense tag values are stored in arrays which match
+                arrays of contiguous entity handles. Dense tags are more
+                efficient in both storage and memory if large numbers of
+                entities are assigned the same tag. Storage for a given dense
+                tag is not allocated until a tag value is set on an entity;
+                memory for a given dense tag is allocated for all entities in a
+                given sequence at the same time.  Sparse: Sparse tags are stored
+                as a list of (entity handle, tag value) tuples, one list per
+                sparse tag, sorted by entity handle.
```
pshriwise: The description updates below apply here as well.

docs/usersguide/source_sampling.rst
```diff
@@ -154,9 +155,7 @@ a single argument: an integer representing the problem mode (0: DEFAULT_USER, 1:
 DEFAULT_UNIFORM, 2: DEFAULT_USER, 3: SUBVOXEL_ANALOG, 4: SUBVOXEL_UNIFORM,
 5: SUBVOXEL_USER). This function assumes the mesh file is "source.h5m" and
 that the tag names are "source_density", "biased_source_density",
```
pshriwise, line 157: The `"source_density"` parameter was removed in the source code below. Does it need to be removed here as well?

pshriwise (changes requested):
The removal of the `e_bounds` file is a significant improvement in streamlining the R2S workflow. Thanks for all your effort @zxkjack123!

Just some line comments here.

bam241:
The Changelog test is failing because you are following the PyNE project on CI.
So CI is running on your account and not the PyNE account.
The Changelog test needs the variable set in the PyNE account to run, but can't access them.

I would suggest you unfollow the PyNE project on you personal CircleCI account, which is not required unless you want CI to run on PR you do on your own PyNE fork.
Let me know if any of this does not make sense...

pyne/mesh.py
zxkjack123: Done in https://github.com/pyne/pyne/pull/1341/commits/e9020ad9f10e21d928064808cad076131b064b1d.

src/source_sampling.cpp
zxkjack123: Done in https://github.com/pyne/pyne/pull/1341/commits/0315a7f4cde0e9080c8d4f77b27e4a5f3665a7e3.

src/source_sampling.cpp
zxkjack123: Done in https://github.com/pyne/pyne/pull/1341/commits/296650d9683e6f43ebbb2dab15ff59edfccf2593.

pyne/source_sampling.pyx
zxkjack123: Parameter description updated in https://github.com/pyne/pyne/pull/1341/commits/6adbeb57e8a52ead1ffe7f28c2140750edfd98a1.

docs/usersguide/source_sampling.rst
zxkjack123: No, the "source_density" is the (hard coded) value of the parameter "src_tag_name". It is still used in the code.

zxkjack123:
> The Changelog test is failing because you are following the PyNE project on CI.
> So CI is running on your account and not the PyNE account.
> The Changelog test needs the variable set in the PyNE account to run, but can't access them.
> 
> I would suggest you unfollow the PyNE project on you personal CircleCI account, which is not required unless you want CI to run on PR you do on your own PyNE fork.
> Let me know if any of this does not make sense...

Thanks @bam241 . Your method works.

pshriwise (approved):
Looks like everything has been addressed and this seems to work nicely. Thanks for your diligence @zxkjack123!

Will merge tomorrow if there are no other comments.

PR #1350, opened 2020-10-23
adding the missing const begin() end() for MaterialLibrary

bam241:
This add some of the missing API to allow the MAterialLibrary to act as a `map<string, shared_ptr<Material>>`

### DIFF
--- a/src/material_library.h
+++ b/src/material_library.h
@@ -151,9 +151,11 @@ class MaterialLibrary {
   typedef mat_map::const_iterator const_iterator;
   iterator begin() { return material_library.begin(); }
   iterator end() { return material_library.end(); }
-
+  const_iterator begin() const { return material_library.begin(); }
+  const_iterator end() const { return material_library.end(); }
   const_iterator cbegin() const { return material_library.cbegin(); }
   const_iterator cend() const { return material_library.cend(); }
+
   std::size_t size() const { return material_library.size(); }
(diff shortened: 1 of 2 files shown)


bam241:
ping @pshriwise @gonuke 
(for @gonuke, @pshriwise told me that some compilers were complaining (when compiling OpenMC against DAGMC) that the const version of those iterator were not present. this is a quick fix that I would like to make into DAGMC...

src/material_library.h
```diff
@@ -151,9 +151,11 @@ class MaterialLibrary {
...
   iterator begin() { return material_library.begin(); }
   iterator end() { return material_library.end(); }
-
+  const_iterator begin() const { return material_library.begin(); }
+  const_iterator end() const { return material_library.end(); }
   const_iterator cbegin() const { return material_library.cbegin(); }
   const_iterator cend() const { return material_library.cend(); }
+
   std::size_t size() const { return material_library.size(); }
   bool emtpy() const { return material_library.empty(); }
```
pshriwise: Looks good to me, while we're at it let's fix the typo here.
```suggestion
  bool empty() const { return material_library.empty(); }
```

pshriwise (changes requested):
Just one small additional correction to a method name. Otherwise, this looks just fine to me.

pshriwise (approved):
Looks good! Merging now as these are rather trivial changes.

PR #1352, opened 2020-10-28
Tag decay_time and total_photon_source_intensity to source.h5m

zxkjack123:
## Description
This PR tags the 'decay_time' and 'total_photon_source_intensity' to 'source.h5m' in r2s step2 for better data consistency. 

## Changes
Previously, the 'decay_time' and 'total_photon_source_intensity' and 'source_density' information are stored in different files. It's prone to mistake the data of different versions/runs/parameters without any warning/error. Now, users can read them from the same file. 

## Todo
- [x] : Tag the 'file format versioning' (#1349 ) info to 'source.h5m' with 'tag_phtn_src_info()'

Issue #1349, opened 2020-10-20
File format versioning for R2S (and all?) output files.


pshriwise:
The output files of R2S (as of #1341 I think this may be only the photon source file) are subject to occasional change. It seems to me that workflows and applications relying on these files would benefit from the addition of version numbers to ensure that a certain format specification can be expected and compatibility with other applications can be ensured.

I'm not as familiar with it anymore, but it seems like this would likely benefit the `MaterialLibrary` as well.

zxkjack123:
@pshriwise Can't agree more. The version conflicts happens frequently for me. A pyne version number (x.xx.xx) is not enough to identify the specific version as many it covers many PRs. Maybe the id of the last PR or the hash id of the commit is a good choice of version tag.

pshriwise:
Glad to hear you're on board! I think the most common practice is to update the version numbers only when code changes are merged that change the information or organization of the file.

zxkjack123:
Close this PR as file format version has been tagged to `source.h5m` in #1352 .

PR #1353, opened 2020-10-30
Fix dagmc read e bounds

zxkjack123:
## Description
The `e_bounds` are tagged into `source.h5m` in PR #1341 . 
The unit of energy boundaries in the previous `e_bounds` text file is `MeV`. In PR #1341 , however, it was set to be `eV`. 
For better consistency, the unit of `e_bounds` tag is changed to `MeV` in this PR.
This PR also fixes an `e_bounds` not error when the old sampler is used without providing `e_bounds` text file.


zxkjack123:
@pshriwise Can you have a look at this PR? It's just a small fix.

pshriwise:
Sorry for the delay, @zxkjack123. Can I suggest that we rely on the version number from #1352 to determine if the `source.h5m` file should have energy bounds or not?

zxkjack123:
@pshriwise I am not quite understand your suggesiton. To make it clearer, let's assume a PyNE user run R2S:
- case 1: pyne verison before v0.7.1 (there isn't the code to write e_bounds in `source.h5m`) is used, and the sampler is older one. The user can run photon transport with two kinds of `source.h5m`:
    - case 1.1: the `source.h5m` generated by the user, without `e_bounds` inside, it will work if the text file name `e_bounds` provided.
    - case 1.2: the `soruce.h5m` generated by another user with higher pyne version, with `e_bounds` inside, it will work too. The `e_bounds` here is not used.
- case 2: pyne version after v0.7.1, the `e_bounds` will automatically write into `source.h5m`. 
    - case 2.1: It works with the new generated `source.h5m` with `e_bounds` tagged, and `e_bounds` tag in `source.h5m` will be used.
    - case 2.1: It works with older `source.h5m` without `e_bounds` too, a text file named `e_bounds` should be provided.

The current situation is, there will be `e_bounds` tag in `soruce.h5m` in higer version pyne. But the pyne version does not influence the source sampling process. Users do not need to care about their `source.h5m` version.

**The pyne version tag is used to identify the change of structure/information in the `source.h5m`**, as disscussed in https://github.com/pyne/pyne/issues/1349. If the contents in the `source.h5m` is changed, for example, in this PR, the unit of `e_bounds` is changed from `eV` to `MeV`, the pyne version tag should be changed. (In another word, this PR should be merged **after** #1352).

pshriwise:
Thanks for breaking this down further @zxkjack123. Here's how I'm suggesting we handle these two cases:

  - case 1: pyne verison before v0.7.1 (there isn't the code to write e_bounds in source.h5m) is used, and the sampler is older one. 
    - The user will use the old combination of `source.h5m` and `e_bounds` files. If a newer `source.h5m` is provided, the workflow will fail as it cannot find the `e_bounds` file.

  - case 2: pyne version after v0.7.1, the e_bounds will automatically write into source.h5m
    - The user will use a new `source.h5m` with tagged energy boundary information. If no version number or an incompatible version number is found, then the workflow will return an error to the user.

In the end, what I'm suggesting is that we move forward without supporting the old format that relied on the combination of multiple, text-based files. I'm of the opinion that moving all of this information into the `.h5m` file is a step in the right direction and that we should encourage users to update to this new file format. A script could be written to convert the combination of old files into the single `source.h5m` file if that is of interest to users wishing to re-run R2S using old file formats.

pshriwise:
> The pyne version tag is used to identify the change of structure/information in the source.h5m, as disscussed in #1349. If the contents in the source.h5m is changed, for example, in this PR, the unit of e_bounds is changed from eV to MeV, the pyne version tag should be changed. (In another word, this PR should be merged after #1352).

Seeing as we haven't yet started tagging these files with a version, I was thinking we could incorporate this fix before merging #1352 and avoid having to immediately update the `source.h5m` version here.

zxkjack123:
@pshriwise Agreed. Encourage users to use newer version of the code is more friendly for developers, and the code will be cleaner and more concise. 
Actually, there are a quite a mount of redundant in the current code such as the samplers in the `source_sampling.cpp` and old test functions. It deserves some cleanup PRs.

pshriwise (approved):
A simple fix and some good discussion here. Thanks @zxkjack123!

Issue #1357, opened 2020-12-10
pyne::ValueError::what() returns a pointer to a local scope variable


pshriwise:
**Describe the Bug**<br/>

https://github.com/pyne/pyne/blob/010c3d3bd2b1530aea4d7e00f2771a129a028673/src/utils.h#L229-L236

The currently returned pointer uses a string, `msgstr` that is of local scopt. The returned pointer should rely on a variable that will remain in memory after the method returns.

**To Reproduce**<br/>
```cpp
#include <string>
#include <iostream>

// PyNE's utils header
#include "utils.h"

int main() {

  throw pyne::ValueError("Error message");

  return 0;
}

```

**Screenshots or Code Snippets**<br/>
Code above produces this:?

```shell
$ ./main
terminate called after throwing an instance of 'pyne::ValueError'
  what():  
```

PR #1358, opened 2020-12-10
Return pointer for the member variable in ValueError::what()

pshriwise:
A proposed fix for #1357.

The main change here is that the `"ValueError: "` prefix is added in the constructor instead of in the `::what` method. The main consequence here is that extracting the original message becomes more tricky, but in practice, I think this an exceedingly rare use case for the object.

Just a note that this is _fix_, but the best long term solution may be to inherit from `std::runtime_error` or the like, which manages the error message and any exceptions that may arise from the internal string's construction (i.e. `ValueErrror::message`) as well.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Next Version
 
 **Change**
 
+   * Update pyne::ValueError to return a pointer to a stored string instead of an out-of-scope string. (#1358)
    * Change the e_bounds tag unit from eV to MeV (#1353)
    * Add functions to tag the decay_time, source_intensity and version to source.h5m (#1352)
 
@@ -18,7 +19,7 @@ Next Version
 
 **Maintenance**
 
-v0.7.3 
+v0.7.3
 ======
 
 **Change**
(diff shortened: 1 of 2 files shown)


gonuke:
Another approach would be to have 2 member variables, one with the user message & one with the full message. 

I don't have a strong feeling for that, but thought I'd mention it before merging

pshriwise:
A fair point. If you'd prefer that approach I can update it. Seeing as these are typically used immediately after construction it probably doesn't matter much in practice. I thought it would be more simple to keep the one member variable.

gonuke:
This solution is fine.  Thanks @pshriwise

gonuke:
Looks like it needs a changelog entry

bam241 (approved):
LGTM!
thx @pshriwise

bam241:
test are failing because you're following your PyNE Fork on CI....

pshriwise:
Just unfollowed the project... it looks like many of the jobs ran successfully though? Does only the `py3_dagmc_pymoab_openmc_test` job need me to unfollow?

pshriwise:
Seems like this is looking good now @gonuke @bam241. Thanks for the reviews!

bam241:
Merging after CI passes
Thx @pshriwise

bam241:
Merging !

Issue #1340, opened 2020-09-21
Build failure: __pyx_builtin_ValueError was not declared in this scope


bunder:
**Describe the Bug**<br/>
```
/home/aleontiev/python-pyne/src/pyne-0.5.11/build/pyne/transmuters.cxx: In function 'PyObject* __pyx_pf_4pyne_11transmuters_cram(PyObject*, PyObject*, PyObject*, int)':
/home/aleontiev/python-pyne/src/pyne-0.5.11/build/pyne/transmuters.cxx:2841:70: error: '__pyx_builtin_ValueError' was not declared in this scope; did you mean '__pyx_builtin_ImportError'?
 2841 |     try { throw; } catch(const std::exception& exn) {PyErr_SetString(__pyx_builtin_ValueError, exn.what());} catch(...) { PyErr_SetNone(__pyx_builtin_ValueError); }
      |                                                                      ^~~~~~~~~~~~~~~~~~~~~~~~
      |                                                                      __pyx_builtin_ImportError
/home/aleontiev/python-pyne/src/pyne-0.5.11/build/pyne/transmuters.cxx:2841:137: error: '__pyx_builtin_ValueError' was not declared in this scope; did you mean '__pyx_builtin_ImportError'?
 2841 |     try { throw; } catch(const std::exception& exn) {PyErr_SetString(__pyx_builtin_ValueError, exn.what());} catch(...) { PyErr_SetNone(__pyx_builtin_ValueError); }
      |                                                                                                                                         ^~~~~~~~~~~~~~~~~~~~~~~~
      |                                                                                                                                         __pyx_builtin_ImportError
make[2]: *** [pyne/CMakeFiles/transmuters.dir/build.make:88: pyne/CMakeFiles/transmuters.dir/transmuters.cxx.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:915: pyne/CMakeFiles/transmuters.dir/all] Error 2
```

**To Reproduce**<br/>
Build with `FFLAGS=-fallow-argument-mismatch python setup.py build`.

**System information:**<br/>
 - OS: ArchLinux
 - PyNE Version: 0.7.1 and 0.5.11
 - Versions of dependencies installed:
   - Python 3.8.5
   - Cython 0.29.14-0.29.21 (I've checked all versions)
   - GCC 10.2.0

arekfu:
[...]

PR #1362, opened 2021-01-07
use except + in transmuters

zxkjack123:
## Description
A build failure related to Numpy , Cython and `cpp_transmuters.pxd` is reported in https://github.com/pyne/pyne/issues/1340. 
This error happens when Numpy 1.19.x is used.
The solution is proposed in https://github.com/pyne/pyne/issues/1340#issuecomment-752052914, and presented in this PR.

## Changes
Change the way to raise an Except to avoid build failure with numpy v1.19.x.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ Next Version
 **Fix**
 
    * An e_bounds reading problem when old sampler is used without e_bounds text file (#1353)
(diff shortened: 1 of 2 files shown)


bam241 (approved):
LGTM!
thx for the quick fix @zxkjack123 

Will try to merge this quick if no-one object.

PR #1364, opened 2021-01-16
Fix to __delitem__ method of MaterialLibrary. 

pshriwise:
This fixes a call in `MaterialLibrary.__delitem__`. Also adds a couple of tests for the material removal methods of the `MaterialLibrary`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,7 @@ Next Version
 
    * An e_bounds reading problem when old sampler is used without e_bounds text file (#1353)
    * Fix the compatibility problem of transmuters with numpy version v1.19.x
+   * Fix to MaterialLibrary.__delitem__ to call the underlying instance.
 
 **Maintenance**
 
--- a/pyne/material_library.pyx
+++ b/pyne/material_library.pyx
@@ -248,7 +248,8 @@ cdef class _MaterialLibrary:
         return self._inst.size()
 
     def __delitem__(self, key):
-        self.del_material(ensure_material_key(key))
+        c_key = ensure_material_key(key)
+        self._inst.del_material(< std_string > c_key)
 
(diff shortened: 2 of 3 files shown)


bam241:
Thx a lot for taking the time for fixing this.


wondering if it would not be better to update:
````
    def remove_material(self, mat_name):
        """Remove a Material from this material library.
        Parameters
        ----------
        mat_name : str Name of the material be removed from this material library
        """
        if isinstance(mat_name, basestring):
            c_matname = mat_name.encode('UTF-8')
            self._inst.del_material(< std_string > c_matname)
        else:
            raise TypeError("the argument must be a string (material name) but is a "
                            "{0}".format(type(mat_name)))
```
and then call it ?

pshriwise:
I'm honestly not sure. It looks like `__delitem__` is designed to take either an `int` or `str` based on the `ensures_mat_key` function, but `remove_material` is always expecting a string. Depends on what the expectations for the different methods are, but I'm not familiar enough with this object at this point to say.

pshriwise:
> I'm honestly not sure. It looks like __delitem__ is designed to take either an int or str based on the ensures_mat_key function, but remove_material is always expecting a string. Depends on what the expectations for the different methods are, but I'm not familiar enough with this object at this point to say.

That is to say, I'll go with whatever you recommend :smiley:

bam241 (approved):
> I'm honestly not sure. It looks like `__delitem__` is designed to take either an `int` or `str` based on the `ensures_mat_key` function, but `remove_material` is always expecting a string. Depends on what the expectations for the different methods are, but I'm not familiar enough with this object at this point to say.

make sense. 
this is good to me.

Merging in a few!

bam241:
thx @pshriwise !
Merging!

PR #1375, opened 2021-05-19
Fix issues with empty mds_line

makeclean:
Please follow these guidelines when making a Pull Request.
This template was adapted from [here](https://github.com/stevemao/github-issue-templates/blob/master/questions-answers/PULL_REQUEST_TEMPLATE.md) and [here](https://github.com/stevemao/github-issue-templates/blob/master/conversational/PULL_REQUEST_TEMPLATE.md).

## Description
Fixed a bug that occured upon reading a material

## Motivation and Context
Fixes a material reading bug

## Changes
Bug fix

## Behavior
Broken now fixed

## Other Information
N/A

### DIFF
--- a/pyne/mcnp.py
+++ b/pyne/mcnp.py
@@ -1507,6 +1507,9 @@ def mat_from_inp_line(filename, mat_line, densities='None'):
                     # set metadata
         line_index += 1
         mds_line = linecache.getline(filename, mat_line - line_index)
+        # its possible this line is blank and will fail on the .split operation
+        if not mds_line: 
+            break
 
     # Check all the densities. If they are atom densities, convert them to mass
     # densities. If they are mass densities they willl be negative, so make


makeclean:
Why did CI not launch?

makeclean:
Ah there we go

pyne/mcnp.py
```diff
@@ -1507,6 +1507,9 @@ def mat_from_inp_line(filename, mat_line, densities='None'):
                     # set metadata
         line_index += 1
         mds_line = linecache.getline(filename, mat_line - line_index)
+        # its possible this line is blank and will fail on the .split operation
```
gonuke, line 1510: which split fails?  A completely blank line is not valid MCNP input.  Do you have an example that fails?

gonuke (changes requested):
Would be great to have a smallest failing example to see what problem this solves? It could be added to tests?

gonuke:
Also needs an addition the `CHANGELOG.rst`

PR #1387, opened 2021-06-25
Correct reference for atomic mass data in theory manual

alexmalins:
## Description
Corrects the reference for atomic mass data in the data theory manual page, and adds a link to the AMDC.

## Motivation and Context
The atomic mass data was attributed to JAEA, but looking back through old discussions on the PyNE groups it looks like PyNE has  always used AMDC atomic mass data.

Note: I'm not aware if JAEA publish atomic mass data or not: I've not checked.

## Changes
Self-explanatory

## Changelog file
Will commit an entry now after finding out the number of this PR.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,8 @@ Next Version
      
 **Fix**
 
+   * Attribution for the atomic mass data in the data theory doc page (#1387)
+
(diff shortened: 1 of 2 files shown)


gonuke:
Thanks @alexmalins

PR #1390, opened 2021-06-30
fixes for Windows cpp17

bam241:
filesystem::canonical() needs to be converted to char* ...

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -391,11 +391,11 @@ std::string pyne::get_full_filepath(std::string filename) {
   filename = pyne::remove_characters(" " , filename);
   // use stdlib call
 #ifndef _WIN32
-  const char* full_filepath = realpath(filename.c_str(), NULL);
+  const std::string full_filepath = realpath(filename.c_str(), NULL);
 #else
-  const char* full_filepath = std::filesystem::canonical(filename.c_str()).string().c_str();
+  const std::string  full_filepath = std::filesystem::canonical(filename.c_str()).string();
 #endif
-  return std::string(full_filepath);
+  return full_filepath;
 }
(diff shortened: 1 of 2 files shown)


bam241:
`const char* full_filepath = std::filesystem::canonical(filename.c_str()).string().c_str();` could have been directly casted as a string:
`string full_filepath = std::filesystem::canonical(filename.c_str()).string();`
but I kept it that way for consistency, but can change it if anyone feels strongly about this...

src/utils.cpp
```diff
@@ -393,7 +393,7 @@ std::string pyne::get_full_filepath(std::string filename) {
 #ifndef _WIN32
   const char* full_filepath = realpath(filename.c_str(), NULL);
 #else
-  const char* full_filepath = std::filesystem::canonical(filename.c_str());
+  const char* full_filepath = std::filesystem::canonical(filename.c_str()).string().c_str();
```
gonuke: We are about to recast this as a `std::string`.  Can't we just make `full_filepath` be `std::string` and avoid the conversion to `char*`?

gonuke (changes requested):
I thought I had already submitted this review (my iPad app is throwing me off)...

and now I see that you already recognized the same thing.  I think it makes sense to define `full_filepath` as a `std::string` in both instances.  The results of `realpath` will be promoted to the `std::string` on assignment.

src/utils.cpp
bam241: yes that's what I was suggesting. I kept it that way so no windows and windows have the same kind of behavior and the temp variable has the same type...

gonuke (approved):
LGTM - thanks @bam241

PR #1392, opened 2021-07-16
Auto docker build using GitHub action

bam241:
this should allow to build the 6 docker image we use in CI.

It is fairly repetitive but I want to leverage docker cache build, this is why I did not build this as a matrix.

I am very open to discuss this point.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,37 +5,37 @@ version: 2.1
 executors:
     py3:
         docker:
-            - image: pyne/ubuntu_18.04_py3_pyne-deps:latest
+            - image: ghcr.io/pyne/ubuntu_18.04_py3_pyne-deps:latest
               auth:
                 username: $DOCKERHUB_USER
                 password: $DOCKERHUB_PASS
     py3_dagmc:
         docker:
-            - image: pyne/ubuntu_18.04_py3_dagmc_pyne-deps:latest
+            - image: ghcr.io/pyne/ubuntu_18.04_py3_dagmc_pyne-deps:latest
               auth:
                 username: $DOCKERHUB_USER
                 password: $DOCKERHUB_PASS
     py3_pymoab:
         docker:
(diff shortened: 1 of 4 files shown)


gonuke:
Do we need to enable actions - this doesn't seem to be running actions

bam241:
it is running on push, not on pull_request that's why.

but it is actually running on my fork: https://github.com/bam241/pyne/actions/runs/1037940249

gonuke:
This doesn't seem to be making the correct combinations... is it still being tested?

bam241:
is on my fork

https://github.com/bam241/pyne/runs/3089736086?check_suite_focus=true

.circleci/config.yml
```diff
@@ -5,37 +5,37 @@ version: 2.1
 executors:
     py3:
         docker:
-            - image: pyne/ubuntu_18.04_py3_pyne-deps:latest
+            - image: ghcr.io/pyne/ubuntu_18.04_py3_pyne-deps:latest
               auth:
```
gonuke, line 9: Do we still need the authorization once we move to GHCR?

gonuke (changes requested):
I see the images now... I keep getting confused by the different PRs and image locations.

.circleci/config.yml
bam241: no we don't

bam241:
> Do we need to enable actions - this doesn't seem to be running actions

but you might be right, I don't understand why #1393 GitHub action is not running....

bam241:
maybe we should hold off here before we don't know for sure....

gonuke:
Thanks @bam241 - I'll merge this now b/c the last commit won't trigger a new image build anyway, by design

gonuke:
> maybe we should hold off here before we don't know for sure....
> 
> 

Too late 😀 - I think i was ok to merge this

gonuke:
We can watch the outcome here: https://github.com/pyne/pyne/actions/runs/1045248495

gonuke:
Missing CHANGELOG here as well!

PR #1393, opened 2021-07-16
Moving CI from Circle to GithubAction

bam241:
attempting to move all CI jobs into GitHub actions


bam241:
@gonuke I can't get the workflow to trigger here.

I got it running on my fork....

bam241:
@gonuke could you approve my PR ? I am wondering if this setting 
```
Require approval for first-time contributors
Only first-time contributors will require approval to run workflows.
```
is the source of the problem ?!

gonuke:
can you rebase to get the circleci update (which is only temporary, anyway)

bam241:
done

gonuke:
and now the GH tests are running!?!?

bam241:
> and now the GH tests are running!?!?

I don't get it, but thx for the fix !

gonuke:
you need to make an entry in the CHANGELOG

bam241:
I'll do that when I have the ci fixed

I prefer to do the changeling at the end, when I know everything I changed.

bam241:
changeling added and added the missing changelog for #1393

changelog test is not working yet

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,109 @@
+name: Standard Build and Test
+
+on:
+  # allows us to run workflows manually
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - develop
+      - master
+  push:
```
gonuke: Do we want to limit which branches this triggers for `push`

bam241:
All thx worked !

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,109 @@
...
+          '',
+          _dagmc
+        ]
+        openmc : [
+          '',
+        ]
+        hdf5 : [
+          '',
+        ]
+        include:
```
gonuke, line 54: I'll have to have you explain how this include works relative to the matrix :)

gonuke (commented):
A couple of questions...

.github/workflows/build_test.yml
bam241: the includes add more combinaison to the one defined to the matrix.
for example the `dagmc=''` `pymoab=''` `hdf5=''` `openmc=''` does not exist in the matrix layout (as pymoab is always present)

bam241: you can use `include` to add more combinaison (with values that don't exist in the matrix) or exclude to `remove` some that exists

.github/workflows/build_test.yml
bam241: we could probably limit them to develop...
and open the PR to all branches ?

gonuke: Doing it on push for all branches, means every push to my own branch on my own repo will trigger an action (for better or worse) - right?

bam241: yes

gonuke: Isn't it wasteful to trigger this on every push? Do we have a best practice?

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,109 @@
+name: Standard Build and Test
+
+on:
+  # allows us to run workflows manually
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - develop
+      - master
```
gonuke: There is no master branch in PyNE

gonuke (changes requested):
I think we should err on the side of limiting CI actions to fewer cases at first

.github/workflows/build_test.yml
bam241: we don't I think requesting a PR would allow to limit the extend of it.
if one needs the CI, one can always issue a PR in a fork

bam241:
here is my guess on the best option:
```
on:
  # allows us to run workflows manually
  workflow_dispatch:
  pull_request:
  push:
    branches:
      - develop
```

gonuke (approved):

gonuke:
This dispatch plan looks good to me... thanks @bam241

gonuke:
and again with the changelog test on merge :facepalm:

PR #1395, opened 2021-07-20
limit changelog check to PR only

gonuke:
Tweak the Github Action CI to only check the changelog on PR and not on merge


gonuke:
This passes PR tests but the real change will be tested on merge.

bam241:
looking good.

thx @gonuke

PR #1397, opened 2021-07-21
Optimize mesh to fluxin

zxkjack123:
## Description
Modify the string concatenation method to speed up function `alara.mesh_to_fluxin` irradiation setup of r2s step.

## Motivation and Context
For R2S with fine mesh with a large amount of volume elements, the current R2S step1 takes a very long time (several days for 400,000 volume elements).
Most of the time is used in the function of alara.mesh_to_fluxin, more specifically, the `alara._output_flux`.
The reason is the low efficiency of string concatenation.

## Changes
Put the string segments into a list, and then concatenation with join. See [here](https://waymoot.org/home/python_string/#:~:text=%20Efficient%20String%20Concatenation%20in%20Python%20%201,build%20the%20string%20and%20the%20amount...%20More%20) for reference.

## Behavior
The running time of alara.mesh_to_fluxin is recorded as below, it gives the behavior change before/after this PR. 
|number of volume elements| 4 | 100 | 1000 | 10000 | 386,848 |
|------------|----|---|---|---| --- |
|mesh_to_fluxin running time before change (s)| 3e-3 | 1e-2 | 0.39 | 36.3 | several days |
|mesh_to_fluxin running time after change (s) | 1.0e-3 | 1.8e-2 | 0.14 | 1.46 | 685.9 |

PR #1398, opened 2021-07-21
Fix test variancereduction

zxkjack123:
## Description
This PR tries to fix the random testing failure of test_variancereduction.

## Motivation and Context
The `test_variancereduction.py` sometimes fails even a PR has nothing to do with variance reduction, for example, the PR #1397 .
But if you rerun the tests, the `test_variancereduction.py` may succeed without any error.

## Changes
This PR updates the meshtally by returning it from `variancereduction.magic` to avoid the wrong reference.

## Behavior
It is hoped to avoid the wrong reference of meshtally after `magic`.


zxkjack123:
This does not fix the problem.

gonuke:
Thanks for trying... I was going to look into this myself - it's very annoying!

PR #1401, opened 2021-07-22
Mesh to fluxin progress bar

zxkjack123:
## Description
This PR uses a progress bar to show the progress of `alara.mesh_to_fluxin`.

## Motivation and Context
The progress print is useful for loops that take a relatively long time. The progress package can be used to do it.

## Behavior
The progress bar for `alara.mesh_to_fluxin` looks like this:
```
Writing alara fluxin |################################| 100.0% - 0s
```

## Performance
The running time of alara.mesh_to_fluxin is recorded as below, it gives the behavior change before/after this PR. 

|number of volume elements| 100 | 1000 | 10000 |
|------------|----|----|----|
| without progress bar |  1.8e-2 | 0.14 | 1.46 | 
| with progress bar | 1.75e-2 | 0.167 | 1.67 |

A slight performance down (14% slower) is acceptable as the `alara.mesh_to_fluxin` does not take too much time now.

## Others
The progress bar can be applied to other functions/loops in PyNE. That's could be done if people like it.

### DIFF
--- a/pyne/alara.py
+++ b/pyne/alara.py
@@ -19,6 +19,7 @@ import numpy as np
 import tables as tb
 from io import open
 import re
+from progress.bar import Bar
 
 QA_warn(__name__)
 
@@ -43,7 +44,7 @@ response_strings = {'decay_heat': 'Total Decay Heat',
 
 def mesh_to_fluxin(flux_mesh, flux_tag, fluxin="fluxin.out",
                    reverse=False, sub_voxel=False, cell_fracs=None,
-                   cell_mats=None, print_progress=100000):
+                   cell_mats=None):
     """This function creates an ALARA fluxin file from fluxes tagged on a PyNE
     Mesh object. Fluxes are printed in the order of the flux_mesh.__iter__().
 
@@ -81,10 +82,6 @@ def mesh_to_fluxin(flux_mesh, flux_tag, fluxin="fluxin.out",
         Maps geometry cell numbers to PyNE Material objects.
         The cell_fracs and cell_mats are used only when sub_voxel=True.
         If sub_voxel=False, neither cell_fracs nor cell_mats will be used.
-    print_progress: int, optional
-        If desired, the number of processed events can be printed to the
-        console each N loops by passing the print_progress=N parameter.
-        The print progress can be turned off by by setting print_progess=0.
     """
 
     tag_flux = flux_mesh.get_tag(flux_tag)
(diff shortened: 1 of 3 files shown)


zxkjack123:
@bam241 Progress bar is suggested for loops that take a long time, https://github.com/pyne/pyne/pull/1397#discussion_r673859473.
It requires a `progress` package. I tried to modify the docker file to install `progress`. But failed. How to fix it?

gonuke:
I think it may all be working, @zxkjack123.

The new docker images are built under your account: https://github.com/zxkjack123/pyne/actions/runs/1055430434

Here you can see they built and then tested (with the VR failure occurring (look at #1399 for a fix).

Normally, I think we will expect docker image updates to be separate from the code updates that rely upon them.  This way, we can test the images and deploy them, before testing the code.

gonuke:
Can you submit this again without the code change - just the docker change?  Then we can manage that one and update the images for testing the code changes in CI.

zxkjack123:
> Can you submit this again without the code change - just the docker change? Then we can manage that one and update the images for testing the code changes in CI.

Sure, done in #1402

gonuke:
You'll need to relaunch CI after the docker containers have updated - it takes about 30 minutes (if successful)

pyne/alara.py
```diff
@@ -43,7 +44,7 @@
 
 def mesh_to_fluxin(flux_mesh, flux_tag, fluxin="fluxin.out",
                    reverse=False, sub_voxel=False, cell_fracs=None,
-                   cell_mats=None, print_progress=100000):
+                   cell_mats=None):
```
gonuke: can we still have a way for users to turn this off?

gonuke (changes requested):
This looks good - just one request about letting users turn off the progress bar

gonuke:
This is now passing tests on the updated docker images... so just a matter of resolving my review.  Thanks for your patience @zxkjack123

pyne/alara.py
zxkjack123: @gonuke The commit https://github.com/pyne/pyne/pull/1401/commits/0dcce921c2ec13f4128c1d1beef4317970337d3f allow users to turn this off.

pyne/alara.py
```diff
@@ -94,17 +96,28 @@ def mesh_to_fluxin(flux_mesh, flux_tag, fluxin="fluxin.out",
...
-            with Bar("Writing alara fluxin", max=len(flux_mesh), suffix='%(percent).1f%% - %(eta)ds') as bar:
+            if print_progress:
+                bar = Bar("Writing alara fluxin", max=len(flux_mesh), suffix='%(percent).1f%% - %(eta)ds')
                 for i, mat, ve in flux_mesh:
                     f.write(_output_flux_block(ve, tag_flux, reverse))
                     bar.next()
+                bar.finish()
+            else:
+                for i, mat, ve in flux_mesh:
+                    f.write(_output_flux_block(ve, tag_flux, reverse))
```
gonuke: I see that's a little messy!

This might work to clean it up a little bit
```suggestion
            bar = Bar("Writing alara fluxin", max=len(flux_mesh), suffix='%(percent).1f%% - %(eta)ds') if print_progress else None
            for i, mat, ve in flux_mesh:
               f.write(_output_flux_block(ve, tag_flux, reverse))
               bar.next() if print_progress else None
            bar.finish() if print_progress else None
```

Another option would be to extend Bar?

```
class IfBar(....., show_bar = True):

    def init(....):

        self.show_bar = show_bar

        if not self.show_bar:
            pass

        super().__init__(....)

    def next():
        if self.show_bar:
            super().next()

    def finish():
        if self.show_bar:
            super().finish()
```

gonuke (changes requested):
I guess that option makes it complicated is you're using that progress bar...

I have a couple of suggestions to streamline it - but neither one is great.

pyne/alara.py
zxkjack123: Agreed. The turn on/off of the print progress requires more indents. The `IfBar` is add in #1409 .

zxkjack123:
Update: IfBar is used here to print progress to cleanup the code.

gonuke (commented):
Thanks @zxkjack123 - it'll be nice to have this progress bar for other things, as well

gonuke (approved):

PR #1402, opened 2021-07-22
docker add progress package

zxkjack123:
## Description
Add progress package into docker image.

## Motivation and Context
The package `progress` is used to show progress of some codes.


gonuke:
Thanks @zxkjack123 - this looks good.

Issue #1403, opened 2021-07-22
Don't run build & test workflow if only changes are in docker image generation


gonuke:
Our Github Actions based CI launches a standard Build & Test action on every merge.  For merges that are only changing the images used for CI, this is unnecessary, and a Build & Test action will be done as part of the docker publishing.

We can exclude CI-specific paths (`ignore-paths`) for the Build & Test action.

gonuke:
We need to add the ChangeLog test for all paths.

gonuke:
This was resolved in #1404

PR #1404, opened 2021-07-22
dedicated changlog workflow

bam241:
this move changelog check in a dedicated workflow

adds path constrain (ignore) to build_test so won't run on docker changes


bam241:
this should address #1403

gonuke:
Thanks @bam241 - LGTM!

PR #1407, opened 2021-07-22
change tag name on CI docker images

gonuke:
## Description
Change the tag used for generating temporary docker images for testing from `latest` to `ci_testing`.

## Motivation and Context
We have been using `latest` as the temporary tag for docker images during testing, but this is a common/default tag used for the most recent stable build.  This change uses `ci_testing` while testing and then changes is it to `stable` and `latest` when done.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -5,14 +5,14 @@ on:
   workflow_dispatch:
   pull_request:
     paths-ignore:
-      - 'workflow/docker_publish.yml'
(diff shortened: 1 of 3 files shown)


gonuke:
Watch the docker builds [here](https://github.com/gonuke/pyne/actions).

bam241:
seems fine.
only one way to be sure....
Merging
thx @gonuke

gonuke:
You'll need to rebase and manage conflicts for #1408 now

PR #1409, opened 2021-07-23
extend Bar to IfBar

zxkjack123:
## Description
Add a IfBar for eaiser turn on/off of the progress print.

## Motivation and Context
Truning on/off of  `progress.bar` make the code messy. This PR add a extended IfBar according to the suggestion of @gonuke [here](https://github.com/pyne/pyne/pull/1401#discussion_r675255868).

## Changes
The `class IfBar` in `pyne.utils`, and a test.

## Others
The `IfBar` will hopefully make the code of progress bar in #1401 cleaner.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,7 @@ Next Version
    * removing old circle-ci related stuff (#1405)
    * adds a manual trigger for docker build workflow (#1406)
    * clarify the tag name for docker images generated in CI (#1407)
+   * add ifbar to cleanup the code about progress bar (#1409)
 
 v0.7.5
 ======
--- a/pyne/utils.py
+++ b/pyne/utils.py
@@ -6,6 +6,7 @@ from warnings import warn
 from distutils.dir_util import remove_tree
 import filecmp
 from io import open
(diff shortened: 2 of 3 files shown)


gonuke (commented):
This looks good to me - thanks for taking the time to implement this @zxkjack123 .

For testing, we normally try to separate generating the output from writing/printing the output, but that's not possible in this case.  I wonder if something like this is worth it? 

https://stackoverflow.com/questions/4219717/how-to-assert-output-with-nosetest-unittest-in-python

zxkjack123:
> This looks good to me - thanks for taking the time to implement this @zxkjack123 .
> 
> For testing, we normally try to separate generating the output from writing/printing the output, but that's not possible in this case. I wonder if something like this is worth it?
> 
> https://stackoverflow.com/questions/4219717/how-to-assert-output-with-nosetest-unittest-in-python

@gonuke I don't think it's worth to use tools such as context manager for the purpose of tesing print strings of progress bar. The test in this PR has tested that the progress bar works when if turned on/off. The detailed printed strings on the screen does not influence the runing of the program.

PR #1410, opened 2021-07-23
limits build_test and housekeeping to pull_request trigger only

bam241:
we run the exact `build_test` and `housekeeping` with the exact same condition on push(merge) and pull_request, 

I think it is a bit overkill and will trigger things in forks without people necessarily be aware of them...
so I thought we might as well limit ourselves to pull_request....

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -9,14 +9,7 @@ on:
       - '.github/workflows/changelog_test.yml'
       - 'docker/**'
       - 'CHANGELOG.rst'
-  push:
-    branches:
-      - develop
-    paths-ignore:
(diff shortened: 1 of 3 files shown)


.github/workflows/build_test.yml
```diff
@@ -9,14 +9,7 @@ on:
       - '.github/workflows/changelog_test.yml'
       - 'docker/**'
       - 'CHANGELOG.rst'
-  push:
```
gonuke, line 12: haven't we traditionally done a built & test on merge (aka push)?  I'm on the fence about whether or not it is necessary, but removing this is a departure from our CircleCI testing (I think)

gonuke (commented):
Thanks @bam241 - not sure if this is something we should discuss.

PR #1417, opened 2021-10-08
Adding method to write material library in an OpenMC XML format.

pshriwise:
## Description
These changes add a method to `MaterialLibrary` that allow one to write the materials in an XML format for use with OpenMC.

## Motivation and Context
For verification, it is sometimes easier to have the materials in a format that OpenMC can read. This is also useful when looking to apply material definitions from a PyNE material library to OpenMC models.

### DIFF
--- a/src/material_library.cpp
+++ b/src/material_library.cpp
@@ -1,6 +1,7 @@
 #ifndef _WIN32
 #include <unistd.h>
 #endif
+#include <fstream>
 #include <iostream>
 
 #ifndef PYNE_IS_AMALGAMATED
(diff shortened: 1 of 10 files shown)


src/material_library.cpp
```diff
@@ -135,6 +136,23 @@ void pyne::MaterialLibrary::write_json(const std::string& filename) {
   f.close();
 }
 
+void pyne::MaterialLibrary::write_openmc(const std::string& filename) const
+{
+  std::ofstream f(filename.c_str());
+  // write header
+  f << "<?xml version=\"1.0\"?>\n";
+  f << "<materials>\n";
```
gonuke, line 144: Is there a desire for any tabs/indentation in these files?

gonuke (commented):
This looks ok to me, just one non-blocking question

src/material_library.cpp
pshriwise: Mmm, good point. I think this function is fine since the `<materials`> tag isn't indented, but the `<material`> tags had some indentation that was off.

I've corrected those and added an argument to the `Material::openmc` method for the indention level (1 by default assuming the materials are being written directly into a standard materials XML file w/o modification).

Changes in 12eeab0e74cebea0e44191a7eed2db612e8cf7cb.

gonuke (approved):
Thanks @pshriwise

PR #1415, opened 2021-09-20
Release workflow update

bam241:
this will change:
- [x] CMakeFile will generate header for python and CPP to deal with  PYNE_VERSION number 
- [x] change the DAGMC target in CI Docker image to target DAGMC/`stable` tag (just been created)

PR #1427, opened 2021-11-02
Update version strings for new release

gonuke:
Updating the two places that refer to the version number of a release in preparation for a microrelease.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,11 @@ PyNE Change Log
 Next Version
 ============
 
+
+
+v0.7.6
+======
+
 **New Capabilities**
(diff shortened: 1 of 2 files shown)


shimwell:
Looks much easier than I was expecting, just the one version number change in a couple of files. I had a look around for other places where the version might be needed but this looks like everything. Many thanks to the implementer of the PYNE_VERSION varible, it is a great time saver. I have updated the DAGMC community docs to include these steps in the documented release list.

shimwell (approved):

gonuke:
> Many thanks to the implementer of the PYNE_VERSION varible, it is a great time saver. 

That was @bam241 in #1415

PR #1428, opened 2021-11-02
Make Pyne Version actions into a CMake include

gonuke:
To make the action available to other CMake workflows, the actions to set and invoke the version number is moved to a file that can be included.

This is driven by the need to invoke these actions when PyNE is being amalgamated from a SubModule

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,17 +11,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 # Setup the project
 project(pyne)
 
-
-# Set PYNE version
-set(PYNE_MAJOR_VERSION 0)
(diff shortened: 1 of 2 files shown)


shimwell:
All the CI has successfully passed :tada: and this includes build so I think these changes are well tested.

I can see that the variables that have been moved are still accessible to the CMakeLists.txt as it calls ```include(PyneVersion)``` where the variables now reside. Looks like it will not impact the PyNE code at all and if it makes PyNE more usable elsewhere in SubModule form then that sounds great.

shimwell (approved):

shimwell:
Happy to merge this into the RC branch if there are no objections

gonuke:
Please do merge this.  There may need to be more changes here to accommodate the submodule use case(s), but let's go with this for now.

PR #1430, opened 2021-11-09
Update version macro to allow for specifying path

gonuke:
## Description
Introduce an optional path for specifying where the PyNE version substitution should occur.

## Motivation and Context
When PyNE is included as a submodule, it may be necessary to process the PyNE Version substitution without running directly CMake on the PyNE directory.  A new Version macro was introduced in #1428, but it still assumes that CMake is running in the base PyNE source directory.

This change allows the user/developer to specify a path to the PyNE source.

## Changes
This is a new feature from PyNE's point of view, but contributes to a bug fix in downstream packages.


gonuke:
This may not be the right way to do this... closing to try a different approach

PR #1431, opened 2021-11-11
Improving ability to invoke PyNE Version CMake macro when submodule

gonuke:
## Description
Introduce features and variables that allow the PyNE Version CMake macro to be called from a different path location

## Motivation and Context
The PyNE Version macro works for build from within PyNE, but needs more support when PyNE is a submodule in a different path.

## Changes
This can be considered a feature enhancement with changes in the PyNE Version macro.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,7 +16,7 @@ v0.7.6
 **Change**
    * exclude docker related stuff from build_test (#1404 #1408)
    * move changelog test in its own workflow (#1404)
-   * Adds 2 templates header files (one for Python one for CPP) to store PYNE version number, cmake will automatically configure proper version files for python and cpp based on the PYNE VERSION number located in the root CMakeFile.txt (#1415)
+   * Adds 2 templates header files (one for Python one for CPP) to store PYNE version number, cmake will automatically configure proper version files for python and cpp based on the PYNE VERSION number located in the root CMakeFile.txt (#1415, #1428)
    * PyNE CI docker container now target DAGMC:'stable' tag instead of develop. (#1415)
 
 **Fix**
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,9 +16,9 @@ v0.7.6
 **Change**
    * exclude docker related stuff from build_test (#1404 #1408)
    * move changelog test in its own workflow (#1404)
-   * Adds 2 templates header files (one for Python one for CPP) to store PYNE version 
-     number, cmake will automatically configure proper version files for python and cpp 
-     based on the PYNE VERSION number located in the root CMakeFile.txt (#1415, #1428, #1430)
+   * Adds 2 templates header files (one for Python one for CPP) to store PYNE version number, 
+     cmake will automatically configure proper version files for python and cpp based on the 
+     PYNE VERSION number located in the root CMakeFile.txt (#1415, #1428)
    * PyNE CI docker container now target DAGMC:'stable' tag instead of develop. (#1415)
 
 **Fix**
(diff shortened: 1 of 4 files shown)


gonuke:
I think this has all the features necessary to allow DAGMC get the version updated successfully before building

gonuke:
see here for tests of a version of DAGMC using this hash of pyne: https://github.com/gonuke/DAGMC/actions/runs/1454211195

gonuke:
Once this is merged, I'll update the DAGMC PR to refer to the merged hash

gonuke:
Needed another DAGMC update, so hopefully passing tests will be here: https://github.com/gonuke/DAGMC/actions/runs/1454268099

gonuke:
I've added Mac and Windows testing, available here: https://github.com/gonuke/DAGMC/actions

shimwell:
> Needed another DAGMC update, so hopefully passing tests will be here: https://github.com/gonuke/DAGMC/actions/runs/1454268099

All 12 passed :tada:

shimwell:
> I've added Mac and Windows testing, available here: https://github.com/gonuke/DAGMC/actions

all 3 passed :tada:

shimwell (approved):

pshriwise (changes requested):
Tried this out and it works quite well! No line comments from me. The only thing I'll request is updates to the docs (maybe near info on amalgamation) to indicate how to use PyNE as a submodule in another project.

Though if we think the situation with using PyNE as a submodule in DAGMC is a one-off occurrence, I'm fine w/ skipping the docs.

gonuke:
> Tried this out and it works quite well! No line comments from me. The only thing I'll request is updates to the docs (maybe near info on amalgamation) to indicate how to use PyNE as a submodule in another project.

Yeah.... so this raises an additional question about ensuring the version substitution happens prior to amalgamation.  At the moment, that solution lives in DAGMC.  I can document how to invoke the version  substitution from a code that includes PyNE as a submodule, but DAGMC then also has a custom amalgamation script, etc...  I think we need a slightly better solution that lives in PyNE before we can document it well.

pshriwise:
Hrmm so you're suggesting we perhaps move the new amalgamation script from DAGMC to PyNE?

gonuke:
No - that won't quite do it, either.

In DAGMC, I call a CMake macro for version substitution (because we're already in the middle of a CMake build), and then we call a bash script that ultimately calls the normal amalgamate script.

Since we are relying on CMake-based version substitution, we need to somehow invoke CMake at the beginning of the amalgamate process.  It might be about making a separate CMakeList config file for amalgamation that invokes the same macro?

gonuke:
But I'm not sure we can invoke that from within the amalgamation python script, because we call that from DAGMC's CMake build and that might just mess things up.... although it's invoked from within a new bash shell (too many layers!)

pshriwise:
Not very pretty, but we could make it an option to run the CMake preprocessing in the amalgamation script? Standard use would execute the CMake substitution, but we'd tell it not to in DAMGC b/c we handle that with the `PYNE_SUBMODULE_PATH` variable?

I'm wondering if we, unfortunately, approached this the wrong way and maybe should have stored the version information in the Python side and generated a `pyne_version.h` file using Python instead?

gonuke:
I think this was right because we probably will ultimately want to have CMake put this info in other places (see current issue with MOAB...)

gonuke:
If we did it in the amalgamation script, we could back out changes in DAGMC and get rid of the SUBMODULE_PATH stuff?

gonuke:
withdrawing this in favor of #1434

Issue #1429, opened 2021-11-02
Ensure PyNE Version substitutions before amalgamation


gonuke:
There are currently no guardrails that prevent PyNE amalgamation before the PyNE Version CMake actions have been invoked.  The consequence will be that we amalgamate a version of PyNE that has placeholders for the PyNE version that have not be replaced yet.

You can see this by running `amalgamate.py` on fresh checkout of the repo.

Ideally, `amalgamate.py` will invoke a CMake workflow that ensures that the PyNE Version substitutions have taken place.

PR #1434, opened 2022-02-06
Move PyNE versioning to python

gonuke:
## Description
Change the system for single definition of PyNE version. 

## Motivation and Context
 A fairly recent update (#1415) created a CMake macro to hold a single location to define the version, and then introduce it to both a Python and C++ file to be included in other places, during configuration.  This posed problems for code that includes PyNE, especially code that includes C++ directly as the C++ header was not processed by the upstream code.

This should provide a mechanism to address #1429 

## Changes
This change allows the version to be defined once in small Python code that can then be directly included in other python and also be run as a script to generate the necessary C++ header.


gonuke:
I have a working DAGMC branch that uses this commit of PyNE: https://github.com/gonuke/DAGMC/actions/runs/1803894281
(on linux, at least)

pyne/pyne_version.py
```diff
@@ -0,0 +1,26 @@
...
+# Set PYNE version
+PYNE_MAJOR_VERSION = 0
+PYNE_MINOR_VERSION = 7
+PYNE_PATCH_VERSION = 6
+
+PYNE_VERSION = f'{PYNE_MAJOR_VERSION}.{PYNE_MINOR_VERSION}.{PYNE_PATCH_VERSION}'
+
+def write_cpp_header(path="."):
+
+    with open(os.path.join(path,'pyne_version.h'),'w') as pvh:
```
pshriwise:
```suggestion
    with open(os.path.join(path, 'pyne_version.h'), 'w') as pvh:
```

pshriwise:
```suggestion
    with open(os.path.join(path, 'pyne_version.h'), 'w') as pvh:
```

pshriwise (approved):
I gave this a try as well. Tested with PyNE standalone, DAGMC's develop branch, and DAGMC's latest release candidate branch.

Thanks @gonuke! Just some minor formatting comments. I'll merge this tomorrow if there are no additional comments.

pyne/pyne_version.py
```diff
@@ -0,0 +1,26 @@
+import os
+
+# Set PYNE version
+PYNE_MAJOR_VERSION = 0
+PYNE_MINOR_VERSION = 7
+PYNE_PATCH_VERSION = 6
+
+PYNE_VERSION = f'{PYNE_MAJOR_VERSION}.{PYNE_MINOR_VERSION}.{PYNE_PATCH_VERSION}'
```
pshriwise: One final thought here - PyNE requires Python >= 3.5.2, but f-strings aren't officially supported until 3.6. We should probably use `.format` here.

gonuke: Good call (or should we up the min version of python??)

pshriwise: Probably should. No time to determine what the right min version should be at the moment though.

pyne/pyne_version.py
```diff
@@ -0,0 +1,26 @@
...
+PYNE_MAJOR_VERSION = 0
+PYNE_MINOR_VERSION = 7
+PYNE_PATCH_VERSION = 6
+
+PYNE_VERSION = '{0}.{1}.{2}'.format(PYNE_MAJOR_VERSION, PYNE_MINOR_VERSION, PYNE_PATCH_VERSION)
+
+def write_cpp_header(path="."):
+
+    with open(os.path.join(path , 'pyne_version.h') , 'w') as pvh:
+        header_text = f'''#ifndef PYNE_VERSION_HEADER
```
pshriwise: f-string here as well.

pyne/pyne_version.py
shimwell: OpenMC recently moved to a minimum python version of 3.6 perhaps the same motivating reasons apply
https://github.com/openmc-dev/openmc/pull/1860

.gitignore
```diff
@@ -69,5 +69,10 @@ rever/
 !tests/files_test_openmc/*.h5
 
 .vscode
-pyne/pyne_version.py
+
+# ignore version file generated from pyne_version.py
 src/pyne_version.h
+
+# ignore amalgamated source
+pyne.cpp
```
pshriwise, line 77: Nice addition 👌🏻

pshriwise (approved):

PR #1441, opened 2022-02-11
Fix build process for docker images for CI

### LINKED ISSUE #1440: Building of docker images for CI broken by changes in pip packages
**Describe the Bug**<br/>
Version management of packages provided by PIP has broken the process of building our docker images for CI.  Specifically, CI runs on Ubuntu 18.04 that only supports up to python 3.6 as a standard package.  Some newer pip-installable require python 3.7+ and causes failures.  More subtly, even though numpy installs a valid version, matplotlib tries to install a new version of numpy on top that is too new.

**To Reproduce**<br/>
This is reproduced by launching our `docker_publish.xml` action.

**Expected Behavior**<br/>
It should create a valid docker image for use in CI.

gonuke:
## Description
Refine versions of some python packages to fix build process for docker images for CI

Fixes #1440 

## Motivation and Context
Changes to python packages that are pip installed caused docker images to fail to build.  Specifically, `matplotlib` was reinstalling a version of `numpy` that was too new for python 3.6, and `setuptools` had deprecated necessary install features.

## Changes
Constrain versions of pip-installed packages.

### DIFF
--- a/docker/ubuntu_18.04-dev.dockerfile
+++ b/docker/ubuntu_18.04-dev.dockerfile
@@ -3,7 +3,7 @@ FROM ubuntu:18.04
 # Ubuntu Setup
 ENV TZ=America/Chicago
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
-ARG py_version=3.8
+ARG py_version=3.6
 
 ENV HOME /root
 RUN if [ "${py_version%.?}" -eq 3 ] ; \
--- a/docker/ubuntu_18.04-dev.dockerfile
+++ b/docker/ubuntu_18.04-dev.dockerfile
@@ -46,6 +46,7 @@ RUN if [ "${py_version%.?}" -eq 3 ] ; \
             sphinx \
             cloud_sptheme \
             prettytable \
+            "setuptools<49" \
             sphinxcontrib_bibtex \
             numpydoc \
             nbconvert \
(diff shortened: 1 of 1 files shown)


gonuke:
This is being tested in this action: https://github.com/gonuke/pyne/actions/runs/1827574802

shimwell:
> This is being tested in this action: https://github.com/gonuke/pyne/actions/runs/1827574802

Looks like everything passed, I guess the last step of the CI called ```pusing_tst_stable_img``` only runs on a release

gonuke:
> > This is being tested in this action: https://github.com/gonuke/pyne/actions/runs/1827574802
> 
> Looks like everything passed, I guess the last step of the CI called `pusing_tst_stable_img` only runs on a release

Essentially, that's right.  Technically, it will only run when the repo owner is `pyne`, but the whole thing only runs on push and not on PR, so the result is the same.  This is to make sure we don't change CI images prematurely.

shimwell:
The changes all look reasonable to me. I have downloaded the fork and am currently building the docker image with this command 

```
git clone git@github.com:gonuke/pyne.git pyne_gonuke_fork
cd pyne_gonuke_fork/
git checkout fix_docker_deps
cd docker
docker build -f ubuntu_18.04-dev.dockerfile .
```

gonuke:
That command line process does not do the full process of the rather complicated Github action....

shimwell:
taking a peak inside the docker image I can see that we have these package versions with the default docker build

matplotlib version 3.3.4
tables version 3.6.1
Setuptools 48.0.0

shimwell (approved):

pshriwise (approved):
Based on offline discussions along with those described in the PR, this all looks reasonable to me. Hopefully we can get these version compatibility issues sorted out soon.

Thanks for digging into this, @gonuke!

gonuke:
I have a PR into Moab to fix it on their end. And we'll move CI to 20.04 & Python 3.8 soon.

PR #1444, opened 2022-04-14
Add option to turn off last FORTRAN code (ENSDF) and default to no fortran

gonuke:
## Description
Add a CMake option to turn off the building of ENSDF processing and turn off that and spatial solvers by default.

## Motivation and Context
ENSDF Fortran code is and has features no longer supported by newer compilers.  To avoid updating the FORTRAN and avoid being stuck with old compilers, these are being turned off by default.

## Changes
Change in build process/options.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ Next Version
 
 **Maintenance**
    * update workflow to accommodate security-driven changes (#1445)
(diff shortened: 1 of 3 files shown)


gonuke:
This now passes tests.  I think it makes sense to merge this and then have someone see if we can build in a docker image that has no FORTRAN support at all.

ngranda (commented):
LGTM! Thanks @gonuke. I'll merge

PR #1447, opened 2022-05-18
Fix for compatibility with newer versions of tables

ngranda:
## Description
Make NUC_DATA_PATH a string

## Motivation and Context
Currently, PyNE must be built with a tables package version lower than 3.7
Using tables package 3.7 or higher makes this variable type byte when it should be a string. 
This change allows for compatibility with newer versions of tables


gonuke:
Thanks @ngranda - this should help us move forward with other dependencies (e.g. pytables)

PR #1448, opened 2022-05-18
try newer versions of some pip packages

gonuke:
## Description
Remove version limits on some python packages installed by pip in CI images.

## Motivation and Context
In order to enable a recent release, we were forced to limit the version of `setuptools` and `pytables` for compatibility. Some recent changes in PyNE have removed the cause (maybe)?

## Changes
This is a build maintenance change.


gonuke:
Waiting for CI to test this on my repo... this may not work, though because of python 3.6

gonuke:
Withdrawing this in favor of update to Ubuntu 20

PR #1454, opened 2022-06-07
fix mcnp6 version check for serial

zxkjack123:
## Description
A small fix for Ptrac version check.

## Behavior
The below code fails when the Ptrac is generated with the serial mcnp6:
https://github.com/pyne/pyne/blob/800c75a6099f81ac73460166aea936dea4835f00/pyne/mcnp.py#L1233-L1234

This PR fixes this problem.


pyne/mcnp.py
```diff
@@ -1231,7 +1231,7 @@ def read_variable_ids(self):
 
         if self.eightbytes:
             mcnp_version = self.mcnp_version_info[8:13]
```
gonuke: Is it more robust to `strip()` the whitespace from the end of this?
```suggestion
            mcnp_version = self.mcnp_version_info[8:13].strip()
```

gonuke (changes requested):
Thanks for this update.  I wonder why we didn't find it before!

I have suggested a slightly different approach that may be more robust.

pyne/mcnp.py
ahnaf-tahmid-chowdhury: Sure, but I like @zxkjack123 to make the change and what he thinks.

ahnaf-tahmid-chowdhury:
The file `mcnp6_serial_ptrac_i4_little.ptrac` is a binary file. I think we need to verify that `self.endianness` in PtracReader matches the actual file’s endianness. Also, We should consider adding a try-except block to catch and log the values of `length` and `length2` when they don't match for better output.

zxkjack123:
@gonuke @ahnaf-tahmid-chowdhury I forgot this PR for a long time. I made a new repo with the same changes to a new PR #1558 to avoid the rebase conflicts fixing. Please move to the #1558 for further discussion.

gonuke:
Please feel free to close this now that #1558 is merged, @zxkjack123

zxkjack123:
Close this PR as #1558 merged.

PR #1456, opened 2022-07-21
tiny typo fix

shimwell:
Really tiny fix for a couple of typos. Probably not worth a PR but just in case


gonuke (approved):
Thanks @shimwell

PR #1459, opened 2022-09-05
Add composite action

bquan0:
## Description
Added a composite action to setup QEMU and log into GHCR and another composite action to setup Docker Buildx and Cache Docker layers for the github workflows. These composite actions are used to replace 5 sections of code in `docker_publish.yml`.

## Motivation and Context
This is a change requested by @gonuke, and it simplifies the `docker_publish.yml` workflow by combining some of its frequently used action sequences. 

## Other Information
I grouped "setup QEMU" and "log into GHCR" into one composite action. However, I omitted the "checkout repository" action in this composite action since "checkout repository" was only used in 2 of the 3 sections where "setup QEMU" and "log into GHCR" were used. 
I grouped "Set up Docker Buildx" and "Cache Docker layers" into one composite action. These two actions were used together in 2 different sections.


gonuke:
Thanks @bquan0 - this looks like a good step forward, as we all learn more about this capability.

gonuke:
I've requested reviews from a few others who have some expertise/experience

shimwell:
Thanks for pinging me to this PR, interesting stuff. I guess this sort of thing scales really well compared to the current solution. I note that this particular change currently doesn't save any lines of code but I like the approach anyway.

shimwell:
Does anyone think it is worth making the setup-QEMU-login-GHCR action have a default token value of ```${{ secrets.GITHUB_TOKEN }}``` if such a thing is possible

gonuke:
> Thanks for pinging me to this PR, interesting stuff. I guess this sort of thing scales really well compared to the current solution. I note that this particular change currently doesn't save any lines of code but I like the approach anyway.

It does save cognitive bandwidth, though!  Especially if we make sure the composite actions are well documented.  That documentation lives outside of our main workflows, so doesn't clutter them.

.github/actions/setup-buildx-cache/action.yml
```diff
@@ -0,0 +1,19 @@
...
+        install: true
+
+    - name: Cache Docker layers
+      uses: actions/cache@v2
+      with:
+        path: /tmp/.buildx-cache-${{ github.sha }}
+        # Key is named differently to avoid collision
+        key: ${{ runner.os }}-multi-buildx-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-multi-buildx
```
bam241:
```suggestion
          ${{ runner.os }}-multi-buildx
          
```

usually it is advisable to end a file with a line return

bam241 (approved):
Great job, 

LGTM (looks good to me)!

BTW @gonuke this might contain the catching solution you were looking for... (I forgot I was able to make it work. This should allow caching between serialized jobs as it allows to specify were jobs are caching their docker build cache)

.github/actions/setup-buildx-cache/action.yml
bam241: Maybe adding more detailed description to the composite action as suggested by @gonuke would be nice.

gonuke:
Thanks @bquan0 - this is a good start for us to revise how we do this

Issue #1460, opened 2022-09-10
Confirm need for QEMU in docker image builds


gonuke:
We currently run the `setup-QEMU` action when building docker images for CI.

It's not obvious that we need to do this - I think it's only really necessary for cross-platform builds.

It doesn't cause technical problems, but makes it more complicated to understand how things work.

gonuke:
Resolved by #1461

PR #1461, opened 2022-09-14
Remove setup QEMU

bquan0:
## Description
I removed the setup QEMU step from the composite action used in `docker_publish.yml`. I also renamed the composite action from `setup-QEMU-login-GHCR` to `login-GHCR`, and updated references in `docker_publish.yml` accordingly.

## Motivation and Context
This change was requested by @gonuke in #1460.

## Behavior
The `docker_publish.yml` workflow appears to work even after the removal of the setup QEMU step, which suggests that the setup QEMU step isn't needed.


gonuke:
Alas ... without the QEMU login, the composite action makes less sense now, I think, since it only does one thing.

Maybe we should combine the GHCR login, buildx and cache setup in a single action, and possibly set some (environment?) variables for use in later steps.  For example, setting the name of the cache once and reusing it can ensure that we have consistency.  It might also make sense to set the basename for the docker image tag: `ghcr.io/${{ github.repository_owner }}/ubuntu_20.04_py3` for reuse/simplification.

bquan0:
I combined the GHCR login with buildx and cache setup composite action, and deleted the standalone login composite action. I also made an env variable for the basename for the docker image tag. I'm not sure what to do for setting the name of the cache, however.

.github/workflows/docker_publish.yml
```diff
@@ -194,7 +191,7 @@ jobs:
 
 
     container:
-      image: ghcr.io/${{ github.repository_owner }}/ubuntu_20.04_py3${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:ci_testing
+      image: ${{ matrix.env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:ci_testing
```
gonuke:
```suggestion
      image: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:ci_testing
```

I think this is causing the build test to fail

gonuke (changes requested):
I think I found a typo that is causing the bootstrap build test to fail.

.github/workflows/docker_publish.yml
bquan0: That's strange... I think I tried it without the `matrix` before and the workflow wouldn't run at all, saying that "env is unrecognized". This was in the workflow run that happened with the `fix basename variable references` commit.

gonuke: I'd be curious to see that error

bquan0: Here's the link to the workflow with the error:
https://github.com/bquan0/pyne/actions/runs/3251869373/workflow

gonuke: That is definitely different from what I expect.  I guess it might make sense since the environment hasn't been established yet - we are defining the container in which we will creating the environment.

However, that variable is also not defined in the matrix.  Maybe we need to also define a variable that is in scope within the Github action?

gonuke: note: it should NOT be added to the matrix, I don't think

gonuke: In fact, we might just want to reuse the original variables in place rather than defining a new variable to be used in a single place.

bquan0: Looks like reusing the original variable in the matrix section solved the problem.

gonuke (approved):
This looks like a good solution for now.

PR #1467, opened 2022-12-15
Fix reorder error. 

### LINKED ISSUE #1466: Fix member variable order to build with -Wreorder
I'm trying to build Pyne using MOOSE's environment, which passes the `-Wreorder` compile flag. This is hitting the following error:

```
/home/anovak/cardinal/contrib/DAGMC/src/pyne/pyne.h: In constructor 'pyne::Sampler::Sampler(std::string, std::string, std::vector<double>, std::string)':
/home/anovak/cardinal/contrib/DAGMC/src/pyne/pyne.h:6243:25: error: 'pyne::Sampler::e_bounds' will be initialized after [-Werror=reorder]
 6243 |     std::vector<double> e_bounds;  ///< Energy boundaries
      |
```

Solution is to just fix the member variable order in the header file.

aprilnovak:
Closes #1466

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ Next Version
    * remove setup-QEMU step from composite action(#1461)
    * add composite actions to github workflows(#1459)
    * fixing typos in user warning message (#1456)
+   * fix reorder error in member variables (#1466)
 
 v0.7.7
 ======
--- a/src/source_sampling.h
+++ b/src/source_sampling.h
@@ -209,12 +209,12 @@ namespace pyne {
     // problem parameters
     std::string filename; ///< MOAB mesh file path
     std::string src_tag_name; ///< Unbiased source density distribution
-    std::string bias_tag_name; ///< Biased source density distribution
     std::string cell_number_tag_name; ///< Cell number tag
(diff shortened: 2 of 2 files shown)


gonuke:
Thanks @aprilnovak - do you have any idea why this generates an error?

aprilnovak:
Yes, the `-Wreorder` compilation flag checks that the order of members declared in a class definition matches that in the member initialization list of a class. The underlying reason for this flag is that, when a class member is declared, it's also initialized (most often to some unknown value, but you could also have something like `int a_ = 1;`. Later in the member initialization list, it's possible that member variable is used to initialize other members so the order could matter.

The particular fix here is that the order in the header file did not match the member initialization list.

aprilnovak:
I think [this post](https://stackoverflow.com/questions/1828037/whats-the-point-of-g-wreorder#:~:text=The%20problem%20is,itself%20been%20initialised.) also does a good job of explaining it as well.

pshriwise (approved):
Thanks @aprilnovak! This is a pretty straightforward improvement, but I'll give it 24 hrs in case others have any additional comments/questions.

gonuke:
Thanks again @aprilnovak - a new tip to remember for all our projects 😀 (as well as building with warnings on)

PR #1469, opened 2022-12-15
Remove unusable c-linkage

### LINKED ISSUE #1468: Remove unusable C-linkage in source_sampling.h
If you try to build Pyne with MOOSE's conda environment (which uses the `-Wreturn-type-c-linkage` flag), you get an error due to unusable C-linkage:

```
/Users/vishalpatel/mambaforge3/envs/moose/include/pyne.h:6104:23: error: 'read_e_bounds' has C-linkage specified, but returns user-defined type 'std::vector<double>' which is incompatible with C [-Werror,-Wreturn-type-c-linkage]
  std::vector<double> read_e_bounds(std::string e_bounds_file);
```

`std::vector` is a C++ type, so it's not possible to compile functions that return this type with C. At least this is my understanding of what is happening.

aprilnovak:
Closes #1468

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,7 @@ Next Version
    * add composite actions to github workflows(#1459)
    * fixing typos in user warning message (#1456)
    * fix reorder error in member variables (#1466)
+   * remove unusable C-linkage with std::vector (#1468)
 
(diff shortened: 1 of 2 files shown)


gonuke:
Thanks @aprilnovak

Issue #1463, opened 2022-11-19
Implement a multistage build for testing docker images


gonuke:
The current test strategy for PyNE is to test on the following variations:

1. base python
2. base python + pymoab
3. base python + pymoab + dagmc
4. base python + pymaob + dagmc + openmc
5. base python + hdf5 1.12 + pymoab + dagmc

We build these with a complex github action that uses a matrix to build different variations in parallel.  Instead, we could use a multistage Dockerfile and build the first 4 in multiple stages using the multistage build action that will store all of the variations anyway.

It would still be useful to rely on the HDF5 build_arg to distinguish between 3 and 5.

gonuke:
This was resolved in #1470, #1471, and #1473

PR #1470, opened 2023-01-01
Multistage docker image build

bquan0:
## Description
This pull request modifies the `docker_publish.yml` workflow to use a [multistage docker build action](https://github.com/marketplace/actions/multi-stage-docker-build). It also adds stages to the dockerfile and removes some build args that were previously used to accomplish the same result as the multistage docker build action. 
Note: the multistage build actions are currently not working and these changes can't be merged yet.

## Motivation and Context
This change should simplify the `docker_publish.yml` workflow and the dockerfile since many of the build args and logic will not be needed anymore. It should also allow images to be cached, which means that images that build off previous images don't need to build those previous images from scratch and can instead retrieve them from the cache and build off from there. 

This change was requested by @gonuke in #1463.

## Changes
This PR introduces a modification to the `docker_publish.yml` workflow and the dockerfile that should make building images more efficient.
 
## Behavior
Currently, the `docker_publish.yml` workflow builds many variations of PyNE in parallel by using build args and if statements in the dockerfile. These changes should make the workflow build the variations of PyNE in stages, and cache those stages so that they can be used to build later stages more quickly. These changes should also simplify `docker_publish.yml` and the dockerfile. 

## Other Information
There is an error when the first multistage docker build action runs, which can be seen [here](https://github.com/bquan0/pyne/actions/runs/3808830433/jobs/6479737855). It seems to be caused by some HDF5 variables not being set in the CMake files. I also tried running the first multistage action by enabling the HDF5 build arg, but the [workflow run](https://github.com/bquan0/pyne/actions/runs/3808598220/jobs/6479315165) ran into some errors when building dagmc.


.github/workflows/docker_publish.yml
```diff
@@ -34,30 +34,25 @@ jobs:
...
-          cache-to: type=local,mode=max,dest=/tmp/.buildx-cache-${{ github.sha }}
+          repository: ${{ env.DOCKER_IMAGE_BASENAME }}
+          stages: base_python, moab
+          server-stage: dagmc
+          quiet: false
+          dockerfile: docker/ubuntu_20.04-dev.dockerfile
+          build-args: py_version=3.8, build_hdf5=hdf5-1_12_0, build_moab=YES, enable_pymoab=YES
 
   # builds on the previous cached images adding extra layers to the base docker
   # images created in the previous step. The resulting images include various
```
gonuke: I think (?) your changes above replace all of this down through both the `build_and_push_dagmc` task and the `cleanup_cache` task

docker/ubuntu_20.04-dev.dockerfile
```diff
@@ -61,10 +62,11 @@ RUN if [ "$build_hdf5" != "NO" ]; then \
         && cd .. \
         && rm -rf hdf5; \
     fi
-# put MOAB on the path
+# put HDF5 on the path
 ENV LD_LIBRARY_PATH $HDF5_INSTALL_PATH/lib:$LD_LIBRARY_PATH
 ENV LIBRARY_PATH $HDF5_INSTALL_PATH/lib:$LIBRARY_PATH
 
+FROM base_python AS moab
```
gonuke, line 71: With the targets defined, I think we can skip the logic at the beginning of this section.  The `moab` target should always build moab and pymoab, I think - similar to the logic you removed in the `dagmc` target.

.github/workflows/docker_publish.yml
```diff
@@ -34,30 +34,25 @@ jobs:
...
-          build-args: |
-            py_version=3.8
-            build_pyne=NO
-          tags: ${{ env.DOCKER_IMAGE_BASENAME }}_pyne-deps:ci_testing
+          repository: ${{ env.DOCKER_IMAGE_BASENAME }}
+          stages: base_python, moab, dagmc
+          server-stage: openmc
+          quiet: false
+          dockerfile: docker/ubuntu_20.04-dev.dockerfile
+          build-args: py_version=3.8, build_hdf5=NO, build_moab=YES, enable_pymoab=YES
```
gonuke: per my comments below, we can probably remove the need for the `build_moab` and `enable_pymoab` arguments  by relying on the targets instead.

.github/workflows/docker_publish.yml
gonuke: Also, we'll need to rethink the `matrix` and other logic for the `BuildTest` task so that the container names used in line 194/189 are correct per the above multistage container builds

gonuke: Last, we do some clever work to build these images for testing before we move them over to be the stable images, and that will need to be corrected too.

gonuke (changes requested):
This is such an awesome contribution, @bquan0 .  I think there are a few more tweaks needed for consistency, but it will be such a valuable simplification when done!  I'll be happy to discuss some of them if/when necessary.

docker/ubuntu_20.04-dev.dockerfile
```diff
@@ -108,58 +110,51 @@ ENV LD_LIBRARY_PATH $HOME/opt/moab/lib:$LD_LIBRARY_PATH
...
+                -DBUILD_STATIC_LIBS=OFF \
+                -DBUILD_UWUW=OFF \
+                -DBUILD_TALLY=OFF \
+                -DBUILD_MAKE_WATERTIGHT=OFF \
+                -DBUILD_OVERLAP_CHECK=OFF \
+                -DBUILD_TESTS=OFF \
+    && make -j 3\
+    && make install \
+    && cd ../.. \
+    && rm -rf DAGMC; \
```
gonuke: Get rid of extra continuation info

```suggestion
    && rm -rf DAGMC
```

docker/ubuntu_20.04-dev.dockerfile
```diff
@@ -108,58 +110,51 @@ ENV LD_LIBRARY_PATH $HOME/opt/moab/lib:$LD_LIBRARY_PATH
...
-        git clone https://github.com/openmc-dev/openmc.git $HOME/opt/openmc \
-        && cd  $HOME/opt/openmc \
-        && pip install . ; \
-    fi
+RUN if [ "$build_hdf5" != "NO" ]; then \
+            export HDF5_ROOT="$HDF5_INSTALL_PATH" ; \
+    fi ;\
+    git clone https://github.com/openmc-dev/openmc.git $HOME/opt/openmc \
+    && cd  $HOME/opt/openmc \
+    && pip install . ; \
```
gonuke: Get rid of unnecessary continuation characters

```suggestion
    && pip install . 
```

gonuke (changes requested):
A little cleanup of the docker file with unnecessary continuation characters

gonuke:
The HDF5 error seems like some kind of configuration problem earlier.  However, it seems to be using a cached image for the `base_python` stage (https://github.com/bquan0/pyne/pkgs/container/ubuntu_20.04_py3%2Fbase_python) so we can't see where the error may be?  It might make sense to delete all the containers being built here between tests so that we can confirm whether things are working.

I wonder if you can just try running the dockerfile locally?

bquan0:
I tried deleting the base_python image and the pymoab image, but the HDF5 error still remains. I then tried running the dockerfile on my computer by using the `make_pyne_docker_image.py` script, and I ran into the same HDF5 error. I also tried running the dockerfile on the `develop` branch, and there was no HDF5 error, which makes me think that the error is somewhere in the dockerfile. However, I'm not entirely sure of that because I only removed some of the if statements surrounding the code, and that shouldn't affect HDF5.

bquan0:
I did some investigation by running some of the commands in the dockerfile on my computer. I ran `cmake .` and got the same HDF5 error. Then I tried running lines 13-27 of the dockerfile and running `cmake .` afterwards, and that seemed to fix the HDF5 error. This makes me think that the error has something to do with lines 13-27 not running correctly or not being cached correctly.

In the dockerfile, I tried moving `apt-get hdf5` related items into the moab section to see if that would solve the HDF5 error, but it didn't work.

gonuke:
I'm confused because when I run this dockerfile locally on my machine, it works OK.  I'll have to dig deeper when I have a little more time to see what's different about the Github build environment.  It's something subtle...

gonuke:
I learned something...  [In this test](https://github.com/gonuke/pyne/actions/runs/3919494616/jobs/6700535948), I used the normal `docker/build-push-action` to make target `moab` and then I also used the `firehed/multistage-docker-build-action` to make the same target.  The first one succeeded and the second one failed.

The failure is during `cmake` where it does not find HDF5.

gonuke:
I've created an issue to request that they move to `buildx` on the multistage action: https://github.com/Firehed/multistage-docker-build-action/issues/36

gonuke:
I have confirmed that this Dockerfile fails locally if I don't use BuildKit/buildx.  More investigation is needed.

gonuke:
Testing:
* build `moab` without BK in 2 stages in 1 build: FAIL
* build `moab` with BK in 2 stages: SUCCESS
* build `base_python` without BK and:
    * build `moab` without BK using this cache: FAIL
    * run image without BK (may not be relevant) and build MOAB manually: SUCCESS
* build `base_python` with BK and:
    *  build `moab` without BK using this cache: didn't use cache and FAIL

gonuke:
I think the problem is the `build_args`!!!

We rely on the `build_arg`s for different build logic and I think we need to explicitly reference the `build_hdf5` arg in the `moab` target so that it's value is known.  Otherwise, the value is not recognized as being "NO".  I think `buildx` may carry the `build_arg`s through across stages but without `buildx` they don't!

gonuke:
I think the best practice for `build_arg`s is the following:
1. declare them before the first `FROM` with default values
2. "import" them into each stage where they are used - they should get the default value unless overridden by the cmd-line value.  Don't declare the default again, just import them into the scope.

docker/ubuntu_20.04-dev.dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:20.04 AS base_python
```
gonuke, line 4:
```suggestion
ARG py_version=3.8
ARG build_hdf5="NO"

FROM ubuntu:20.04 AS base_python
```

docker/ubuntu_20.04-dev.dockerfile
```diff
@@ -22,8 +22,6 @@ RUN if [ "${py_version%.?}" -eq 3 ] ; \
             libblas-dev \
             liblapack-dev \
             libeigen3-dev \
-            libhdf5-dev \
-            hdf5-tools \
```
gonuke: Keep these here, it should work if they are here and the other problems are fixed.

docker/ubuntu_20.04-dev.dockerfile
```diff
@@ -61,105 +59,97 @@ RUN if [ "$build_hdf5" != "NO" ]; then \
...
         && rm -rf hdf5; \
     fi
-# put MOAB on the path
+# put HDF5 on the path
 ENV LD_LIBRARY_PATH $HDF5_INSTALL_PATH/lib:$LD_LIBRARY_PATH
 ENV LIBRARY_PATH $HDF5_INSTALL_PATH/lib:$LIBRARY_PATH
 
-ARG build_moab="NO"
-ARG enable_pymoab="NO"
+FROM base_python AS moab
```
gonuke, line 71:
```suggestion
FROM base_python AS moab
ARG py_version
ARG build_hdf5
```

docker/ubuntu_20.04-dev.dockerfile
```diff
@@ -61,105 +59,97 @@ RUN if [ "$build_hdf5" != "NO" ]; then \
...
+                                --moab $HOME/opt/moab --dagmc $HOME/opt/dagmc \
+                                $PYNE_HDF5_ARGS \
+                                --clean -j 3; \
 ENV PATH $HOME/.local/bin:$PATH
 RUN if [ "$build_pyne" = "YES" ]; then \
         cd $HOME \
         && nuc_data_make ; \
     fi
 
+FROM dagmc AS openmc
```
gonuke, line 152:
```suggestion
FROM dagmc AS openmc
ARG build_hdf5
```

docker/ubuntu_20.04-dev.dockerfile
```diff
@@ -61,105 +59,97 @@ RUN if [ "$build_hdf5" != "NO" ]; then \
...
-        fi \
-        && cd $HOME/opt \
-        && git clone -b develop --single-branch https://github.com/pyne/pyne.git \
-        && cd pyne \
-        && python setup.py install --user \
-                                    --moab $HOME/opt/moab --dagmc $HOME/opt/dagmc \
-                                    $PYNE_HDF5_ARGS \
-                                    --clean -j 3; \
-    fi
+RUN export PYNE_HDF5_ARGS="" ;\
```
gonuke, line 135:
```suggestion
FROM dagmc AS pyne
ARG build_hdf5

RUN export PYNE_HDF5_ARGS="" ;\
```

gonuke (changes requested):
Introducing some of the `build-arg` best practice I discovered today

docker/ubuntu_20.04-dev.dockerfile
bquan0: Should I change the stages in the multistage build action in `docker_publish.yml` to reflect the addition of this new `pyne` stage? Also, should the `openmc` stage still be based on `dagmc` or should it now be based on `pyne`?

bquan0:
Looks like the multistage build is now working. Thanks @gonuke for figuring out the problem with the build_args!

I changed the `matrix` of the `BuildTest` and the `pushing_test_stable_img` jobs to match the names of the new docker images, and the docker images built by the multistage docker build action are now correctly getting used in the tests. However, there are some common errors being thrown in the tests, like `numpy` not having `int` or `bool` attributes, and trying to create an inhomogeneous array (not all rows have the same number of columns). I think this might be caused by a different version of numpy being downloaded than before. I will probably look into this tomorrow

bquan0:
After looking into it briefly, it seems like `numpy` 1.24 is being downloaded, and there were a few changes with this version that caused the attribute errors. See [this link](https://levelup.gitconnected.com/fix-attributeerror-module-numpy-has-no-attribute-float-d7d68c5a4971) for more details. 
I downgraded to `numpy` 1.23 in the dockerfile, and that fixed the errors. According to the link, if we want to use `numpy` 1.24, we have to fix all the places where the errors were being thrown previously. 
There is a new error being thrown for the `openmc` test which says this: `OSError: statepoint file has a version of 17.0 which is not consistent with the version expected by OpenMC, 18`

gonuke:
It looks like OpenMC has updated its statepoint format and we need to update some of our test files in PyNE.  Here is a plan:
* immediately: peg the version of OpenMC to 0.13.0
* short term: regenerate the statepoint files using a newer version of OpenMC
* long term: have the testing framework generate its new statepoint files to use for testing with what ever version of OpenMC it has

bquan0:
I changed the version of OpenMC to 0.13.0, and now all the tests are working.

.github/workflows/docker_publish.yml
```diff
@@ -278,11 +136,11 @@ jobs:
       - name: Push Image to stable img
         uses: akhilerm/tag-push-action@v2.0.0
         with:
-          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:ci_testing
-          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:stable
+          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:refs_heads_multistage-docker-bk0:ci_testing
+          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:refs_heads_multistage-docker-bk0:stable
```
gonuke: I don't think this double tag works, and I'm not sure there was ever a `ci_testing` version of these in your new multistage build
```suggestion
          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:refs_heads_multistage-docker-bk0
          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:stable
```

.github/workflows/docker_publish.yml
```diff
@@ -278,11 +136,11 @@ jobs:
...
 
       - name: Push Image to latest img
         uses: akhilerm/tag-push-action@v2.0.0
         with:
-          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:ci_testing
-          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:latest
+          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:refs_heads_multistage-docker-bk0:ci_testing
+          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:refs_heads_multistage-docker-bk0:latest
```
gonuke: as above
```suggestion
          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:refs_heads_multistage-docker-bk0
          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:latest
```

docker/ubuntu_20.04-dev.dockerfile
gonuke: I think the openmc should be based on dagmc and not on PyNE

gonuke (changes requested):
I'm glad the OpenMC fix worked!

A few other items:
* It looks like you do 2 checkout actions, one after the other on lines 29 & 31
* we can probably get rid of your composite action and just move the docker/login back to the main workflow
* warning messages on your build suggest that we should update the version of the `checkout` and `docker/login` actions (there is also a warning about the multistage-build-action that I don't understand - it should be fixed in that action)
* some suggestions about  (re)naming of the final images

bquan0:
I updated the versions of the `checkout` and `docker/login` actions to the most recent versions, removed the composite action, and added the suggestions for the image names. Everything is working, let me know if there are any other changes I should make.

.github/workflows/docker_publish.yml
```diff
@@ -26,192 +26,73 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3.3.0
```
gonuke: Do you have to specify the full version number, or will just `v3` work? The reason I ask is that it is often nice to automatically take advantage of minor version updates without having to update the workflow file.

gonuke (changes requested):
Thanks again @bquan0 - Just one last (!) question...

.github/workflows/docker_publish.yml
bquan0: Based on what I've read [here](https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-release-management-for-actions) in the "Using tags for release management" section, it seems like `v3` should work if the author of the action follows the tip that says "Move the major version tag (such as `v1`, `v2`) to point to the Git ref of the current release". I'll edit the workflow to use the major releases like `v3`. 
If it doesn't work like that, there seems to be a few custom github actions that allow for automatic updating to use the most recent release of a github action.

.github/workflows/docker_publish.yml
```diff
@@ -8,15 +8,15 @@ on:
       - 'docker/*'
       - '.github/workflows/docker_publish.yml'
 
-env: 
-  DOCKER_IMAGE_BASENAME: ghcr.io/${{ github.repository_owner }}/ubuntu_20.04_py3
+env:
+  DOCKER_IMAGE_BASENAME: ghcr.io/${{ github.repository_owner }}/pyne/ubuntu_20.04_py3
```
gonuke: I am not confident in how all these directory-like names work.  I think it would be better if the basename was
```suggestion
  DOCKER_IMAGE_BASENAME: ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_20.04_py3
```

.github/workflows/docker_publish.yml
```diff
@@ -26,192 +26,73 @@ jobs:
...
-            pymoab: ''
-            openmc: ''
+          - stage: dagmc
+            hdf5: /hdf5
       fail-fast: false
 
 
     container:
-      image: ghcr.io/${{ github.repository_owner }}/ubuntu_20.04_py3${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:ci_testing
+      image: ghcr.io/${{ github.repository_owner }}/pyne/ubuntu_20.04_py3${{ matrix.hdf5 }}/${{ matrix.stage }}:refs_heads_multistage-docker-bk0
```
gonuke: Can you use ${{ env.DOCKER_IMAGE_BASENAME }} here?  Also switch to underscore instead of directory-like naming
```suggestion
      image: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}_${{ matrix.stage }}:refs_heads_multistage-docker-bk0
```

.github/workflows/docker_publish.yml
```diff
@@ -26,192 +26,73 @@ jobs:
...
-          - hdf5: ''
-            dagmc: _dagmc
-            pymoab: _pymoab
-            openmc: _openmc
-          - hdf5: ''
-            dagmc: ''
-            pymoab: ''
-            openmc: ''
+          - stage: dagmc
+            hdf5: /hdf5
```
gonuke: Consistent naming
```suggestion
            hdf5: _hdf5
```

.github/workflows/docker_publish.yml
```diff
@@ -238,51 +119,30 @@ jobs:
...
-          - hdf5: ''
-            dagmc: ''
-            pymoab: ''
-            openmc: ''
-
+          - stage: dagmc
+            hdf5: /hdf5
 
-    name: "${{ github.repository_owner }}/ubuntu_20.04_py3${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}: latest -> stable"
+    name: "${{ github.repository_owner }}/pyne/ubuntu_20.04_py3${{ matrix.hdf5 }}/${{ matrix.stage }}:refs_heads_multistage-docker-bk0: latest -> stable"
```
gonuke: consistent naming with the basename and underscore
```suggestion
    name: "${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}_${{ matrix.stage }}:refs_heads_multistage-docker-bk0: latest -> stable"
```

.github/workflows/docker_publish.yml
```diff
@@ -238,51 +119,30 @@ jobs:
...
-            dagmc: _dagmc
-            pymoab: _pymoab
-            openmc: _openmc
-          - hdf5: ''
-            dagmc: ''
-            pymoab: ''
-            openmc: ''
-
+          - stage: dagmc
+            hdf5: /hdf5
```
gonuke: consistent naming
```suggestion
            hdf5: _hdf5
```

.github/workflows/docker_publish.yml
```diff
@@ -238,51 +119,30 @@ jobs:
...
+          password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Push Image to stable img
-        uses: akhilerm/tag-push-action@v2.0.0
+        uses: akhilerm/tag-push-action@v2
         with:
-          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:ci_testing
-          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:stable
+          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:refs_heads_multistage-docker-bk0
+          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:stable
```
gonuke, line 142:
```suggestion
          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}_${{ matrix.stage }}:refs_heads_multistage-docker-bk0
          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}_${{ matrix.stage }}:stable
```

.github/workflows/docker_publish.yml
```diff
@@ -238,51 +119,30 @@ jobs:
...
 
       - name: Push Image to latest img
-        uses: akhilerm/tag-push-action@v2.0.0
+        uses: akhilerm/tag-push-action@v2
         with:
-          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:ci_testing
-          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}${{ matrix.dagmc }}${{ matrix.pymoab }}${{ matrix.openmc }}_pyne-deps:latest
+          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:refs_heads_multistage-docker-bk0
+          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:latest
```
gonuke, line 148:
```suggestion
          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}_${{ matrix.stage }}:refs_heads_multistage-docker-bk0
          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}_${{ matrix.stage }}:latest
```

gonuke (changes requested):
Sorry for still more comments - but just looking for consistent naming of images.

.github/workflows/docker_publish.yml
bquan0: I think we don't use the env variable here because of an error we ran into in #1461 where the env variable couldn't be used because env isn't defined before matrix in the workflow run. I will apply the underscore instead of slash change, though.

bquan0:
I think the multistage docker build action always puts a '/' before the stage name as part of the package name, so we can't completely get rid of the slashes. Ex. `ghcr.io/bquan0/pyne_ubuntu_20.04_py3_hdf5/dagmc`.

gonuke:
I'm a little confused by why the test is suddenly failing.  I understand the error message, but can't tell what changes may have caused it.  I wonder if it's a complexity that has come from all the different docker images that have been built and rebuilt.  The solution would be to delete many/most of the docker images from your packages and try again?

bquan0:
I tried deleting the hdf5 packages and rerunning the workflow, but the same error happened again. I tried deleting all the packages that start with `pyne_` and still got the same error. I'll look into this more later today.

bquan0:
Fixed the error -- it was caused by one of the if statements in the BuildTest job still checking the hdf5 matrix variable against `/hdf5` instead of `_hdf5` on line 88.

gonuke (approved):

PR #1471, opened 2023-02-10
Fix multistage action image tag problem

bquan0:
## Description
- added an env variable that stores the tag of the docker image to be built by the multistage build action. This tag is based off the current branch's name
- added an if statement to set the env variable to "" if the branch is `develop`, the default branch
- added an output to the first job to send this env variable to the `BuildTest` job so it can use it to pull the correct image. (this is because you can't use env variables in the `image:` declaration of a job)
- made the multistage build actions NOT tag latest on the default branch (this is because in the `pushing_test_stable_img` job, we push the built images to other images with `latest` tags, so having the `latest` tag on the image when it's built would cause naming conflicts)
- switched the order of the pyne and openmc stage in `docker_publish.yml`
- made the pyne stage be based off the openmc stage

## Motivation and Context
Pull request #1470 was merged, but the Build Tests weren't able to run because they couldn't find the docker images. 

In the workflow file, the image names to be searched for have a tag of `refs_heads_multistage-docker-bk0`. This worked on the `multistage-docker` branch because the multistage build action applies a tag with the branch's name to the docker images it builds when the workflow is run on a branch that is not on the default branch of a repository. However, when the multistage build action is run on the default branch (in this case, `develop`), it tags the docker images with `latest` or nothing if the `tag-latest-on-default` option is made false. 

## Changes
This PR is a bug fix to allow the Build Tests to run properly in the main repository on the default `develop` branch. 

## Behavior
What is the current behavior? What is the new behavior?
The current behavior is that the Build Tests don't run at all because they can't find the correct images. The new expected behavior is that in non-default branches, the workflow will use images tagged with the branch name, and in the default branch `develop`, it will use images tagged with `latest` or nothing. 

However, the code in this PR currently isn't working as intended because the Build Tests are using the images tagged with latest and not tagged with the branch name, which means they aren't testing on the newly built images (newly built images are tagged with the branch name). 

The error has something to do with the fact that I'm not using `echo` correctly or I'm not referencing the env variables correctly to transfer the env variable for the tag to the `BuildTest` job via the output system from Github Actions.

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -38,8 +38,6 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - run: echo $DOCKER_IMAGE_TAG
-
       # build base python, moab, dagmc, openmc using multistage docker build action
       - uses: firehed/multistage-docker-build-action@v1
         with:
@@ -65,7 +63,6 @@ jobs:
       - id: output_tag
         run: |
           echo "image_tag=$DOCKER_IMAGE_TAG" >> $GITHUB_OUTPUT
-          echo "$DOCKER_IMAGE_TAG"
 
   # Downloads the images uploaded to ghcr in previous stages and runs pyne
   # tests to check that they work.
@@ -100,8 +97,6 @@ jobs:
           fi
           export ADD_FLAG="${ADD_FLAG} "
           echo "ADD_FLAG=${ADD_FLAG}" >> $GITHUB_ENV
-
-      - run: echo "$needs.multistage_image_build.outputs.image_tag"
 
       - name: Checkout repository
         uses: actions/checkout@v3
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -38,9 +38,10 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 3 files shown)


bquan0:
BuildTest is now using the correct images now (the images built by the workflow, which has a tag with the branch's name). It wasn't working before because I was referencing the env variables incorrectly (it's supposed to be just `$DOCKER_IMAGE_TAG` instead of `$env.DOCKER_IMAGE_TAG`) and because I wasn't echoing to the output correctly (it's supposed to be `echo "image_tag=$DOCKER_IMAGE_TAG"` instead of just `echo "$DOCKER_IMAGE_TAG"`, see [this example](https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs)). I also deleted the if statement that set `DOCKER_IMAGE_TAG` to be nothing if the workflow is run on a default branch because we realized that the multistage build action will tag the image with the branch name if `tag-latest-on-default` is turned off (see this [packages page](https://github.com/bquan0/pyne/pkgs/container/pyne_ubuntu_20.04_py3_hdf5%2Fbase_python), it has a tag for the `develop` branch)

gonuke:
LGTM - let's see how it works on the main repo

.github/workflows/docker_publish.yml
```diff
@@ -138,11 +145,11 @@ jobs:
       - name: Push Image to stable img
         uses: akhilerm/tag-push-action@v2
```
gonuke, line 146: @bquan0 - This may need to be fully resolved as `@v2.1.0`

PR #1472, opened 2023-02-13
fixed compatibility issue with Python 3.10 in endf.py

ahnaf-tahmid-chowdhury:
### Description
This pull request fixes a compatibility issue with Python 3.10 for the **endf.py** file in the PyNE library.

### Motivation and Context
The collections module in Python 3.10 changed the way the OrderedDict and Iterable classes are imported. This caused an ImportError in the endf.py file, preventing the PyNE library from functioning correctly.

### Changes
This PR updates the import statement in the endf.py file to correctly import the OrderedDict and Iterable classes in a way that is compatible with both Python 3.10 and earlier versions.

### Behavior
Before this change, the endf.py file would raise an ImportError in Python 3.10, preventing the PyNE library from being used. With this change, the library should now function correctly in Python 3.10.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,7 @@ v0.7.7
    * update where to look for DAGMC CMake files (#1446)
    * ensure NUC_DATA_PATH is a string (#1447)
    * update CI to Ubuntu 20.04 (#1449)
+   * fixed compatibility issue with Python 3.10 in endf.py 
 
 v0.7.6
 ======
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ Next Version
    * fixing typos in user warning message (#1456)
    * fix reorder error in member variables (#1466)
    * remove unusable C-linkage with std::vector (#1468)
+   * fixed compatibility issue with Python 3.10 in endf.py (#1472)
 
 v0.7.7
 ======
@@ -22,8 +23,7 @@ v0.7.7
    * turn off all fortran (spatial solvers and ENSDF processing) by default (#1444)
    * update where to look for DAGMC CMake files (#1446)
    * ensure NUC_DATA_PATH is a string (#1447)
-   * update CI to Ubuntu 20.04 (#1449)
-   * fixed compatibility issue with Python 3.10 in endf.py 
+   * update CI to Ubuntu 20.04 (#1449) 
 
 v0.7.6
 ======
--- a/pyne/utils.py
+++ b/pyne/utils.py
@@ -216,31 +216,20 @@ def remove(path):
 
 def str_to_unicode(s):
     """
-    This function convert a str from binary or unicode to str (unicode).
-    If it is a list of string, convert every element of the list.
-
-    Parameters:
-    -----------
-    s : str or list of str
-
-    Returns:
-    --------
-    s : text str or list of unicode str
+    Convert str to unicode str.
     """
-    if isinstance(s, str) or isinstance(s, bytes):
-        # it is a str, convert to text str
-        try:
-            s = s.decode("utf-8")
-        except:
-            pass
+    if isinstance(s, bytes):
+        return s.decode()
+    elif isinstance(s, str):
         return s
+    elif isinstance(s, list):
+        return [str_to_unicode(x) for x in s]
+    elif isinstance(s, set):
+        return set(x if isinstance(x, str) else x.decode() for x in list(s))
+    elif isinstance(s, tuple):
+        return tuple(str_to_unicode(x) for x in s)
     else:
-        for i, item in enumerate(s):
-            try:
-                s[i] = item.decode("utf-8")
-            except:
-                pass
-        return s
+        raise TypeError("Expected str, bytes, list, set, or tuple, but got %s" % type(s))
 
 
 def is_close(a, b, rel_tol=1e-9, abs_tol=0.0):
--- a/pyne/utils.py
+++ b/pyne/utils.py
@@ -161,7 +161,7 @@ def from_barns(xs, units):
 ### message functions ###
 #########################
 
-USE_COLOR = os.name is "posix"
+USE_COLOR = os.name == "posix"
 
 
 def message(s):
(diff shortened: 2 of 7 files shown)


CHANGELOG.rst
```diff
@@ -23,6 +23,7 @@ v0.7.7
    * update where to look for DAGMC CMake files (#1446)
    * ensure NUC_DATA_PATH is a string (#1447)
    * update CI to Ubuntu 20.04 (#1449)
+   * fixed compatibility issue with Python 3.10 in endf.py
```
gonuke: Please add the pull request number to the end of this line.

tests/test_transmuters.py
```diff
@@ -1,6 +1,6 @@
 """Transmuter tests"""
 from nose.tools import assert_equal, assert_almost_equal
-
+import nose
```
gonuke, line 3: Why do we need to import `nose` here?  We import from `nose.tools` above - did it cause an error?

tests/test_utils.py
```diff
@@ -219,15 +219,16 @@ def test_str_to_unicode():
     s = ["test1", "test2", b"test3"]
     exp_answer = ["test1", "test2", "test3"]
     assert_array_equal(utils.str_to_unicode(s), exp_answer)
-
     # set of str
     s = {"test1", "test2", b"test3"}
     exp_answer = {"test1", "test2", "test3"}
+    # Fix: assert_equal not working
```
gonuke: Not sure how to interpret this comment - do you plan to fix it?  If so, we can mark this PR as a draft or work in progres...

gonuke (changes requested):
Thanks so much for your contribution @ahnaf-tahmid-chowdhury!  It's always great to get contributions from new people in our community.

I've launched the tests to see how these changes impact our current version of python (3.8) and ensure we're compatible across them all.

I also have a few comments/questions here.

tests/test_utils.py
ahnaf-tahmid-chowdhury: Yes, I will try. I have found **assert_equal** only works in the string object. In this case, the values are in the dictionary and tuple format.

ahnaf-tahmid-chowdhury (commented):
pull request assigned #1472

tests/test_transmuters.py
ahnaf-tahmid-chowdhury: The nose module is used to run the test function `test_transmuters_cram()` as a test suite. The `nose.tools` module contains helper functions that can be used to write tests with `nose`, but `nose` is the actual test runner that executes the test functions and reports the results.

The `if __name__ == "__main__":` block at the end of the script is used to run the test suite if the script is being executed directly, rather than being imported as a module. In this case, `nose.runmodule()` is called to run the `test_transmuters_cram()` function as a test suite using nose.

gonuke: Thanks. We must not test in this way so didn't see it fail in the past.

tests/test_utils.py
ahnaf-tahmid-chowdhury: I have fixed the issue here [3e57c2a](https://github.com/pyne/pyne/pull/1472/commits/3e57c2ae4c0dee7e122067e6c3b836cb71f87cbb)

ahnaf-tahmid-chowdhury:
Dear @gonuke, please let me know if you've encountered any issues while running the tests.

pyne/utils.py
```diff
@@ -216,31 +216,20 @@ def remove(path):
...
-        except:
-            pass
+    if isinstance(s, bytes):
+        return s.decode()
+    elif isinstance(s, str):
         return s
+    elif isinstance(s, list):
+        return [str_to_unicode(x) for x in s]
+    elif isinstance(s, set):
+        return set(x if isinstance(x, str) else x.decode() for x in list(s))
```
gonuke: Why not invoke this function recursively as you've done in the other cases?

pyne/utils.py
```diff
@@ -216,31 +216,20 @@ def remove(path):
...
-    If it is a list of string, convert every element of the list.
-
-    Parameters:
-    -----------
-    s : str or list of str
-
-    Returns:
-    --------
-    s : text str or list of unicode str
+    Convert str to unicode str.
```
gonuke, line 219: We generally require more documentation in the docstring

tests/test_utils.py
```diff
@@ -219,16 +218,14 @@ def test_str_to_unicode():
     s = ["test1", "test2", b"test3"]
     exp_answer = ["test1", "test2", "test3"]
     assert_array_equal(utils.str_to_unicode(s), exp_answer)
-
     # set of str
     s = {"test1", "test2", b"test3"}
     exp_answer = {"test1", "test2", "test3"}
-    assert (utils.str_to_unicode(s), exp_answer)
-
+    assert_equal(utils.str_to_unicode(s), exp_answer)
```
gonuke, line 228: Putting back the blank line for readability 
```suggestion
    assert_equal(utils.str_to_unicode(s), exp_answer)

```

gonuke (changes requested):
Thanks for following this through, a few small comments, but all the tests are passing

pyne/utils.py
ahnaf-tahmid-chowdhury: Added in [6f281cd](https://github.com/pyne/pyne/pull/1472/commits/6f281cd77d546cb21dd31fdf34fe1cba39f7af58)
This docstring provides a clear explanation of what the function does, what arguments it takes, what it returns, and what exceptions it may raise. It also provides examples of how the function works in different cases.

pyne/utils.py
ahnaf-tahmid-chowdhury: Fixed in [6f281cd](https://github.com/pyne/pyne/pull/1472/commits/6f281cd77d546cb21dd31fdf34fe1cba39f7af58)
I have found that whenever I convert a list to a set, the values get rearranged, which can lead to incorrect results. However, I have found that using curly braces `{}` instead of `set()` fixes this issue.

gonuke (approved):

gonuke:
Thanks @ahnaf-tahmid-chowdhury for seeing this through!

Issue #1474, opened 2023-02-17
Change docker images used in CI


gonuke:
With the success of #1473 (and predecessors) we have new docker images based on our streamlined build system.  We need to update our standard CI testing to use these images instead of the older images.

gonuke:
resolved in #1475

PR #1475, opened 2023-02-22
Update build_test.yml to use new images

bquan0:
## Description
I updated `build_test.yml` to use the new images we build from `docker_publish.yml`. I also changed the `matrix` and the if statements in `build_test.yml` to match the use of the new images.

## Motivation and Context
Requested in #1474. This is a next step after modifying `docker_publish.yml` to build new images using the multistage docker build action. 

## Changes
This is a small update to make sure the workflows that build and test images are testing the ones we most recently built. 

## Behavior
Old behavior was that `build_test.yml` was testing the images built by the pre #1470 `docker_publish.yml` workflow. New behavior is that it tests the images built by the new `docker_publish.yml` workflow. 

## Additional Information
I put this PR under the same entry as #1470, #1471, #1473 in `CHANGELOG.rst` since it seemed like another part of updating the workflows to use the multistage docker build action.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -26,57 +26,35 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        pymoab : [
-          _pymoab,
-          ]
-        dagmc : [
-          '',
-          _dagmc
-        ]
-        openmc : [
-          '',
-        ]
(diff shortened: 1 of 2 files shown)


gonuke (approved):
This looks good @bquan0 - nice to have an easy PR for a change :)

Issue #1462, opened 2022-11-19
Move standard installation script into main PyNE repo for regular testing


gonuke:
The PyNE team has historically maintained a separate repository with installation scripts for different platforms: [pyne/install_scripts](https://github.com/pyne/install_scripts).

The scripts in this repo frequently fall out of sync with the state of the main code repository, causing frustration for users.  It will be better to provide such a script as part of this repository, and add an action that forces that/those scripts to be tested with each PR to ensure that those scripts are updated as appropriate with each code addition.

gonuke:
While I had a conversation about this with @ngranda, it might make more sense for @bquan0 to take this on.

I think the best way to make an install script is to use the sequence of commands that are in the current `Dockerfile` and implement them as a stand-alone bash script.

Then we need to add a workflow that runs this script.  I think there may be some open questions about which version of the code should be tested with this script.  Perhaps the script should only be updated with a release?

ahnaf-tahmid-chowdhury:
Certainly, it is possible to use both Docker and a Python environment such as venv in a development workflow. In fact, using both can provide the best of both worlds in terms of flexibility, isolation and portability.

Docker allows for creating a standardized and portable runtime environment for an application, making it easy to move between different environments and ensuring consistency. Meanwhile, a Python environment such as venv can provide a lightweight and isolated environment for testing and developing code, avoiding conflicts with other packages and ensuring the application runs as intended. By using both, it is possible to take advantage of the benefits of each approach to create a robust and effective development workflow that meets the needs of the project and the team.

When considering the use of bash scripting for installation, it may be beneficial to provide users with the ability to choose which modules to install. This can help ensure that only the necessary modules are installed, streamlining the process and potentially saving disk space.

That being said, it is important to note that DAGMC supports Geant4. Therefore, it may also be worthwhile to include the option for users to include Geant4 in the same environment. This would allow users to benefit from the full functionality of DAGMC and Geant4 without the need for separate installations.

By providing users with flexibility in the installation process and ensuring that all necessary dependencies are included, we can create a more streamlined and efficient workflow for the development and use of DAGMC.

gonuke:
I think it is reasonable to support a variety of installation script variations, but we should consider these high-level philosophical points:
1. Will too many variations be confusing to potential users? Can we provide adequate documentation to help users find what they need?
2. There is maintenance burden with updating multiple installation scripts, so we should find the right balance on this front, as well.
[...]

PR #1477, opened 2023-03-03
Move installation script to main repo

bquan0:
## Description
I added the `ubuntu.sh` bash install script from the [install_scripts repo](https://github.com/pyne/install_scripts) to the `scripts` folder of this repo. I also edited it slightly to match the commands in `ubuntu_22.04-dev.dockerfile`. 

## Motivation and Context
This change was requested in #1462.

## Changes
Along with adding the install script and updating it to match the dockerfile, I also updated `ubuntu_22.04-dev.dockerfile` to not check for python versions (since we only use version 3.x now) and updated it from ubuntu 20.04 to 22.04, which were two changes requested by @gonuke.

## Behavior
The current behavior is that the install script is in a separate repo and isn't getting tested/updated when the code in this repo is updated. The new behavior should be that the install script is in this repo along with the other code and it will be tested and updated whenever code in this repo is updated. 

## Other Information
There are probably some things yet to be updated/polished in `ubuntu.sh`. When I was editing it, I only updated parts where `ubuntu.sh` had something different or missing compared to the dockerfile. I left the commands that were only in `ubuntu.sh` but not in the dockerfile intact . 
Also, I'm not sure how to go about testing this bash install script.

### DIFF
--- a/docker/ubuntu_22.04-dev.dockerfile
+++ b/docker/ubuntu_22.04-dev.dockerfile
@@ -1,7 +1,7 @@
 ARG py_version=3.8
 ARG build_hdf5="NO"
 
-FROM ubuntu:20.04 AS base_python
+FROM ubuntu:22.04 AS base_python
 
 # Ubuntu Setup
 ENV TZ=America/Chicago
--- a/docker/ubuntu_22.04-dev.dockerfile
+++ b/docker/ubuntu_22.04-dev.dockerfile
@@ -100,7 +100,7 @@ RUN export PYMOAB_FLAG="-DENABLE_PYMOAB=ON"; \
 # put MOAB on the path
 ENV LD_LIBRARY_PATH $HOME/opt/moab/lib:$LD_LIBRARY_PATH
 ENV LIBRARY_PATH $HOME/opt/moab/lib:$LIBRARY_PATH
-ENV PYTHONPATH=$HOME/opt/moab/lib/python3.10/site-packages/
+# ENV PYTHONPATH=$HOME/opt/moab/lib/python3.10/site-packages/
 
 FROM moab AS dagmc
 # build/install DAGMC
--- a/docker/ubuntu_22.04-dev.dockerfile
+++ b/docker/ubuntu_22.04-dev.dockerfile
@@ -69,7 +69,7 @@ ENV INSTALL_PATH=$HOME/opt/moab
 RUN export PYMOAB_FLAG="-DENABLE_PYMOAB=ON"; \
     echo $PYMOAB_FLAG ;\
     export MOAB_HDF5_ARGS=""; \
-    if [ "$build_hdf5" != "NO" ] ; \ 
+    if [ "$build_hdf5" != "NO" ] ; \
     then \
             export MOAB_HDF5_ARGS="-DHDF5_ROOT=$HDF5_INSTALL_PATH"; \
     fi \
@@ -90,6 +90,9 @@ RUN export PYMOAB_FLAG="-DENABLE_PYMOAB=ON"; \
             -DENABLE_BLASLAPACK=OFF \
             -DENABLE_FORTRAN=OFF \
     && make -j 3 \
+    && cd pymoab \
+    && pip install . --user \
+    && cd ..
     && make install \
     && cd .. \
     && rm -rf moab ;
@@ -129,7 +132,7 @@ RUN if [ "$build_hdf5" != "NO" ]; then \
     git clone https://github.com/openmc-dev/openmc.git $HOME/opt/openmc \
     && cd  $HOME/opt/openmc \
     && git checkout tags/v0.13.0 \
-    && pip install . 
+    && pip install .
 
 # Build/Install PyNE
 FROM openmc AS pyne
@@ -145,7 +148,7 @@ RUN export PYNE_HDF5_ARGS="" ;\
     && python setup.py install --user \
                                 --moab $HOME/opt/moab --dagmc $HOME/opt/dagmc \
                                 $PYNE_HDF5_ARGS \
-                                --clean -j 3; 
+                                --clean -j 3;
 ENV PATH $HOME/.local/bin:$PATH
 RUN if [ "$build_pyne" = "YES" ]; then \
         cd $HOME \
--- a/docker/ubuntu_22.04-dev.dockerfile
+++ b/docker/ubuntu_22.04-dev.dockerfile
@@ -92,7 +92,7 @@ RUN export PYMOAB_FLAG="-DENABLE_PYMOAB=ON"; \
     && make -j 3 \
     && cd pymoab \
     && pip install . --user \
-    && cd ..
+    && cd .. \
     && make install \
     && cd .. \
     && rm -rf moab ;
--- a/docker/ubuntu_22.04-dev.dockerfile
+++ b/docker/ubuntu_22.04-dev.dockerfile
@@ -10,11 +10,10 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 ARG py_version
 
 ENV HOME /root
-RUN export PY_SUFIX=${py_version%.?}; \
-    apt-get update \
+RUN apt-get update \
     && apt-get install -y --fix-missing \
             software-properties-common \
-            python${PY_SUFIX}-pip \
+            python3-pip \
             wget \
             build-essential \
             git \
--- a/docker/ubuntu_22.04-dev.dockerfile
+++ b/docker/ubuntu_22.04-dev.dockerfile
@@ -98,7 +98,7 @@ RUN export PYMOAB_FLAG="-DENABLE_PYMOAB=ON"; \
 # put MOAB on the path
 ENV LD_LIBRARY_PATH $HOME/opt/moab/lib:$LD_LIBRARY_PATH
 ENV LIBRARY_PATH $HOME/opt/moab/lib:$LIBRARY_PATH
-ENV PYTHONPATH=$HOME/opt/moab/lib/python3.10/site-packages/
+ENV PYTHONPATH=$HOME/opt/moab/lib/python3.10/site-packages/:$PYTHONPATH
 
 FROM moab AS dagmc
 # build/install DAGMC
--- a/docker/ubuntu_22.04-dev.dockerfile
+++ b/docker/ubuntu_22.04-dev.dockerfile
@@ -1,7 +1,8 @@
 ARG py_version=3.8
 ARG build_hdf5="NO"
+ARG ubuntu_version=22.04
 
-FROM ubuntu:22.04 AS base_python
+FROM ubuntu:${ubuntu_version} AS base_python
 
 # Ubuntu Setup
 ENV TZ=America/Chicago
--- a/docker/ubuntu_22.04-dev.dockerfile
+++ b/docker/ubuntu_22.04-dev.dockerfile
@@ -69,7 +69,7 @@ ENV INSTALL_PATH=$HOME/opt/moab
 RUN export PYMOAB_FLAG="-DENABLE_PYMOAB=ON"; \
     echo $PYMOAB_FLAG ;\
     export MOAB_HDF5_ARGS=""; \
-    if [ "$build_hdf5" != "NO" ] ; \ 
+    if [ "$build_hdf5" != "NO" ] ; \
     then \
             export MOAB_HDF5_ARGS="-DHDF5_ROOT=$HDF5_INSTALL_PATH"; \
     fi \
@@ -90,6 +90,9 @@ RUN export PYMOAB_FLAG="-DENABLE_PYMOAB=ON"; \
             -DENABLE_BLASLAPACK=OFF \
             -DENABLE_FORTRAN=OFF \
     && make -j 3 \
+    && cd pymoab \
+    && pip install . \
+    && cd .. \
     && make install \
     && cd .. \
     && rm -rf moab ;
@@ -129,7 +132,7 @@ RUN if [ "$build_hdf5" != "NO" ]; then \
     git clone https://github.com/openmc-dev/openmc.git $HOME/opt/openmc \
     && cd  $HOME/opt/openmc \
     && git checkout tags/v0.13.0 \
-    && pip install . 
+    && pip install .
 
 # Build/Install PyNE
 FROM openmc AS pyne
@@ -145,7 +148,7 @@ RUN export PYNE_HDF5_ARGS="" ;\
     && python setup.py install --user \
                                 --moab $HOME/opt/moab --dagmc $HOME/opt/dagmc \
                                 $PYNE_HDF5_ARGS \
-                                --clean -j 3; 
+                                --clean -j 3;
 ENV PATH $HOME/.local/bin:$PATH
 RUN if [ "$build_pyne" = "YES" ]; then \
         cd $HOME \
--- a/docker/ubuntu_22.04-dev.dockerfile
+++ b/docker/ubuntu_22.04-dev.dockerfile
@@ -1,4 +1,3 @@
-ARG py_version=3.8
 ARG build_hdf5="NO"
 ARG ubuntu_version=22.04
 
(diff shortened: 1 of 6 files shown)


docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -1,22 +1,19 @@
 ARG py_version=3.8
```
ngranda: Is there a reason why Python 3.8 and not 3.10 which is the default on Ubuntu 22.04?

gonuke: Good catch @ngranda - I think getting rid of this build arg entirely probably makes sense

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -105,7 +96,7 @@ RUN export PYMOAB_FLAG="-DENABLE_PYMOAB=ON"; \
 # put MOAB on the path
 ENV LD_LIBRARY_PATH $HOME/opt/moab/lib:$LD_LIBRARY_PATH
 ENV LIBRARY_PATH $HOME/opt/moab/lib:$LIBRARY_PATH
-ENV PYTHONPATH=$HOME/opt/moab/lib/python${py_version}/site-packages/
+ENV PYTHONPATH=$HOME/opt/moab/lib/python3.10/site-packages/
```
gonuke: This isn't robust to the introduction of newer versions of python, but I'm not 100% sure what the solution is.

gonuke (changes requested):
Thanks @bquan0 .

Can you also add a workflow that runs the install script on a clean Ubuntu 22.04 runner? i.e. not one of our docker images

gonuke:
Thanks for adding the new workflow but I don’t see it being triggered anywhere

bquan0:
I noticed that in the `Building PyNE` step of the `BuildTest` job, the `PYTHONPATH` was specified with python3.8 instead of python3.10. I tried changing it to see if it would fix the `ModuleNotFoundError: No module named 'pymoab'` errors I've been getting on the `BuildTest` parts of the [workflow run](https://github.com/bquan0/pyne/actions/runs/4769047799), but it didn't work.

pshriwise:
Regarding our attempt at installing PyMOAB to the user location today: I found that the reason this doesn't work is because the moab & pymoab installation happens during the Dockerfile build when `HOME=/root`. Later, when the action uses the container `HOME=/github/home` and the pymoab installation we did can't be found in `$HOME/.local`, one of the default search paths for Python packages. 

I'm not sure what the right fix for this is. There's an issue on it [here](https://github.com/actions/runner/issues/863) without any clear solution that I could see.

FWIW I diagnosed this using https://github.com/marketplace/actions/debugging-with-tmate, a really handy tool that lets you ssh into the CI machine at any point in the build.

gonuke:
> Regarding our attempt at installing PyMOAB to the user location today: I found that the reason this doesn't work is because the moab & pymoab installation happens during the Dockerfile build when `HOME=/root`. Later, when the action uses the container `HOME=/github/home` and the pymoab installation we did can't be found in `$HOME/.local`, one of the default search paths for Python packages.
> 
> I'm not sure what the right fix for this is. There's an issue on it [here](https://github.com/actions/runner/issues/863) without any clear solution that I could see.
> 
> FWIW I diagnosed this using https://github.com/marketplace/actions/debugging-with-tmate, a really handy tool that lets you ssh into the CI machine at any point in the build.

This is only a problem because we are installing with `--user` right?  I guess for CI we don't need to do that?

pshriwise:
True. We could do a system install instead. I'll try that.

gonuke:
We could also move all testing to be part of a docker build like we are aiming towards for DAGMC

pshriwise:
Also true. I'll try this first to confirm the plan for the PyMOAB pip install is viable.

pshriwise:
Huzzah! https://github.com/pshriwise/pyne/actions/runs/4840438166

pshriwise:
I'll submit another PR into @bquan0's branch after a little history cleanup

scripts/ubuntu.sh
```diff
@@ -0,0 +1,220 @@
...
+cd build
+
+# cmake, build and install
+cmake ../ -DENABLE_PYMOAB=ON \
+          -DCMAKE_INSTALL_PREFIX=${install_dir}/moab \
+          -DENABLE_HDF5=ON -DHDF5_ROOT=${hdf5_libdir} \
+          -DBUILD_SHARED_LIBS=ON \
+          -DENABLE_BLASLAPACK=OFF \
+          -DENABLE_FORTRAN=OFF    
+make -j 3
```
gonuke, line 113: I think we should add 
```
cd pymoab && pip3 install .
```
like we did in the Dockerfile

scripts/ubuntu.sh
```diff
@@ -0,0 +1,220 @@
...
+# Adding .local//bin to $PATH
+export PATH="${HOME}/.local/bin:$PATH"
+
+# Make Pyne Nuclear Data
+cd  # cd without argument will take you back to your $HOME directory
+nuc_data_make
+
+# Run tests
+cd ${install_dir}/pyne/pyne-repo/tests
+./travis-run-tests.sh
```
gonuke: Isn't this 
```suggestion
./ci-run-tests.sh
```
Now?

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -98,14 +90,17 @@ RUN export PYMOAB_FLAG="-DENABLE_PYMOAB=ON"; \
...
+    && cd .. \
     && make install \
     && cd .. \
     && rm -rf moab ;
 
 # put MOAB on the path
 ENV LD_LIBRARY_PATH $HOME/opt/moab/lib:$LD_LIBRARY_PATH
 ENV LIBRARY_PATH $HOME/opt/moab/lib:$LIBRARY_PATH
-ENV PYTHONPATH=$HOME/opt/moab/lib/python${py_version}/site-packages/
+ENV PYTHONPATH=$HOME/opt/moab/lib/python3.10/site-packages/
```
gonuke: Might not need this anymore if we install pymoab using pip

gonuke (changes requested):
A few more changes related to pip installing moab

scripts/ubuntu.sh
```diff
@@ -0,0 +1,223 @@
...
+# cmake, build and install
+cmake ../ -DENABLE_PYMOAB=ON \
+          -DCMAKE_INSTALL_PREFIX=${install_dir}/moab \
+          -DENABLE_HDF5=ON -DHDF5_ROOT=${hdf5_libdir} \
+          -DBUILD_SHARED_LIBS=ON \
+          -DENABLE_BLASLAPACK=OFF \
+          -DENABLE_FORTRAN=OFF    
+make -j 3 
+cd pymoab
+pip install .
```
gonuke: We need to be careful about how users will run this, i.e. not as `root`.  Should install this as `--user` and set `PYTHONPATH` accordingly

scripts/ubuntu.sh
```diff
@@ -0,0 +1,223 @@
...
+export PATH="${install_dir}/dagmc/bin:$PATH"
+
+####################
+### OpenMC API #####
+####################
+
+cd ${install_dir}
+git clone https://github.com/openmc-dev/openmc.git
+cd openmc
+git checkout tags/v0.13.0
```
gonuke, line 170: Can we clone only this tag, as we do with Moab and DAGMC above?

scripts/ubuntu.sh
```diff
@@ -0,0 +1,223 @@
...
+mkdir bld
+cd bld
+cmake .. -DCMAKE_INSTALL_PREFIX=$HOME/.local
+make
+make install
+cd ..
+pip3 install .
```
gonuke: I think we've aliased `pip` to `pip3` already and used just `pip` above. Let's be consistent

scripts/ubuntu.sh
```diff
@@ -0,0 +1,223 @@
...
+cd ${install_dir}
+check_repo pyne
+mkdir -p pyne
+cd pyne
+# clone and version
+git clone -b develop --single-branch https://github.com/pyne/pyne.git pyne-repo
+cd pyne-repo
+if [ $1 == 'stable' ] ; then
+  TAG=$(git describe --abbrev=0 --tags)
+  git checkout tags/`echo ${TAG}` -b `echo ${TAG}`
```
gonuke, line 191: Can we do this if we've only cloned a single branch?

gonuke (changes requested):
A few suggestions for users running not as root.  I wonder if we should update the workflow that tests the install script to run as non-root?

scripts/ubuntu.sh
bquan0: I don't think lines 176-178 are being run. In [this workflow run](https://github.com/bquan0/pyne/actions/runs/4960351409/jobs/8875715298), line 1034 corresponds to the `git clone` on line 173 in `ubuntu.sh`, and the lines that follow in the workflow run don't seem to have any tags echoed.

scripts/ubuntu.sh
bquan0: Should I apply this change for all instances of `pip install .`?

gonuke: Probably makes sense to do them all since that's what a user may do?

scripts/ubuntu.sh
gonuke: I think they are being run but the condition is never true.  If it were ever true, it would probably fail.  Since this is a sample script for users to use to install, we should choose a good practice.

I think this was to allow people to run this script to install a guaranteed stable version rather than the develop branch.  That's probably a good thing to preserve, and therefore we might want to test both versions

bquan0: After implementing the matrix idea for testing both the develop branch and the stable branch, it seems like it is possible to switch to the stable branch even though only a single branch was cloned. In line 924 of [this workflow run](https://github.com/bquan0/pyne/actions/runs/5150609113/jobs/9274941733) with a command line arg of `stable`, it says `Switched to a new branch '0.7.7'`. This line is not there in [the workflow run](https://github.com/bquan0/pyne/actions/runs/5150609113/jobs/9274941603) with a different command line arg (see line 929)

bquan0:
After looking at the workflow runs, it seems like the install script has been running into errors such as `./ubuntu.sh: line 191: nuc_data_make: command not found` and `./ci-run-tests.sh: line 11: pytest: command not found` since [the very first successful workflow run](https://github.com/bquan0/pyne/actions/runs/4759177570/jobs/8458152347), which probably prevent PyNE from being installed (and is also probably why the workflow runs only take ~5 mins)

The new workflow runs run into an error of not being able to find the moab repo (line 302 of [this workflow run](https://github.com/bquan0/pyne/actions/runs/5150609113/jobs/9274941603))

scripts/ubuntu.sh
```diff
@@ -0,0 +1,222 @@
...
+                  libhdf5-dev \
+                  hdf5-tools"
+
+# list of python package required for PyNE and its depedencies (installed using pip3 python package manager)
+pip_package_list="numpy==1.23 \
+                  scipy \
+                  cython \
+                  nose \
+                  tables \
+                  matplotlib \
```
gonuke, line 27: need to add `pytest` here - see the Dockerfile for guidance

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -69,15 +62,14 @@ ENV LD_LIBRARY_PATH $HDF5_INSTALL_PATH/lib:$LD_LIBRARY_PATH
 ENV LIBRARY_PATH $HDF5_INSTALL_PATH/lib:$LIBRARY_PATH
 
 FROM base_python AS moab
-ARG py_version
 ARG build_hdf5
 ENV INSTALL_PATH=$HOME/opt/moab
 
 # build MOAB
 RUN export PYMOAB_FLAG="-DENABLE_PYMOAB=ON"; \
```
gonuke: If we hardcode this flag to be ON, then why not just remove the use of a variable and hard code this in the cmake command below?

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -98,14 +90,17 @@ RUN export PYMOAB_FLAG="-DENABLE_PYMOAB=ON"; \
...
+    && cd .. \
     && make install \
     && cd .. \
     && rm -rf moab ;
 
 # put MOAB on the path
 ENV LD_LIBRARY_PATH $HOME/opt/moab/lib:$LD_LIBRARY_PATH
 ENV LIBRARY_PATH $HOME/opt/moab/lib:$LIBRARY_PATH
-ENV PYTHONPATH=$HOME/opt/moab/lib/python${py_version}/site-packages/
+# ENV PYTHONPATH=$HOME/opt/moab/lib/python3.10/site-packages/
```
gonuke: remove this if we don't need it anymore, rather than comment

scripts/ubuntu.sh
```diff
@@ -0,0 +1,222 @@
...
+# Adding MOAB/lib to $LD_LIBRARY_PATH and $LIBRARY_PATH
+export LD_LIBRARY_PATH="${install_dir}/moab/lib:$LD_LIBRARY_PATH"
+
+# Adding pymoab to $PYTHONPATH
+PYTHON_VERSION=$(python -c 'import sys; print(sys.version.split('')[0][0:3])')
+if [ -z $PYTHONPATH ]; then
+  export PYTHONPATH="$install_dir/moab/lib/python${PYTHON_VERSION}/site-packages"
+else
+  export PYTHONPATH="$install_dir/moab/lib/python${PYTHON_VERSION}/site-packages:$PYTHONPATH"
+fi
```
gonuke: Might not need this if we install with `pip` - see how this works in the dockerfile?

gonuke (changes requested):
I noted a few things, but most are related to using the working(?) Dockerfile to guide the updates of the install script

gonuke:
> this workflow run

definitely strange that it could not find the repo - I wonder if that was a transient error that would be fixed by running the workflow again?

gonuke:
It looks like this is still not successful.  I recommend that you do a line-by-line comparison of the Dockerfile with the ubuntu.sh script.  We know that the former is working, so it is a good guide for creating this script.

The script does historically have the goal of being a kind of tutorial for building pyne as well, but it might be useful to first just get it working and then we can worry about how to embellish for tutorial purposes.

bquan0:
I think after I added the if statements for hdf5 that were in the dockerfile but not `ubuntu.sh`, it seems to be working. I also added another matrix variable to make the script build with hdf5 and without hdf5. The [workflow run](https://github.com/bquan0/pyne/actions/runs/5193706150/jobs/9364586964) lasts longer now and I think the previous errors I mentioned have been fixed.

gonuke:
Definitely much better!!  It's building and installing everything correctly, but it doesn't find the script to run the tests.  It's the second last line in the CI log for the "run install script" job. I'm not quite sure why, except that you may be in the wrong directory? Or the file is not executable?

bquan0:
The tests run in the workflows where we use the `develop` branch instead of the stable branch. I ran the git commands for pyne (lines 187 to 192 in `ubuntu.sh`) and found that when the code in the stable if statement ran, we got switched to the branch `0.7.7`, which has a file called `travis-run-tests.sh` instead of the file `ci-run-tests.sh` in the `tests` directory. I think this got changed with [Nancy's PR](https://github.com/pyne/pyne/pull/1478#event-9007835711).

For now, I'm just going to use an if statement to make it run `./travis-run-tests.sh` if the stable branch is used and `./ci-run-tests.sh` if the develop branch is used. After a new stable version is released with the new test script name, we can omit this part  and just do `./ci-run-tests.sh`. 

It looks like the tests are now running on [the workflows that use the stable branch](https://github.com/bquan0/pyne/actions/runs/5213933014/jobs/9409572807), but they fail at `test_ace.py`. This is probably due to the fact that we download python 3.10 in the install script and the old tests in `travis-run-test.sh` are designed for an older version of python. 

After a quick discussion at the software meeting on 6/9, it seems like our options are to either
- stop supporting the stable branch and only worry about the develop branch
- use if statements to download an older version of python when we run it on the stable branch.

gonuke:
Good catch on the conditional for stable vs develop.  I think this is all working now so I will merge.

We'll probably want to do a release and update stable (and then update this again).

gonuke (approved):

Issue #1453, opened 2022-05-26
Nose not supported with python 3.10


ngranda:
**Describe the Bug**<br/>
Nosetest is deprecated and no longer supported with python 3.10. Update to pytest might be needed 

**To Reproduce**<br/>
Building PyNE on Ubuntu 22.04 with python 3.10 cannot run tests. The problem can be reproduced with the docker file [here](https://github.com/ngranda/install_scripts/blob/ubuntu22.04/ubuntu_22.04-stable.dockerfile) 

**Expected Behavior**<br/>
Tests should pass

**Please complete the following information regarding your system:**<br/>
 - OS: Ubuntu22.04
 - PyNE Version: 0.7.7
 - Versions of dependencies installed with: hdf5 1.10.7, python 3.10, nose 1.3.7, moab 5.3

ahnaf-tahmid-chowdhury:
From Python v3.10, `Callable` object is removed from `collections` class. It is now moved to `collections.abc` class.

You may need to replace this in `suite.py` and `case.py`:

    import collections

To:

     import collections.abc

gonuke:
Resolved in #1478

PR #1478, opened 2023-03-06
Move from nose tests to pytest

ngranda:
## Description
This PR moves the current tests from nose-based tests to pytests

## Motivation and Context
This PR addresses issue #1453 

## Other Information
This PR is currently in progress. Several things need addressing before ready for merge

- [x] Many tests use the keyword `yield`, but pytest no longer supports `yield` tests and causes tests to be skipped. The following files use the `yield` keyword:
  - `test_bins.py`
  - `test_cram.py`
  - `test_data.py`
  - `test_decay.py`
  - `test_enrichment.py`
  - `test_mcnp.py`
  - `test_nucname.py`
  - `test_rxname.py`
- [x] Some tests currently use a teardown decorator to remove files. These lines are currently commented out, and files are not being removed. These files are presently affected 
  - `test_source_sampling.py `
  - `test_mesh.py`
- [x] test.dagmc.py currently fails three tests potentially having to do with how `pytest.raise()` is set up
  - `test_failures`
  - `test_descritize_non_square`
  - `test_discretize_geom_centers`


gonuke:
You may want to update the `build-test.yml` workflow by installing `pytest` (maybe with `pip`?).  Once we know it all works, we can have @bquan0 add it directly to the images used for CI.

tests/test_ahot.py
```diff
@@ -64,4 +62,4 @@ def test_sct_step_alternating():
 
 
 if __name__ == "__main__":
-    nose.runmodule()
+    pytest.main()
```
gonuke: This is not a like-for-like replacement and the functionality may not be necessary???

Probably just remove the whole block and don't allow these to be run directly as scripts.

tests/ci-run-tests.sh
```diff
@@ -6,12 +6,12 @@ for x in $(ls test*.py */test*.py); do
...
   if [ $x == "ensdf_processing.py" ]; then
-    nosetests test_ensdf_processing.py --process-timeout=120
+    pytest test_ensdf_processing.py --process-timeout=120
   else
-    nosetests "$x"
+    pytest "$x"
   fi
   status=$?
-  if [ $status -ne 0 ]; then 
+  if [ $status -ne 0 ] && [ $status -ne 5 ]; then
```
gonuke, line 14: Can you add a comment on this new status test?

tests/test_activation_responses.py
```diff
@@ -236,16 +239,16 @@ def _activation_responses_test_step1(activation_responses_run_dir):
     os.remove(step1_file)
     os.remove(dst)
 
-    assert_equal(f1, True)
-    assert_equal(f2, True)
-    assert_equal(f3, True)
+    assert f1 == True
+    assert f2 == True
+    assert f3 == True
```
gonuke:
```suggestion
    assert f1
    assert f2
    assert f3
```

tests/test_alara.py
```diff
@@ -253,28 +250,28 @@ def test_photon_source_to_hdf5():
...
+            assert tokens[1].strip() == row["time"].decode()
             assert_array_equal(np.array(tokens[2:], dtype=np.float64), row["phtn_src"])
             old = tokens[0]
 
     if os.path.isfile(filename + ".h5"):
         os.remove(filename + ".h5")
 
     # wrong nuc option raise test
-    assert_raises(ValueError, photon_source_to_hdf5, filename, "test")
+    pytest.raises(ValueError, photon_source_to_hdf5, filename, "test")
```
gonuke, line 263: Does this syntax work?  I thought we found it needed to be a different syntax?

tests/test_alara.py
```diff
@@ -840,18 +837,18 @@ def test_is_data():
         "Zone #1: zone_0",
     ]
     for line in lines:
-        assert_equal(_is_data(line), False)
+        assert _is_data(line) == False
```
gonuke:
```suggestion
        assert !_is_data(line)
```

tests/test_alara.py
```diff
@@ -840,18 +837,18 @@ def test_is_data():
...
 
     lines = [
         "h-3     9.5258e-18  9.5258e-18  9.5251e-18  9.5214e-18",
         "n-16    3.1588e-09  0.0000e+00  0.0000e+00  0.0000e+00",
         "total   1.4597e-08  9.3939e-09  5.3291e-10  8.3915e-11",
     ]
     for line in lines:
-        assert_equal(_is_data(line), True)
+        assert _is_data(line) == True
```
gonuke:
```suggestion
        assert _is_data(line)
```

tests/test_alara.py
```diff
@@ -840,18 +837,18 @@ def test_is_data():
...
     # test special line
     line = "===========\n"
-    assert_equal(_is_data(line), False)
+    assert _is_data(line) == False
```
gonuke:
```suggestion
    assert !_is_data(line)
```

tests/test_cccc.py
```diff
@@ -66,25 +65,25 @@ def test_methods(self):
...
+    assert rt.ndim == 3
+    assert rt.ngroup == 4
+    assert rt.ninti == 4
+    assert rt.nintj == 4
+    assert rt.nintk == 4
+    assert rt.niter == 23
+    assert rt.effk == 0.892322838306427
+    assert rt.power == 1.0
+    assert rt.nblok == 1
+    assert rt.adjoint == False
```
gonuke:
```suggestion
    assert !rt.adjoint
```

tests/test_cccc.py
```diff
@@ -245,25 +244,25 @@ def test_rt_flux_1D():
...
     m = Mesh(structured=True, structured_coords=structured_coords)
-    with assert_raises(ValueError):
+    with pytest.raises(ValueError):
         rt.to_mesh(m, "flux")
 
 
 def test_atflux_adjoint():
     at = Atflux("files_test_cccc/atflux_3D")
-    assert_equal(at.adjoint, True)
+    assert at.adjoint == True
```
gonuke:
```suggestion
    assert at.adjoint
```

tests/test_cram.py
```diff
@@ -2,74 +2,75 @@
...
-
-from nose.tools import assert_true
+import pytest
 
 from pyne import cram
 
-DTYPES = ["f8", np.complex128]
+@pytest.fixture()
+def dtype_array():
+    return ["f8", np.complex128]
```
gonuke: I wonder if we can do this with parametrize instead?  This implementation populates a list that is used for a loop, and the first failure in each loop will cause failure of the test.  With parametrize, both instances will proceed.  I also wonder if you apply a parametrization to many tests at once.

tests/test_enrichment.py
```diff
@@ -19,7 +19,9 @@
 if not os.path.isfile(pyne.nuc_data):
     raise RuntimeError("Tests require nuc_data.h5.  Please run nuc_data_make.")
 
-SOLVERS = ["symbolic", "numeric"]
+@pytest.fixture()
+def solver_array():
+    return ["symbolic", "numeric"]
```
gonuke: Can we do this with parametrize?

tests/test_mcnp.py
```diff
@@ -45,7 +45,7 @@ def test_read_header_block():
     """
 
     for ssrname in ssrnames:
```
gonuke: Can we do this with parametrize?

tests/test_mcnp.py
```diff
@@ -143,7 +143,7 @@ def test_compare():
     Tricky to test... this just verifies that comparisons are done right.
     """
     for ssrname in ssrnames:
-        yield check_compare, ssrname
+        check_compare(ssrname)
```
gonuke: parametrize?

tests/test_mcnp.py
```diff
@@ -164,7 +164,7 @@ def test_put_header_block():
     This tests that information is preserved correctly when written.
     """
     for ssrname, sswname in zip(ssrnames, sswnames):
-        yield check_put_header_block, ssrname, sswname
+        check_put_header_block(ssrname, sswname)
```
gonuke: parametrize?

tests/test_nucname.py
```diff
@@ -488,7 +488,7 @@ def test_znum():
         92,
     ]
     for case, exp in zip(cases, exps):
-        yield check_cases, nucname.znum, case, exp
+        check_cases(nucname.znum, case, exp)
```
gonuke: parametrize?

tests/test_nucname.py
```diff
@@ -518,13 +518,13 @@ def test_anum():
         0,
     ]
     for case, exp in zip(cases, exps):
-        yield check_cases, nucname.anum, case, exp
+        check_cases(nucname.anum, case, exp)
```
gonuke: parametrize?

tests/test_nucname.py
```diff
@@ -518,13 +518,13 @@ def test_anum():
...
 
 
 def test_snum():
     exps = [0, 0, 0, 0, 1, 0, 0, 1, 1, 4, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0]
     for case, exp in zip(cases, exps):
-        yield check_cases, nucname.snum, case, exp
+        check_cases(nucname.snum, case, exp)
```
gonuke: parametrize?

tests/test_nucname.py
```diff
@@ -587,7 +587,7 @@ def test_zzaaam_to_id():
     for val, id in set(zip(vals, caseids)):
         if val is None:
             continue
-        yield check_cases, nucname.zzaaam_to_id, val, id
+        check_cases(nucname.zzaaam_to_id, val, id)
```
gonuke, line 586: parametrize?

tests/test_nucname.py
```diff
@@ -717,10 +717,10 @@ def test_mcnp_to_id():
     for val, id in set(zip(vals, caseids)):
         if val is None:
             continue
-        yield check_cases, nucname.mcnp_to_id, val, id
+        check_cases(nucname.mcnp_to_id, val, id)
```
gonuke, line 714: parametrize?

tests/test_nucname.py
```diff
@@ -781,10 +781,10 @@ def test_openmc_to_id():
     for val, id in set(zip(vals, caseids)):
         if val is None:
             continue
-        yield check_cases, nucname.openmc_to_id, val, id
+        check_cases(nucname.openmc_to_id, val, id)
```
gonuke, line 776: parametrize?

tests/test_nucname.py
```diff
@@ -846,7 +846,7 @@ def test_serpent_to_id():
     for val, id in set(zip(vals, caseids)):
         if val is None:
             continue
-        yield check_cases, nucname.serpent_to_id, val, id
+        check_cases(nucname.serpent_to_id, val, id)
```
gonuke, line 838: parametrize?

tests/test_nucname.py
```diff
@@ -896,7 +896,7 @@ def test_nist_to_id():
     for val, id in set(zip(vals, caseids)):
         if val is None:
             continue
-        yield check_cases, nucname.nist_to_id, val, id
+        check_cases(nucname.nist_to_id, val, id)
```
gonuke, line 885: parametrize?

tests/test_nucname.py
```diff
@@ -946,7 +946,7 @@ def test_cinder_to_id():
     for val, id in set(zip(vals, caseids)):
         if val is None:
             continue
-        yield check_cases, nucname.cinder_to_id, val, id
+        check_cases(nucname.cinder_to_id, val, id)
```
gonuke, line 933: parametrize?

tests/test_nucname.py
```diff
@@ -1000,7 +1000,7 @@ def test_alara_to_id():
     for val, id in set(zip(vals, caseids)):
         if val is None:
             continue
-        yield check_cases, nucname.alara_to_id, val, id
+        check_cases(nucname.alara_to_id, val, id)
```
gonuke, line 984: parametrize?

tests/test_nucname.py
```diff
@@ -1098,34 +1098,34 @@ def test_sza_to_id():
...
-        yield check_cases, nucname.sza_to_id, val, id
+        check_cases(nucname.sza_to_id, val, id)
 
 
 def test_isnuclide():
     are = [922350, "U235"]
     arent = ["U3", -30060000]
     for nuc in are:
-        yield assert_true, nucname.isnuclide(nuc)
+        assert nucname.isnuclide(nuc)
```
gonuke: parametrize?

tests/test_nucname.py
```diff
@@ -1098,34 +1098,34 @@ def test_sza_to_id():
...
-        yield assert_false, nucname.isnuclide(nuc)
+        assert not nucname.isnuclide(nuc)
 
 
 def test_iselement_U235():
     are = [92, "U"]
     arent = [922350, "U235"]
     for nuc in are:
-        yield assert_true, nucname.iselement(nuc)
+        assert nucname.iselement(nuc)
```
gonuke: parametrize?

tests/test_nucname.py
```diff
@@ -1098,34 +1098,34 @@ def test_sza_to_id():
...
-        yield assert_false, nucname.iselement(nuc)
+        assert not nucname.iselement(nuc)
 
 
 def test_iselement_H1():
     are = [1, "H"]
     arent = [1001, "H1"]
     for nuc in are:
-        yield assert_true, nucname.iselement(nuc)
+        assert nucname.iselement(nuc)
```
gonuke: parametrize?

tests/test_rxname.py
```diff
@@ -58,7 +58,7 @@ def test_hash():
         "n_8",
     ]
     for rx in rxs:
-        yield assert_equal, rxname.hash(rx), _hash(rx)
+        assert rxname.hash(rx) == _hash(rx)
```
gonuke: parametrize?

gonuke (changes requested):
Thanks @ngranda - this was a big effort!! A few comments, mostly about using parametrize in more places....

tests/ci-run-tests.sh
ngranda: Exit code  `5` is given if no tests are collected. I added it because `test_dtypes.py` has no tests in it causing tests after this file to stop running. I am still determining the purpose of this file. We might consider removing this file or populating it with tests.

tests/test_cram.py
ngranda: My understanding of pytest parametrize is that you can parametrize one function or a class at a time but I didn't easily find if I could parametrize many tests at once.  I might be able to parametrize the fixture so I can call the fixture from each of the tests.

tests/test_nucname.py
ngranda: Question: The next 6-8 functions that were yield-based tests have an `if` statement that skipped testing the value if the value is `None`. I have rewritten it to only check the case if the value is not `None`. It also seems that in some tests there is no value `None` so I am wondering if we should keep the `if` statement for these tests? and what the value of adding `None` to a test suite is if we are skipping it anyway?

tests/test_alara.py
ngranda: Yeah, this syntax works. The problem in `test_dagmc.py`  was that the value if was expecting when the `Exception` is raised was `None` but with pytests you get an `ExceptionInfo` object

tests/test_nucname.py
gonuke: I see what you are saying.  I wonder if we can setup the parameters differently??

In each case, it is matching some formatted version of a nuclide name (in `vals`) with its canonical version (in `caseids`).  

What about 
```
set( (val, casid) in zip([...], caseids) if val not None )
```
or something like this..?? not sure if that will work.

This will strip out the `None` values from `vals` after zipping them together with `caseids`.

gonuke: I wonder if it would work to make a function that did this that you use in the `parametrize()`?
```
def build_params(vals, caseids):
    params = zip(vals, caseids)
    params = set( (val, caseid) for val, caseid in params if val not None)
    return params

@pytest.mark.parametrize("val, id", build_params([...], caseids))
def fucn():
```

tests/test_partisn.py
```diff
@@ -216,14 +215,7 @@ def test_get_coord_sys_3D():
...
 
-    try:
-        from pyne import dagmc
-    except:
-        raise SkipTest
-
-    if not HAVE_PYMOAB:
-        raise SkipTest
-
+    from pyne import dagmc
```
gonuke, line 218: don't we need to do the same set of exception handling and tests for `HAVE_PYMOAB` here?

tests/test_partisn.py
```diff
@@ -282,14 +282,8 @@ def test_get_zones_no_void():
...
     """Test that _get_zones function gives results in zyx order."""
-    try:
-        from pyne import dagmc
-    except:
-        raise SkipTest
-
-    if not HAVE_PYMOAB:
-        raise SkipTest
 
+    from pyne import dagmc
```
gonuke, line 286: In some places you have replaced these tests and in others you haven't?

gonuke (approved):
I think this looks good as-is.  I had some thoughts on how to better do the `nucname` tests, but that can be a separate improvement.

Thanks for sticking through this long and tedious PR @ngranda !

PR #1484, opened 2023-06-22
Test install script on multiple ubuntu versions

gonuke:
Please follow these guidelines when making a Pull Request.

## Description
This expands the testing of the install script to the include the latest version of Ubuntu (22.04) and one version older (20.04), using default python versions and pip installed packages.

## Motivation and Context
This helps ensure compatibility with the widest range of potential users.

## Changes
Testing improvement.

## Behavior
The testing previously only used the newest Ubuntu version but now tests on 2 versions.


.github/workflows/install_script.yml
```diff
@@ -18,11 +18,12 @@ on:
 
 jobs:
   run_install_script:
-    runs-on: ubuntu-latest
     strategy: 
       matrix:
         branch: ['develop', 'stable'] # 'develop' can be anything that's not 'stable'. This triggers an if statement in ubuntu.sh
         build_hdf5: ['NO', 'hdf5-1_12_0']
+        os: ['ubuntu-latest', 'ubuntu-20.04']
```
ahnaf-tahmid-chowdhury: I believe we should consider using a specific version of Ubuntu as the latest regularly updated version, and soon we will have Ubuntu 24.04. 

```suggestion
        os: ['ubuntu-22.04', 'ubuntu-20.04']
```

ahnaf-tahmid-chowdhury (changes requested):
Thanks, @gonuke. I just have a small suggestion I'd like to share.

.github/workflows/install_script.yml
```diff
@@ -18,11 +18,12 @@ on:
 
 jobs:
   run_install_script:
-    runs-on: ubuntu-latest
     strategy: 
       matrix:
         branch: ['develop', 'stable'] # 'develop' can be anything that's not 'stable'. This triggers an if statement in ubuntu.sh
         build_hdf5: ['NO', 'hdf5-1_12_0']
+        os: ['ubuntu-22.04', 'ubuntu-20.04']
```
gonuke, line 25: Now that I accepted @ahnaf-tahmid-chowdhury's change, I realize that our standard testing all happens on `ubuntu-latest` so this was mostly about adding `ubuntu-20.04`.  Thus there may be a larger question about whether we want to use `ubuntu-latest` in our regular testing?

ahnaf-tahmid-chowdhury: For now, the latest version of Ubuntu is 22.04. So, I believe we can re-add ubuntu-latest when 24.04 is released. Alternatively, we can always use latest and create a PR to add 22.04 when Ubuntu 24.04 is released.

Let me know which option we may proceed.

gonuke: Probably makes sense to change all of them to `22.04` so that things don't suddenly break when 24.04 is released.

gonuke:
Feel free to approve/merge this, but then we'll want a PR that carefully updates our other CI pieces, perhaps

.github/workflows/install_script.yml
ahnaf-tahmid-chowdhury: Yes, by setting specific version, this will be under our control.

ahnaf-tahmid-chowdhury (approved):

PR #1485, opened 2023-06-24
avoid use of deprecated numpy.bool alias

neutrinoceros:
## Description

Fix a compatibility issue with numpy>=1.24 (similar to #1479)

## Motivation and Context
`np.bool` was an alias to builtin `bool`. This alias was deprecated in numpy 1.20 and was removed in numpy 1.24

### DIFF
--- a/pyne/mesh.py
+++ b/pyne/mesh.py
@@ -67,7 +67,8 @@ err__ops = {
     ),
 }
 
-_INTEGRAL_TYPES = (int, np.integer, np.bool_)
+_INTEGRAL_TYPES = (int, np.integer, bool, np.bool_)
+_BOOLEAN_TYPES = (bool, np.bool_)
 _SEQUENCE_TYPES = (Sequence, np.ndarray)
 
 
@@ -155,7 +156,7 @@ class MaterialPropertyTag(Tag):
             return getattr(mats[int(key)], name)
         elif isinstance(key, slice):
             return np.array([getattr(mats[i], name) for i in range(*key.indices(size))])
-        elif isinstance(key, np.ndarray) and key.dtype == np.bool:
+        elif isinstance(key, np.ndarray) and key.dtype in _BOOLEAN_TYPES:
             if len(key) != size:
                 raise KeyError("boolean mask must match the length of the mesh.")
             return np.array([getattr(mats[i], name) for i, b in enumerate(key) if b])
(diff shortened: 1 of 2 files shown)


pyne/mesh.py
```diff
@@ -155,7 +156,7 @@ def __getitem__(self, key):
             return getattr(mats[int(key)], name)
         elif isinstance(key, slice):
             return np.array([getattr(mats[i], name) for i in range(*key.indices(size))])
-        elif isinstance(key, np.ndarray) and key.dtype == np.bool:
+        elif isinstance(key, np.ndarray) and key.dtype in _BOOLEAN_TYPES:
```
ahnaf-tahmid-chowdhury, line 159: I believe that setting `key.dtype == bool` may be more efficient in this case. This comparison checks if the `data type (dtype)` of key is exactly `bool`. On the other hand, using `key.dtype in (bool, np.bool_)` checks if the dtype is either `bool` or `np.bool_`. This approach involves iterating over the elements of the tuple to find a match, which could potentially be slower.

ahnaf-tahmid-chowdhury (changes requested):

pyne/mesh.py
neutrinoceros: You are right that my proposed change would be technically slower, involving 2 equality checks instead of one, but note that
- we're talking 10s of nano-seconds here (hopefully this doesn't matter)
- because `bool` is the *first* element in `_BOOLEAN_TYPES`, the second check is only performed in case `key.dtype != bool`
- if performance was important here (again, I'm hoping it's not), `key.dtype is bool` would be the way to go (identity check is always cheaper than equality check, although here it's 20ns VS 24ns on my machine)

ahnaf-tahmid-chowdhury: Considering the potential impact on simulation time when the code is called multiple times (considering large data), it becomes more important to optimize the comparison. In that case, using `key.dtype == bool` would be more efficient compared to `key.dtype in (bool, np.bool_)` due to the reduced number of operations involved. However, it's still advisable to measure the actual performance gain before making a decision, as the difference might still be minimal.

gonuke: Does the proposed approach ensure some kind of backward compatibility?  Are there configurations that may use one vs the other?  Or is `bool` always available so that any invocation will always work if we restrict to `bool`?  Put another way, what's the benefit to this proposed approach?

neutrinoceros: My approach is supposed to be 100% backward compatible. Additionally, it survives `np.bool_` (current implementation doesn't)

gonuke:
Thanks @neutrinoceros

PR #1489, opened 2023-07-10
Add composite action for BuildTest job

bquan0:
## Description
Adds a composite action to replace the BuildTest job used in `docker_publish.yml` and `build_test.yml`.

## Motivation and Context
Requested by @gonuke. This change will keep the BuildTest jobs in the two workflows consistent and requires less effort to maintain the consistency.

## Behavior
Current behavior is that the BuildTest jobs in `docker_publish.yml` and `build_test.yml` have become inconsistent with each other due to previous pull requests I made where I only updated the BuildTest job in `docker_publish.yml`. Now, I have put the BuildTest job from `docker_publish.yml` (the most updated version) into the composite action and it is being used in both workflows. 

## Other Information
The composite action can only include the `steps` of the BuildTest job, which means that the `matrix` and the `container` of the BuildTest job still need to be updated in their respective workflow files. However, the `container` of the BuildTest job already had differences between the two workflow files in terms of their tag, so this issue wouldn't have been fixed with the composite action anyway. 
Here's a [workflow run](https://github.com/bquan0/pyne/actions/runs/5509596417) to show that the BuildTest job still functions as intended with the composite action.

### DIFF
--- a/.github/actions/build-test/action.yml
+++ b/.github/actions/build-test/action.yml
@@ -29,18 +29,18 @@ runs:
         export ADD_FLAG="${ADD_FLAG} "
         echo "ADD_FLAG=${ADD_FLAG}" >> $GITHUB_ENV
         
-      - name: Building PyNE
-        shell: bash -l {0}
-        run: |
-          cd $GITHUB_WORKSPACE
-          python setup.py install --user --clean ${{ env.ADD_FLAG}}
-          export PATH="$PATH:/github/home/.local/bin"
-          export PYTHONPATH="$PYTHONPATH:/github/home/.local/lib/python3.10/site-packages/"
-          cd ../
-          nuc_data_make
+    - name: Building PyNE
+      shell: bash -l {0}
+      run: |
+        cd $GITHUB_WORKSPACE
+        python setup.py install --user --clean ${{ env.ADD_FLAG}}
+        export PATH="$PATH:/github/home/.local/bin"
+        export PYTHONPATH="$PYTHONPATH:/github/home/.local/lib/python3.10/site-packages/"
+        cd ../
+        nuc_data_make
 
-      - name: Testing PyNE
-        shell: bash -l {0}
-        run: |
-          cd $GITHUB_WORKSPACE/tests
-          ./ci-run-tests.sh python3
+    - name: Testing PyNE
+      shell: bash -l {0}
+      run: |
+        cd $GITHUB_WORKSPACE/tests
+        ./ci-run-tests.sh python3
--- a/.github/actions/build-test/action.yml
+++ b/.github/actions/build-test/action.yml
@@ -20,8 +20,7 @@ inputs:
 
 runs:
   using: "composite"
-  container:
-    image: ghcr.io/${{ inputs.owner }}/pyne_ubuntu_22.04_py3${{ inputs.hdf5 }}/${{ inputs.stage }}${{ inputs.tag }}
+  image: ghcr.io/${{ inputs.owner }}/pyne_ubuntu_22.04_py3${{ inputs.hdf5 }}/${{ inputs.stage }}${{ inputs.tag }}
   steps: 
     - name: setup
       shell: bash -l {0}
--- a/.github/actions/build-test/action.yml
+++ b/.github/actions/build-test/action.yml
@@ -1,13 +1,12 @@
 name: BuildTest
-description: tests one docker image built in docker_publish.yml. Used to keep 
-    the build test procedure consistent in docker_publish.yml and build_test.yml
+description: Builds and tests an installation of PyNE in a single docker image.  This action is used for bootstrap testing when building the docker images as well as testing of each PR using those docker images.
 inputs:
   stage:
-    description: matrix variable
+    description: The docker stage that will be used as the docker image for testing.
     required: true
     default: ''
   hdf5:
-    description: matrix variable
+    description: Information about which version of HDF5 will be used for testing.
     required: true
     default: ''
 runs:
--- a/.github/actions/build-test/action.yml
+++ b/.github/actions/build-test/action.yml
@@ -9,6 +9,16 @@ inputs:
     description: Information about which version of HDF5 will be used for testing.
     required: true
     default: ''
+  tag:
+    description: The tag for the docker image used to run the tests
+    required: true
+    default: ''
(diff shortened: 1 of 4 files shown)


.github/actions/build-test/action.yml
```diff
@@ -0,0 +1,46 @@
+name: BuildTest
+description: tests one docker image built in docker_publish.yml. Used to keep 
+    the build test procedure consistent in docker_publish.yml and build_test.yml
```
gonuke:
```suggestion
description: Builds and tests an installation of PyNE in a single docker image.  This action is used for bootstrap testing when building the docker images as well as testing of each PR using those docker images.
```

.github/actions/build-test/action.yml
```diff
@@ -0,0 +1,46 @@
...
+inputs:
+  stage:
+    description: matrix variable
```
gonuke:
```suggestion
    description: The docker stage that will be used as the docker image for testing.
```

.github/actions/build-test/action.yml
```diff
@@ -0,0 +1,46 @@
...
+    required: true
+    default: ''
+  hdf5:
+    description: matrix variable
```
gonuke:
```suggestion
    description: Information about which version of HDF5 will be used for testing.
```

.github/actions/build-test/action.yml
```diff
@@ -0,0 +1,46 @@
...
+        export ADD_FLAG="${ADD_FLAG} "
+        echo "ADD_FLAG=${ADD_FLAG}" >> $GITHUB_ENV
+        
+    - name: Building PyNE
+      shell: bash -l {0}
+      run: |
+        cd $GITHUB_WORKSPACE
+        python setup.py install --user --clean ${{ env.ADD_FLAG}}
+        export PATH="$PATH:/github/home/.local/bin"
+        export PYTHONPATH="$PYTHONPATH:/github/home/.local/lib/python3.10/site-packages/"
```
gonuke: Do we know if this is necessary? If so, how do we make it more robust to changing python versions?

.github/workflows/build_test.yml
```diff
@@ -34,41 +34,15 @@ jobs:
       fail-fast: false
 
     container:
-      image: ghcr.io/pyne/pyne_ubuntu_20.04_py3${{ matrix.hdf5 }}/${{ matrix.stage }}:stable
+      image: ghcr.io/pyne/pyne_ubuntu_22.04_py3${{ matrix.hdf5 }}/${{ matrix.stage }}:stable
```
gonuke, line 37: How do we make sure this is consistent with the docker images we've built? Maybe there is no good way to automate this? Or maybe we just take some of the modifiers out of the image name (e.g. ubuntu version).  This could be a good topic for a software meeting.

gonuke (changes requested):
Thanks for this @bquan0 .  I've included a few suggestions to improve documentation, but also think this would be a good to clean up a few issues that are not part of these changes (and therefore can't be places where I can place comments).
* we don't use the matrix variable `ubuntu_versions` in the docker_publish action
* the use of output and/or env variables for tag names seems a little random and may need to be cleaned up in docker_publish
* might want more descriptive `id` for the two builds in docker_publish

.github/workflows/build_test.yml
bquan0: There isn't a good way to achieve consistency because env variables can't be used in `image:`. Furthermore, there's no way to ensure consistency across `docker_publish.yml` and `build_test.yml` because `image:` can't be in the composite action. We do have the option to remove the ubuntu version from the image name since that's something that we specify in the env variable `DOCKER_IMAGE_BASENAME`.

bquan0:
The use of output and env variables for tag names is currently necessary in order to test and push the images that we build in the first job of `docker_publish.yml` because we need to have the tag the action generates in order to use the correct image. The `tag-latest-on-default` option doesn't help here because it only gives a `latest` tag if the workflow is run on the default branch.

.github/workflows/build_test.yml
gonuke: Yes - I was wondering we can just do without the ubuntu version given the streamlining we are doing.  We can probably remove the `py3` as well since we are always using Python 3 now.  In short, maybe all identifiers should be machine-generated rather than hard coded.

gonuke (approved):
This looks better now.  I'll check the builds and then merge as-is.  We can revisit the change of the image name more carefully since we probably need to modify it in two separate steps, starting with the docker_publish workflow first.

.github/workflows/docker_publish.yml
```diff
@@ -88,40 +79,16 @@ jobs:
...
-          cd $GITHUB_WORKSPACE/tests
-          ./ci-run-tests.sh python3
+
+      - name: use BuildTest composite action
+        uses: ./.github/actions/build-test
+        with:
+          stage: ${{ matrix.stage }}
+          hdf5: ${{ matrix.hdf5 }}
+          tag: ${{ env.DOCKER_IMAGE_TAG }}
+          owner: ${{ github.repository_owner }}
```
gonuke: There are error messages saying that these inputs are not used.  Maybe you discovered that we can't use them?? If so, we should remove them here.

.github/workflows/build_test.yml
```diff
@@ -34,41 +34,17 @@ jobs:
...
-        run: |
-          cd $GITHUB_WORKSPACE/tests
-          ./ci-run-tests.sh
+      - name: use BuildTest composite action
+        uses: ./.github/actions/build-test
+        with:
+          stage: ${{ matrix.stage }}
+          hdf5: ${{ matrix.hdf5 }}
+          tag: ':stable'
+          owner: 'pyne'
```
gonuke: There are warnings that these inputs aren't used.  Do we need to update the action to use these? or remove the inputs?

gonuke (changes requested):
I found some warnings in the github builds around these extra inputs

.github/workflows/docker_publish.yml
bquan0: Yep, those are from when I tried to use `image:` in the composite action. I'll remove them.

gonuke (approved):

gonuke:
Thanks @bquan0 - let's put this into production

PR #1490, opened 2023-07-21
Add workflow to create virtual machine from Dockerfile

bquan0:
## Description
Adds the workflow `virtualbox_image.yml` to create a virtual box disk image using the [d2vm](https://github.com/linka-cloud/d2vm) command. 

## Motivation and Context
Requested by @gonuke. This will allow us to distribute virtual machines that are based on our Docker images. 

## Behavior
The new behavior is that the workflow generates virtual machines based on `ubuntu_22.04-dev.dockerfile`. However, I ran into some errors while the d2vm command was building the virtual machines, as can be seen in this [workflow run](https://github.com/bquan0/pyne/actions/runs/5624454309). I'm not sure how to fix these errors as it seems like it's an issue with the contents of the dockerfile instead of an issue with the d2vm command.

### DIFF
--- a/.github/workflows/virtualbox_image.yml
+++ b/.github/workflows/virtualbox_image.yml
@@ -27,6 +27,13 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Build VirtualBox image using d2vm
         run: |
           curl -sL "https://github.com/linka-cloud/d2vm/releases/download/v0.2.0/d2vm_v0.2.0_linux_amd64.tar.gz" | tar -xvz d2vm
--- a/.github/workflows/virtualbox_image.yml
+++ b/.github/workflows/virtualbox_image.yml
@@ -81,7 +81,15 @@ jobs:
       - name: Build VirtualBox image using d2vm
         run: |
           sudo apt-get update
-          sudo apt-get install qemu-utils
+          sudo apt-get install -y --fix-missing \
+            util-linux \
+            udev \
+            parted \
+            e2fsprogs \
+            mount \
+            tar \
+            extlinux \
+            qemu-utils            
           curl -sL "https://github.com/linka-cloud/d2vm/releases/download/v0.2.0/d2vm_v0.2.0_linux_amd64.tar.gz" | tar -xvz d2vm
           sudo mv d2vm /usr/local/bin/
           echo "make worked"
--- a/.github/workflows/virtualbox_image.yml
+++ b/.github/workflows/virtualbox_image.yml
@@ -80,7 +80,8 @@ jobs:
 
       - name: Build VirtualBox image using d2vm
         run: |
-          sudo apt-get qemu-utils
+          sudo apt-get update
+          sudo apt-get install qemu-utils
           curl -sL "https://github.com/linka-cloud/d2vm/releases/download/v0.2.0/d2vm_v0.2.0_linux_amd64.tar.gz" | tar -xvz d2vm
           sudo mv d2vm /usr/local/bin/
           echo "make worked"
--- a/.github/workflows/virtualbox_image.yml
+++ b/.github/workflows/virtualbox_image.yml
@@ -68,6 +68,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
       - name: Log in to the Container registry
         uses: docker/login-action@v2
         with:
--- a/.github/workflows/virtualbox_image.yml
+++ b/.github/workflows/virtualbox_image.yml
@@ -20,9 +20,7 @@ jobs:
 
       - name: Build VirtualBox image using d2vm
         run: |
-          git clone https://github.com/linka-cloud/d2vm && cd d2vm
-          make install
-          cd ..
+          curl -sL "https://github.com/linka-cloud/d2vm/releases/download/v0.2.0/d2vm_v0.2.0_linux_arm64.tar.gz" | tar -xvz d2vm
+          sudo mv d2vm /usr/local/bin/
           echo "make worked"
-          which d2vm
-          sudo ./go/bin/d2vm build -p ${{ env.VM_PASSWORD }} -f ./docker/ubuntu_22.04-dev.dockerfile -o pyne.vdi .
+          sudo d2vm build -p ${{ env.VM_PASSWORD }} -f ./docker/ubuntu_22.04-dev.dockerfile -o pyne.vdi .
--- a/.github/workflows/virtualbox_image.yml
+++ b/.github/workflows/virtualbox_image.yml
@@ -31,7 +31,7 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Build PyNE docker image
-        ses: firehed/multistage-docker-build-action@v1
+        uses: firehed/multistage-docker-build-action@v1
         id: build_pyne
         with:
           repository: ${{ env.DOCKER_IMAGE_BASENAME }}
--- a/.github/workflows/virtualbox_image.yml
+++ b/.github/workflows/virtualbox_image.yml
@@ -22,6 +22,7 @@ jobs:
         run: |
           git clone https://github.com/linka-cloud/d2vm && cd d2vm
           make install
-          which d2vm
+          cd ..
           echo "make worked"
+          which d2vm
           sudo ./go/bin/d2vm build -p ${{ env.VM_PASSWORD }} -f ./docker/ubuntu_22.04-dev.dockerfile -o pyne.vdi .
--- a/.github/workflows/virtualbox_image.yml
+++ b/.github/workflows/virtualbox_image.yml
@@ -50,7 +50,7 @@ jobs:
     runs-on: ubuntu-latest
 
     container:
-      image: ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3/pyne${{ needs.multistage_image_build.outputs.image_tag }}
+      image: ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3/pyne${{ needs.pyne_image_build.outputs.image_tag }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
--- a/.github/workflows/virtualbox_image.yml
+++ b/.github/workflows/virtualbox_image.yml
@@ -14,6 +14,15 @@ env:
 jobs:
   virtualbox_image_build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        stage: [base_python, moab, dagmc, openmc]
+        hdf5: ['']
+        include:
+          - stage: dagmc
+            hdf5: _hdf5
+      fail-fast: false
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
@@ -23,4 +32,5 @@ jobs:
           curl -sL "https://github.com/linka-cloud/d2vm/releases/download/v0.2.0/d2vm_v0.2.0_linux_amd64.tar.gz" | tar -xvz d2vm
           sudo mv d2vm /usr/local/bin/
           echo "make worked"
-          sudo d2vm build -p ${{ env.VM_PASSWORD }} -f ./docker/ubuntu_22.04-dev.dockerfile -o pyne.vdi .
+          sudo d2vm convert ghcr.io/pyne/pyne_ubuntu_22.04_py3${{ matrix.hdf5 }}/${{ matrix.stage }}:stable -p ${{ env.VM_PASSWORD }} -o pyne.vdi
+          
--- a/.github/workflows/virtualbox_image.yml
+++ b/.github/workflows/virtualbox_image.yml
@@ -10,19 +10,60 @@ on:
 
 env:
   VM_PASSWORD: temppwd
+  DOCKER_IMAGE_BASENAME: ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3
+  DOCKER_IMAGE_TAG: :refs_heads_${{ github.ref_name }}-bk0
 
 jobs:
-  virtualbox_image_build:
+  pyne_image_build:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        stage: [base_python, moab, dagmc, openmc]
-        hdf5: ['']
-        include:
-          - stage: dagmc
-            hdf5: _hdf5
-      fail-fast: false
+    outputs:
+      image_tag: ${{ steps.output_tag.outputs.image_tag }}
 
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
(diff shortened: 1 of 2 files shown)


.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Build VirtualBox image using d2vm
+        run: |
+          curl -sL "https://github.com/linka-cloud/d2vm/releases/download/v0.2.0/d2vm_v0.2.0_linux_amd64.tar.gz" | tar -xvz d2vm
+          sudo mv d2vm /usr/local/bin/
+          echo "make worked"
+          sudo d2vm build -p ${{ env.VM_PASSWORD }} -f ./docker/ubuntu_22.04-dev.dockerfile -o pyne.vdi .
```
gonuke: Is it possible to convert an existing image from ghcr.io?

gonuke: Once this is working, I think you'll need to do something in github actions to capture the file that produced as an asset.

gonuke (changes requested):
Thanks @bquan0!

I think this is failing because of a new incompatibility between Cython 3 and MOAB.  I'm working on that separately!  It will definitely get in the way of demonstrating this, though.

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,26 @@
...
+env:
+  VM_PASSWORD: temppwd
+
+jobs:
+  virtualbox_image_build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
```
ahnaf-tahmid-chowdhury:
```suggestion

      - name: Log in to GitHub Container Registry
        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
        
      - name: Pull Docker image from ghcr.io
        run: docker pull ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3:refs_heads_${{ github.ref_name }}-bk0
        
```

.github/workflows/virtualbox_image.yml
ahnaf-tahmid-chowdhury:
```suggestion
          sudo d2vm build -p ${{ env.VM_PASSWORD }} -i ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3:refs_heads_${{ github.ref_name }}-bk0
          
      - name: Upload VirtualBox image as artifact
        uses: actions/upload-artifact@v2
        with:
          name: virtualbox_image
          path: pyne.vdi
```

ahnaf-tahmid-chowdhury: Yes, it is possible to convert an existing Docker image from ghcr.io

ahnaf-tahmid-chowdhury (changes requested):
As @gonuke requested, here is the modified version:

gonuke:
Thanks for your suggestions @ahnaf-tahmid-chowdhury.  I think there may be other ways that @bquan0 has experience addressing some of theses, and others he'll probably incorporate directly.  I'll work with him to bring this forward.

bquan0:
I tried converting an existing image from ghcr.io, but I ran into some issues with the name of the images. I don't think it's possible for a forked repo to use an image from the ghcr of `pyne`, so it has to be an image from the user's package library. However, forked repos don't have an image with the `stable` tag because `docker_publish.yml` doesn't run the job where it publishes images with the `stable` tag unless the user is `pyne`, which means that this workflow won't work on forked repos and the pyne repo at the same time if we want to use the `stable` image. 
I also ran into some QEMU errors in some of the [workflow runs](https://github.com/bquan0/pyne/actions/runs/5675669899/job/15381246354#logs) which I am confused about.

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,43 @@
...
+env:
+  VM_PASSWORD: temppwd
+
+jobs:
+  virtualbox_image_build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        stage: [base_python, moab, dagmc, openmc]
```
gonuke: I think we'll only want to convert the final images with PyNE installed, and not the full sequence of these.  Most of these are only relevant for testing.  Therefore, we might not even need a matrix.

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,43 @@
...
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build VirtualBox image using d2vm
+        run: |
+          curl -sL "https://github.com/linka-cloud/d2vm/releases/download/v0.2.0/d2vm_v0.2.0_linux_amd64.tar.gz" | tar -xvz d2vm
+          sudo mv d2vm /usr/local/bin/
+          echo "make worked"
+          sudo d2vm convert ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3${{ matrix.hdf5 }}/${{ matrix.stage }}:latest -p ${{ env.VM_PASSWORD }} -o pyne.vdi
```
gonuke: This may work for the PyNE GHCR repo if we only pick the final image with PyNE installed??

gonuke (changes requested):
Not sure if my suggestions will help your other issue?

.github/workflows/virtualbox_image.yml
bquan0: I've just looked at the ubuntu 22.04 packages of PyNE, `docker_publish.yml`, and the dockerfile and I realized that we don't ever build the `pyne` stage at the end of the dockerfile, which means that there are no packages of PyNE that have PyNE installed. We might have to merge a PR that changes the inputs to the multistage build action to include the `pyne` stage before we continue with with this PR.

gonuke: Right! I keep getting this confused with the DAGMC case.

Maybe the following?
- Add a stage to build PyNE in the Dockerfile 
- change this new workflow to be about deployment so it will:
   * build & pus the docker image with PyNE using the multistage build for efficiency/consistency 
   * convert it to a VM image

bquan0: I think we should add the building of the docker image with PyNE to the `docker_publish.yml` workflow because it already has all the infrastructure of the multistage build actions and the jobs that will test and push the images if they are working. I think the `virtualbox_image.yml` workflow should only take the PyNE docker image from ghcr.io and convert it into a VM image.

gonuke: It's definitely a simpler option, but then don't we have to run that action with every merged pull-request? Maybe the docker image caching provided by the multistage merge action means that it's not too much load? We currently only run this action rarely to update the dependencies.

bquan0:
In `virtualbox_image.yml`, I added a job to build the pyne image with the multistage build action as well as a job to test the new image. However, I ran into the Cython3 errors again while building the pyne image in [this workflow](https://github.com/bquan0/pyne/actions/runs/5693759200/job/15433537459), so we won't be able to see this working for the time being.

gonuke:
Take a look at the dependencies on the `d2vm` Readme page.  Some of these must not be present in the standard GH action runner.  I think you'll need to run `apt` to install them before trying to run `d2vm`

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,97 @@
...
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build VirtualBox image using d2vm
+        run: |
+          sudo apt-get update
+          sudo apt-get install qemu-utils
```
gonuke: You may want to go ahead and install the full list of dependencies

gonuke (commented):
Just a thought to leap frog the next CI failure

.github/workflows/virtualbox_image.yml
bquan0: After adding all the dependencies, it seems like `d2vm` has [successfully](https://github.com/bquan0/pyne/actions/runs/5772596807/job/15647740501) converted an existing image from ghcr.io.

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,105 @@
...
+            tar \
+            extlinux \
+            qemu-utils            
+          curl -sL "https://github.com/linka-cloud/d2vm/releases/download/v0.2.0/d2vm_v0.2.0_linux_amd64.tar.gz" | tar -xvz d2vm
+          sudo mv d2vm /usr/local/bin/
+          echo "make worked"
+          sudo d2vm convert ghcr.io/bquan0/pyne_ubuntu_22.04_py3/openmc:df359b744263d9869f345ad0f374f7338b920c0d -p ${{ env.VM_PASSWORD }} -o pyne.vdi
+
+      - name: Upload VirtualBox image as artifact
+        uses: actions/upload-artifact@v2
```
gonuke: maybe try v3 to avoid the warning you got in the action
```suggestion
        uses: actions/upload-artifact@v3
```

gonuke (changes requested):
It's great that you got this to work!!! Nice work.  

I saw that the action produced a warning about which version of node was expected/available, and also saw that there is a `v3` of the action - maybe that will help?

bquan0:
Yep, the new version got rid of the warning [(workflow run)](https://github.com/bquan0/pyne/actions/runs/5773556835). Thanks for the suggestion!

gonuke:
Since this is working in principle, can you please clean it up so that we can consider merging it.

bquan0:
Cleaned it up by removing the commented out build and testing jobs which built a new docker image at the `pyne` stage. Also made `virtualbox_image.yml` convert an image built to the `openmc` stage in the `github.actor`'s ghcr.io instead of a specific image in my ghcr.io.
I also rebased this branch onto `develop` as part of the clean up.

gonuke:
The ultimate goal is to be able to create a VM that includes PyNE.  Somewhere, we'll need to create an image that includes PyNE.  One reason to do it as part of this workflow is that we can get the dependencies right - we can build it immediately before we pull it to make VM.  If we build it in the `docker_publish` workflow, we need to worry about the race condition of running this workflow before that one is finished making the new image.

gonuke:
I just made the "openmc" stages public on the upstream repo, so try that as well.  (I spent a bunch of time experimenting with permissions on the weekend...)

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,52 @@
...
+            parted \
+            e2fsprogs \
+            mount \
+            tar \
+            extlinux \
+            qemu-utils            
+          curl -sL "https://github.com/linka-cloud/d2vm/releases/download/v0.2.0/d2vm_v0.2.0_linux_amd64.tar.gz" | tar -xvz d2vm
+          sudo mv d2vm /usr/local/bin/
+          echo "make worked"
+          sudo d2vm convert ghcr.io/${{ github.actor }}/pyne_ubuntu_22.04_py3/openmc:${{ env.DOCKER_IMAGE_TAG }} -p ${{ env.VM_PASSWORD }} -o pyne.vdi
```
gonuke: I've made the packages public so try this?
```suggestion
          sudo d2vm convert ghcr.io/pyne/pyne_ubuntu_22.04_py3/openmc:${{ env.DOCKER_IMAGE_TAG }} -p ${{ env.VM_PASSWORD }} -o pyne.vdi
```

gonuke (changes requested):
Try the public package from PyNE

.github/workflows/virtualbox_image.yml
bquan0: Yep, looks like[ it's working](https://github.com/bquan0/pyne/actions/runs/5792107017/job/15697909920) now that the package has been made public. I didn't take the suggestion because I had to change the `env.DOCKER_IMAGE_TAG` into `latest` because the ghcr.io of pyne doesn't have a `virtual-machine` branch, so there wouldn't have been any images with a tag that would've matched `env.DOCKER_IMAGE_TAG` since it's based on the branch that the workflow is run on.

gonuke:
We (I) have to think through more carefully what's going on here. You've added a build using Docker, that will a,ways use the `develop` branch and then you BuildTest using the current checkout (but that never becomes part of the image), and then you use an existing image (this is in flux pending permissions) for making the VM.

Seeing this had reminded me of the `release` trigger and maybe we only want to do this action on release.  Therefore, perhaps we should also be developing this as a PR from a branch in the upstream repo rather than your fork.

gonuke:
I've given you write access to the upstream repo (should have done this long ago) so you can push a branch directly there for testing

bquan0:
I'm confused about what to do next. Currently, we do both the building and the testing on the `virtual-machine` branch, then 
I made some changes yesterday to the image that gets converted into the VM so that it now converts the image that was just built by the previous jobs.

ahnaf-tahmid-chowdhury:
I think, @gonuke is suggesting the workflow should be developed as a PR from an upstream branch rather than using a fork. Also, consider triggering this action only on release, as it seems to be building a VM image which could avoid unnecessary executions during normal development.

```yml
on:
  release:
    types:
      - published
```

gonuke:
> I'm confused about what to do next. Currently, we do both the building and the testing on the `virtual-machine` branch, then I made some changes yesterday to the image that gets converted into the VM so that it now converts the image that was just built by the previous jobs.

There are a few different issues here, so let me elaborate:

1. The goal of this PR is to provide final assets (docker image or VM) for potential users to be able to use instead of the trouble of installing PyNE themselves.  (other choices could be conda and/or pip) As a result, it should probably only be invoked when we issue a `release`, and not for every `push` or for every `pull_request`.  This was not obvious to me when we first started this task, but as I was trying to think through the workflow as it looks now, this occurred to me.

2. Because it only happens when a release occurs, we'll want to be careful about which version/branch of PyNE we actually build when creating the final docker image (from which we will be making the VM).  As the workflow stands in this PR right now, it builds PyNE twice:

     a. once  when you build the `pyne` stage in the multistage-build - this uses the `develop` branch as defined in the Dockerfile
     b. once when you run the BuildTest action, in this case using the branch associated with the current event via the `checkout` action

     We ultimately want to install PyNE once in a docker image using the release version/branch.

3. Once we build and push this docker image, we'll want to tag it as the latest docker image for PyNE so that we can advertise/share it widely.  We'll also want to use exactly that one to make the VM.

4. Because we can't push to the PyNE package repo from public forks, it will be best to do this development as a branch on the upstream PyNE repo, rather than from your fork.  It works largely the same, but instead of pushing your changes to bquan0:virtual-machine you'll push them to pyne:virtual-machine.  Then we'll be able to test it manually including the step of pushing to the GHCR repo.  

     We should probably do this for all workflow changes that result in pushing images.  This is a slight departure from our prior practice, although a common practice in other projects.  We have, until now, expected development to happen in forked repos and PR's to come from those repos.  Allowing that is very inclusive as it allows anyone in the world to make a suggestion without us having to give them write access to the repo.  

     Arguable, if we have contributors recommending changes that will affect our docker images, we are already placing some faith in those contributors, so it makes sense to grant them write access and allow them to develop on branches in the upstream repo.  We can still protect the primary branch by requiring a review, as I did yesterday.

That's the long version of all the thoughts I had in my head when I wrote my last comments. I hope that clarifies.  (I should stop trying to be too terse in my GH comments.)

bquan0:
Sounds good. I've pushed some changes regarding these instructions to the `virtual-machine` branch in the upstream PyNE repo and also opened a PR (the one above this comment). Should I close this PR now that we're using the branch in PyNE?

gonuke:
I think you should close this PR @bquan0

bquan0:
Closing this PR as requested.

Issue #1491, opened 2023-07-22
Test install scripts with every merge


gonuke:
With #1477 we created a workflow to test a standard installation script within this repo.

Currently, he conditions under which that script runs are too limiting:
```
  push:
    paths:
      - 'scripts/ubuntu.sh'
      - '.github/workflows/install_script.yml'
```

I think this test should run on every `push` to `develop` and on every `pull_request` to `develop`.  By doing it on `pull_request` it will provide early insight to contributors that the installation is broken, but may also place too much burden on them.  We can revisit this after some time.

PR #1494, opened 2023-07-25
Update run conditions for workflows

### LINKED ISSUE #1292: Add check to avoid CI build/test for pure documentation changes
To the extent possible, it would be great to detect if a PR has only documentation changes and then skip doing any build/test in CI.

bquan0:
## Description
Makes `build_test.yml` run when it or the composite action it uses is updated AND it is a push to the `develop` branch in the main repo (not forks). Also makes `install_script.yml` run on every `push` to `develop` and every `pull_request` to develop. 

## Motivation and Context
Closes #1292 and #1491, which request the two features I mentioned above. 

## Behavior
The previous behavior was that `build_test.yml` ran even though there were no changes to itself or the dockerfile that built the image it tested. This is because it used `paths-ignore:` to filter out files instead of using `paths:` to specify files to check if they have been changed.
The previous behavior of `install_script.yml` was that it only ran when it or the install script `ubuntu.sh` was changed. The current behavior is that it now also runs whenever there is a `push` or `pull` to `develop`. 

## Other Information
One thing I was wondering was if I should add `docker/*` to the `paths:` of `build_test.yml` since it's also in `docker_publish.yml`. I didn't add it because I thought it would be redundant since `docker_publish.yml` has `docker/*` in its `paths:` and `docker_publish.yml` has a job which runs the same job found in `build_test.yml`.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -4,16 +4,9 @@ on:
   # allows us to run workflows manually
   workflow_dispatch:
   pull_request:
-    paths:
-      - '.github/workflows/build_test.yml'
-      - '.github/actions/build-test/action.yml'
   push:
     branches:
       - 'develop'
-    paths:
-      - '.github/workflows/build_test.yml'
-      - '.github/actions/build-test/action.yml'
-
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -4,19 +4,13 @@ on:
   # allows us to run workflows manually
   workflow_dispatch:
   pull_request:
-    paths-ignore:
-      - '.github/workflows/docker_publish.yml'
-      - '.github/workflows/changelog_test.yml'
-      - 'docker/**'
-      - 'CHANGELOG.rst'
+    paths:
+      - '.github/workflows/build_test.yml'
+      - '.github/actions/build-test/action.yml'
   push:
-    branches:
-      - develop
-    paths-ignore:
-      - '.github/workflows/docker_publish.yml'
-      - '.github/workflows/changelog_test.yml'
-      - 'docker/**'
-      - 'CHANGELOG.rst'
+    paths:
+      - '.github/workflows/build_test.yml'
+      - '.github/actions/build-test/action.yml'
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -8,6 +8,8 @@ on:
       - '.github/workflows/build_test.yml'
       - '.github/actions/build-test/action.yml'
   push:
+    branches:
+      - 'develop'
     paths:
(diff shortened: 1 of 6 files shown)


.github/workflows/build_test.yml
```diff
@@ -4,19 +4,15 @@ on:
...
   workflow_dispatch:
   pull_request:
-    paths-ignore:
-      - '.github/workflows/docker_publish.yml'
-      - '.github/workflows/changelog_test.yml'
-      - 'docker/**'
-      - 'CHANGELOG.rst'
+    paths:
+      - '.github/workflows/build_test.yml'
+      - '.github/actions/build-test/action.yml'
```
gonuke: This says it will only run if these files change, but we want testing to run for any file changes

.github/workflows/install_script.yml
```diff
@@ -3,7 +3,12 @@ name: run install script
...
   workflow_dispatch:
+  pull_request:
+    branches:
+      - develop
   push:
+    branches:
+      - develop
     paths:
       - 'scripts/ubuntu.sh'
       - '.github/workflows/install_script.yml'
```
gonuke: I think we want this to run for all file changes

gonuke (changes requested):
unless I misunderstand the `paths` scope, I think we want this to run more often than indicated here.

bquan0:
Right, I now see that I misinterpreted the issue for `install_script.yml`. However, on the issue about `build_test.yml`, it says that the workflow shouldn't run if there are only documentation changes.

gonuke:
Fair enough, we can add a path exclusion for documentation on the `build_test.yml`

bquan0:
Just added a list of files to `paths-ignore:` for `build_test.yml`. Let me know if there are any changes I should make to the list!

gonuke:
Maybe we need to revisit/talk through all the circumstances we use to run all of these at a future software meeting.

gonuke:
I think we need to restrict the `cython` installation to be <3.0.  This is a new change based on which `cython` version is automatically being installed.

gonuke:
I encountered this in DAGMC recently and didn't port it over to here.  May also require an update to the install script.

bquan0:
Right now, the `build_test.yml` workflow is running on my fork, but it runs into errors because the docker image it tries to access if from `pyne`. Should I add a run condition that restricts it to only run when the owner of the repo is `pyne`?
After trying out a bunch of variations for installing `cython` but restricting its version to be <3.0, I found that `pip install cython==0.*` was the syntax that works (since the versions before 3.0 are numbered starting with 0.). I'm not sure why `pip install cython<3.0` doesn't work, however. This is the [working run](https://github.com/bquan0/pyne/actions/runs/5772735502/job/15648016783).

gonuke:
I think you need `'cython<3'` with quotation marks since the less than symbol has a special meaning on the command line

bquan0:
Ah, makes sense. Looks like that syntax is working now! [(workflow run)](https://github.com/bquan0/pyne/actions/runs/5773523104)

gonuke:
Thanks @bquan0 - I think this looks ready to merge - let's try...

gonuke (approved):

PR #1498, opened 2023-08-09
Add workflow to create virtual machine from PyNE docker image in ghcr.io

bquan0:
## Description
This is a continuation of [this PR](https://github.com/pyne/pyne/pull/1490). The difference between the two PRs is that this PR is based on a branch from the PyNE repo, while the old PR was based on a branch from my `bquan0` forked repo. The reason we decided to move the PR to a branch on the main repo was because we needed to push the PyNE image that we build in the `virtualbox_image.yml` workflow to the ghcr.io (package space) of PyNE. 

This is the [d2vm](https://github.com/linka-cloud/d2vm) command we use to convert the PyNE image from ghcr.io into a virtual box disk image.

## Motivation and Context
This change allows users of PyNE to download a prebuilt VM with PyNE installed instead of having to build PyNE on their own computers. 

## Changes
Compared to where we left off on the old PR, the latest commits of this PR attempts to add the 3 requested things from the latest comment by @gonuke (not counting the 4th item because it's already accomplished by making this PR):
1. make the workflow only run on `release`. (Actually, it's probably not working as intended right now because [the workflow still ran](https://github.com/pyne/pyne/actions/runs/5814771722/job/15765273913) even though I added that run condition)
2. build PyNE once, not twice. I tried to do this by adding a condition in the composite action that will allow it to run the tests without building PyNE first, but it seems like the [tests don't work](https://github.com/pyne/pyne/actions/runs/5814771722/job/15765273913) if PyNE isn't built in the composite action. I'm thinking of just removing the `BuildTest` job from the workflow, but that means that the `pyne` image isn't being tested.
3. after building and testing the `pyne` image, tag it as `latest`. I added this job to `virtualbox_image.yml` and made the `virtualbox_image_build` job dependent to it. It's the same code that's found in `docker_publish.yml`

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -110,7 +110,7 @@ jobs:
           parallel: true
           tag-latest-on-default: ${{ env.USE_LATEST_TAG }}
           dockerfile: docker/ubuntu_22.04-dev.dockerfile
-          build-args: build_hdf5=${{ matrix.hdf5_build_arg }}, pyne_test_base=${{ matrix.pyne_test_base }}
+          build-args: build_hdf5=${{ matrix.hdf5_build_arg }}, pyne_test_base=${{ matrix.pyne_test_base }}, build_pyne=YES
 
   # if the previous step that tests the docker images passes then the images
   # can be copied from the ghcr where they are saved using :ci_testing tags to
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -60,7 +60,7 @@ jobs:
           quiet: false
           tag-latest-on-default: false
           dockerfile: docker/ubuntu_22.04-dev.dockerfile
-          build-args: build_hdf5=hdf5-1_12_0, build_pyne=YES
+          build-args: build_hdf5=hdf5-1_12_0
 
       - id: output_tag
         run: |
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -24,7 +24,19 @@ jobs:
   # to be built upon by the subsequent stage.
   multistage_image_build:
     runs-on: ubuntu-latest
+    outputs:
+      image_tag: ${{ steps.output_tag.outputs.image_tag }}
 
+    strategy:
+      matrix: 
+        pyne_test_base: [base_python, moab, dagmc, openmc]
+        hdf5: ['']
+        hdf5_build_arg: ['']
+        include:
+          - stage: dagmc
+            hdf5: _hdf5
+            hdf5_build_arg: hdf5-1_12_0
+      fail-fast: false
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


bquan0:
Something interesting that I noticed about having both `pull_request` and `push` as part of the run conditions of `docker_publish.yml` and `build_test.yml` is that the workflows are run twice, one because of the commit and another because of the pull request. See the [actions page](https://github.com/pyne/pyne/actions) for more details.

[This workflow run](https://github.com/pyne/pyne/actions/runs/5838378779) shows that the workflow does error out if some of the tests don't pass.

After fixing some of the syntax and other errors I ended up with [this workflow run](https://github.com/pyne/pyne/actions/runs/5838584384/job/15835807186) that has a `FileNotFoundError` for `pyne/tests/files_test_activation_responses/activation_responses_examples/alara_inp`. I looked in the repo and I found an `exp_alara_inp` file but I didn't find the `alara_inp` file. It seems like the test is comparing the `alara_inp` to `exp_alara_inp`.

One question that I have is should we include `hdf5` for the virtual box image? Currently, I have it included because it seems like a good idea to have all functionalities in the virtual machine. 

Also, should we keep the if statements with the `build_pyne` variables in the Dockerfile? It seems a bit redundant since we will want to run `nuc_data_make` if we build the pyne stages. I think it's a leftover syntax from before we made the change to use the multistage docker action ([there used to be another one of those if statements](https://github.com/bquan0/pyne/blob/composite-action/docker/ubuntu_20.04-dev.dockerfile) with `build_pyne`, but it was removed and replaced with `FROM openmc AS pyne`

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,110 @@
...
+        uses: firehed/multistage-docker-build-action@v1
+        id: build_pyne
+        with:
+          repository: ${{ env.DOCKER_IMAGE_BASENAME }}
+          stages: base_python, moab, dagmc, openmc
+          server-stage: pyne
+          quiet: false
+          tag-latest-on-default: false
+          dockerfile: docker/ubuntu_22.04-dev.dockerfile
+          build-args: build_hdf5=hdf5-1_12_0, build_pyne=YES
```
gonuke: The `repository` doesn't match for this hdf5 `build-arg`

We should probably use the one without `hdf5-1_12_0` since it doesn't build openmc in the docker publish workflow.

gonuke (changes requested):
This mismatch is causing failures.

bquan0:
It's [working](https://github.com/pyne/pyne/actions/runs/5849592341) after I removed the `hdf5` build arg from the multistage build action. I also had to change `github.actor` back to `github.repository_owner` in the `virtualbox_image_build` job in order to make it pull from the `pyne` package space instead of my package space.

gonuke:
Something strange is happening here with the use of the `DOCKER_IMAGE_TAG` across the docker publish jobs, now... 😢  I think we are making assumptions about how that image tag is formatted that is not always right, so that it fails in some cases and works in others.

gonuke:
I'm trying to understand just what conditions fail when you try to load the just-built container to perform build testing as part of the docker deployment.  Although it fails during testing because the `github.ref_name` that you are using, I think it will probably succeed when it is merged because the `github.ref_name` will be different.  The question is to try and figure out how to make it pass under both conditions.

bquan0:
Something we could do is try other `github` [variables ](https://docs.github.com/en/actions/learn-github-actions/variables)to see if there's one that will be the name of the branch for both `pull_request` and `push` events. 

I also saw your ["Cleanup docker deploy" PR](https://github.com/svalinn/DAGMC/pull/894#event-10080552779) and I think we might be able to apply a similar idea here. We could always set `tag-latest-on-default` to be true, then use the `akhilerm/tag-push-action` to apply a `testing` tag or a `final-version` tag on the image that's just built (which has the `latest` tag). This does mean we can't use a `latest` tag to specify our latest built image, but it does offer a way to apply our own tags. 

The best thing is if we could get a feature added to the multistage action where we can specify the tag name. I know you already asked this in [one of the issues](https://github.com/Firehed/multistage-docker-build-action/issues/41) on the multistage action repo, so it might be in the works? Maybe I could also try opening a PR on that repo to try to add the feature.

bquan0:
After applying the if statement changes to alter `env.DOCKER_IMAGE_TAG` if the triggering event is `pull_request` (more details can be found in the test PR that I did in the `learn-maintenance` repo), the checks from `docker_publish.yml` that were failing before are now passing. 

Edit; I just checked [one of the packages](https://github.com/pyne/pyne/pkgs/container/pyne_ubuntu_22.04_py3%2Fmoab) and it seems like those checks that were failing previously are building images with different tags, such as `refs_pull_1498_merge-bk0`, and my solution makes those checks use the images with tags that have the branch name, like `refs_heads_virtual-machine-bk0` instead of the image that the checks build. This shouldn't be a problem since the images should be the same (they're built from the same Dockerfile, same workflow, just different triggering event), but let me know if I should be testing the images that are built with this `refs_pull_1498_merge-bk0` tag instead of the ones with this `refs_heads_virtual-machine-bk0` tag.

gonuke:
The approach I took with DAGMC was to recognize that since we only want to retag things when we also want them to be tagged with `latest`, then lets set `tag-latest-on-default` as true in that case.  I think that will be simpler and won't rely on ever having to know the tag.

.github/workflows/docker_publish.yml
```diff
@@ -47,7 +44,7 @@ jobs:
           stages: base_python, moab, dagmc
           server-stage: openmc
           quiet: false
-          tag-latest-on-default: false
+          tag-latest-on-default: true
```
gonuke: I think we want this to be set conditionally depending on whether it's a PR or a push

.github/workflows/docker_publish.yml
```diff
@@ -82,7 +75,7 @@ jobs:
       fail-fast: false
 
     container:
-      image: ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3${{ matrix.hdf5 }}/${{ matrix.stage }}${{ needs.multistage_image_build.outputs.image_tag }}
+      image: ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3${{ matrix.hdf5 }}/${{ matrix.stage }}:latest
```
gonuke: maybe we can use the `output` from the multistage build to set this tag?

.github/workflows/docker_publish.yml
bquan0: This is what I'm thinking now:
- `push`: we keep it how it used to be, with `tag-latest-on-default` as false, using the `DOCKER_IMAGE_TAG` env variable as well as the step that exports it to the `output` of the first job
- `pull_request`: we set `tag-latest-on-default` to be true. But then, we would still need the step that exports `DOCKER_IMAGE_TAG` to `outputs:` because the tag for the image would change from `refs-heads...bk-0` to `latest`, and that change needs to be reflected in `container:` of the `BuildTest` job. We would also need an if statement to change `DOCKER_IMAGE_TAG` to `latest and change the value of `tag-latest-on-default`.

Overall, I'm not sure what setting `tag-latest-on-default` conditionally does that we didn't already have with the commit I did yesterday

.github/workflows/docker_publish.yml
bquan0: That's true. Looking at the [tags in the packages list](https://github.com/pyne/pyne/pkgs/container/pyne_ubuntu_22.04_py3%2Fmoab/versions) there does seem to be this `sha256:10b87d094a43f426b103cc7d69842ad57494536973db0824efa111a7049d62d5` tag that would be the output from the multistage action. However, we would still need to export to the `outputs:` of the first job in order to use it in this second job because we are using it in the `container:` section of the second job.

gonuke:
Here is another idea - let me know if I'm missing anything....

We move all the testing back to the Dockerfile and have the `pyne` stage be based on a stage that is a build-arg.
```
BUILD_ARG pyne_test_base=openmc

...

FROM ${pyne_test_base} as pyne
```

Then the `build-test` action is about running a multistage docker build with different values of `pyne_test_base`.  We may need to do some work to get the right `repository` variable for the multistage build action depending on the test base?

bquan0:
Just to clarify, the goal here is to remove the need to know what tag the multistage action attaches to the images it builds, right? I'm confused about what all these changes are trying to accomplish. 

The issue I see with this solution is that not only would we need to vary the `pyne_test_base` build arg, we would also need to vary the `stages` input of the multistage action, which would have to be accomplished either with an if statement or by having a use of the multistage action for every different dependency stage. 
Ex. for the stage `base_python` it would be `stages: base_python`, but for the stage `openmc` it would have to be `stages: base_python, moab, dagmc, openmc`
Theoretically, `stages: base_python, moab, dagmc, openmc` would work for every situation if we vary what stage `pyne` is based on using the `pyne_test_base` arg, but that would mean we pull stages from the cache that we don't use. 

The other concern that I have is that while the multistage action is fast, it still takes some time to pull the cached stages. Ex. It takes around [2 mins](https://github.com/pyne/pyne/actions/runs/5921236044/job/16053510519) to build the `base_python` stage. On the other hand, the `BuildTest` job that we currently have that uses `container:` to get the image only takes [20s](https://github.com/pyne/pyne/actions/runs/5933997757/job/16090447844) to pull the image. 

This solution also doesn't solve the goal I mentioned at the top of this comment (if that's what our goal is) because we would still have to know what the tag of the image is in order to use `akhilerm/tag-push-action@v2.1.0` to change the tag of the images.

gonuke:
The goal is to remove the need to know the tag.  The action authors are clear that folks should not rely on knowing that and that it could change.

I think the build test would use a matrix and look something like this:
```
uses: multistage-docker-build-action
with:
    repository: ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3${{ matrix.hdf5 }}
    stages: ${{ matrix.pyne_test_base }}
    server-stage: pyne
    build-args: pyne_test_base=${{ matrix.pyne_test_base }}
```

bquan0:
Now we're running into the issue where `pyne_test_base` is empty, and it can only be reimported after `FROM`, meaning we can't use it in `FROM` to specify which stage to build from. I think we need to use buildkit in order to make `ARG` variables usable in `FROM`, but I tried enabling `parallel:` in the multistage action to use buildx and it didn't end up working.

8/23 edit: Just tried this on the `learn-maintenance` repo and it does seem like enabling buildx by setting `parallel:true` in the multistage action makes the use of `ARG`s in `FROM` work (they don't show up as empty).

bquan0:
After implementing the suggested changes and fixing a bunch of silly errors (which were causing the empty `pyne_test_base` issue yesterday) I think it's now working, but we are running into this `denied: permission_denied: write_package` error again. Curiously, it seems like the ones that failed in the [push run](https://github.com/pyne/pyne/actions/runs/5953926725/job/16149395094) succeeded in the [PR run](https://github.com/pyne/pyne/actions/runs/5953927415/job/16149375114) and vice versa. 

Also, should I apply the changes to `docker_publish.yml` to `virtualbox_image.yml`?
- make all the jobs only run when `github.repository_owner` is `pyne` and when the branch is `develop`
- set `tag-latest-on-default` and `parallel` to be true for the multistage action

Should I remove the if statements with `build_pyne` in the Dockerfile? It doesn't seem like they're doing anything because we want to run `nuc_data_make` whenever we do build one of the pyne stages, and from my investigation it seems like something leftover from an old version of the Dockerfile back before we switched to using the multistage action. I mentioned this issue 2 weeks before, but it wasn't addressed then. I'm only now bringing it up again because it caused errors in the workflow and I had to add the build arg into the multistage action.

bquan0:
I made the multistage build job (the 1st one) in `docker_publish.yml` run the builds with and without `hdf5` run in parallel using a matrix as discussed in the meeting today, and also made the multistage action have the same `build-args` as the multistage action in the `BuildTest` job.

Unfortunately, I ran into the `denied: permission_denied: write_package` error again in this [workflow run](https://github.com/pyne/pyne/actions/runs/5982027434/job/16230804215), and this time it didn't work after rerunning the job. On the bright side, it seems like the `BuildTest` job is now [correctly pulling from the cache](https://github.com/pyne/pyne/actions/runs/5981976115/job/16230578578).

gonuke:
I reran this again and it worked... I'm not sure I can explain it, but I'm going to take one more look at the changes and plan to merge.

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,100 @@
...
+env:
+  VM_PASSWORD: temppwd
+  DOCKER_IMAGE_BASENAME: ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3
+
+
+jobs:
+  pyne_image_build:
+    runs-on: ubuntu-latest
+    outputs:
+      image_tag: ${{ steps.output_tag.outputs.image_tag }}
```
gonuke: Do we still need this? I don't see where it's used

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,100 @@
...
+      # before merging, change tag-latest-on-default to true
+      - name: Build PyNE docker image
+        uses: firehed/multistage-docker-build-action@v1
+        id: build_pyne
+        with:
+          repository: ${{ env.DOCKER_IMAGE_BASENAME }}
+          stages: openmc
+          server-stage: pyne
+          quiet: false
+          tag-latest-on-default: false
```
gonuke: I think we set this to `true` and remove the trigger for push and leave only the `release`

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,100 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push Image to stable img
+        uses: akhilerm/tag-push-action@v2.1.0
+        with:
+          src: ${{ env.DOCKER_IMAGE_BASENAME }}/pyne:latest
+          dst: ${{ env.DOCKER_IMAGE_BASENAME }}/pyne:stable
```
gonuke, line 58: Since this is running on release, can we get the release name and tag with that as well?

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,100 @@
...
+            udev \
+            parted \
+            e2fsprogs \
+            mount \
+            tar \
+            extlinux \
+            qemu-utils            
+          curl -sL "https://github.com/linka-cloud/d2vm/releases/download/v0.2.0/d2vm_v0.2.0_linux_amd64.tar.gz" | tar -xvz d2vm
+          sudo mv d2vm /usr/local/bin/
+          sudo d2vm convert ghcr.io/${{ github.repository_owner  }}/pyne_ubuntu_22.04_py3/pyne:latest -p ${{ env.VM_PASSWORD }} -o pyne.vdi
```
gonuke: If we can get the release number/name, let's use it here for the docker tag and also build it into the VDI file name

gonuke (changes requested):
Just a few changes as we prepare to merge for use in release only

.github/workflows/virtualbox_image.yml
bquan0: According to the [list of default variables](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables), we can use `github.ref` to get the release name if the workflow is triggered by a `release` event. I would like to test this on the `learn-maintenance` repo, but I think that would require me to open a PR with a workflow that runs on `release`, then merge my own PR and create a new release, which I don't know if it's the best practice.

.github/workflows/virtualbox_image.yml
bquan0: You're right, that's something left over from when we were using `container:` in the `BuildTest` job and had to know the tag of the image we just built. I'll remove it.

bquan0:
Just pushed a commit with all the requested changes. I also notice that this PR has a conflict with `CHANGELOG.rst` due to there being some changes that were merged into `develop`. Should I rebase this branch onto `develop` or should I just resolve the conflict with `CHANGELOG.rst`?

.github/workflows/virtualbox_image.yml
gonuke: It probably doesn't matter so much over there, but I can merge the PR and then let you play with the release

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,94 @@
...
+            qemu-utils            
+          curl -sL "https://github.com/linka-cloud/d2vm/releases/download/v0.2.0/d2vm_v0.2.0_linux_amd64.tar.gz" | tar -xvz d2vm
+          sudo mv d2vm /usr/local/bin/
+          sudo d2vm convert ghcr.io/${{ github.repository_owner  }}/pyne_ubuntu_22.04_py3/pyne:stable_${{ github.ref }} -p ${{ env.VM_PASSWORD }} -o pyne_${{ github.ref }}.vdi
+
+      - name: Upload VirtualBox image as artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: virtualbox_image
+          path: pyne.vdi
```
gonuke: this needs the name updated too with the `release`

.github/workflows/virtualbox_image.yml
```diff
@@ -0,0 +1,94 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push Image to stable img
+        uses: akhilerm/tag-push-action@v2.1.0
+        with:
+          src: ${{ env.DOCKER_IMAGE_BASENAME }}/pyne:latest
+          dst: ${{ env.DOCKER_IMAGE_BASENAME }}/pyne:stable_${{ github.ref }}
```
gonuke: It might best to have one tagged as `stable` and one tagged as `github.ref`

gonuke (changes requested):
A couple of additional comments about using the release name

.github/workflows/virtualbox_image.yml
gonuke: (once again too brief in my response) If you want to draft a PR on `learn-maintenace` , I can review/merge it and then we can see how it works - let me know

gonuke:
> Just pushed a commit with all the requested changes. I also notice that this PR has a conflict with `CHANGELOG.rst` due to there being some changes that were merged into `develop`. Should I rebase this branch onto `develop` or should I just resolve the conflict with `CHANGELOG.rst`?

It's always safest to do the rebase

.github/workflows/virtualbox_image.yml
bquan0: Just opened [a PR](https://github.com/cnerg/learn-maintenance/pull/28) in `learn-maintenance` to add a workflow that will run on `release` and echo the contents of `github.ref`. Hopefully this can be merged soon for testing!

bquan0:
Just did a rebase as well as the requested changes. I also opened [this PR](https://github.com/cnerg/learn-maintenance/pull/28) on the `learn-maintenance` repo, which adds a workflow to print out the contents of `github.ref` when the workflow is triggered by a `release` event. Hopefully that PR can be merged soon so I can test out the `github.ref` variable on releases.

bquan0:
Okay, a little bit of bad news: when a workflow is triggered by a release, `github.ref` isn't just the tag of the release (such as `v1.0.0`), it's actually something like `refs/tags/v1.0.0`. [This workflow](https://github.com/cnerg/learn-maintenance/actions/runs/6032069889/job/16366705438) on the `learn-maintenance` repo demonstrates this.
However, looking at the list of [default variables](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables), it seems like we can use `github.ref_name` because it is the tag name, which is what something like `v1.0.0` is. I didn't look at this one at first because it didn't have "release" in its description, but since we want the tag of the release I think it will work. I'd like to test this out again in `learn-maintenance` with [a PR](https://github.com/cnerg/learn-maintenance/pull/29) before making any changes here, though.

ahnaf-tahmid-chowdhury:
> Okay, a little bit of bad news: when a workflow is triggered by a release, `github.ref` isn't just the tag of the release (such as `v1.0.0`), it's actually something like `refs/tags/v1.0.0`. [This workflow](https://github.com/cnerg/learn-maintenance/actions/runs/6032069889/job/16366705438) on the `learn-maintenance` repo demonstrates this. However, looking at the list of [default variables](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables), it seems like we can use `github.ref_name` because it is the tag name, which is what something like `v1.0.0` is. I didn't look at this one at first because it didn't have "release" in its description, but since we want the tag of the release I think it will work. I'd like to test this out again in `learn-maintenance` with [a PR](https://github.com/cnerg/learn-maintenance/pull/29) before making any changes here, though.

`$GITHUB_REF_NAME` contains the tag (e.g, v1.0.0)

gonuke:
Thanks for helping figure this out @bquan0 - let's try it out

gonuke:
This looks good to go

gonuke (approved):

PR #1502, opened 2023-09-11
Move from `#define isnan` to `std::isnan`.

### LINKED ISSUE #1501: Redefinition of std::isnan results in compiler errors
**Describe the Bug**<br/>
By redefining `isnan` as `std::isnan` in `utils.h`, any further usage of `std::isnan` is translated to `std::std::isnan`, which errors. This can cause compilation to fail in libraries that require pyne, like DAGMC. It would be best to define a new internal function/macro, perhaps `pyne_isnan`.
https://github.com/pyne/pyne/blob/8c17fa9c1f2a08d0968d89f7b6bb70e773ed7cb9/src/utils.h#L30

For inspiration see MOAB's `moab_isfinite(f)`

image

pshriwise:
## Description

Resolves #1501.

Fixes a problem with the preprocessor macro in `utils.h` breaking usage of `std::isnan` in other packages b/c it is in the global namespace. 

## Motivation and Context
This was noted first [here](https://github.com/pyne/pyne/pull/836 ). And causes a problem when compiling DAGMC with recent versions of MOAB as noted here #1501.

## Changes
After conferring with @gonuke, the determined solution was to replace the `isnan` implementation selected by PyNE with `std::isnan` in the `.cpp` files where it is used.

It's worth noting that this may break compatibility with versions of GCC older than 4.0 and may break compilations on Windows using visual studio compilers from before 2015. Both sacrifices we were willing to make. 

## Changelog file

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,7 +25,7 @@ Next Version
    * Update run conditions for workflows (#1494)
    * Fix gamma intensity calculation (#1496)
    * Add workflow to build virtual machines based on Dockerfile (#1490)
-   * Replacig '#define isnan' with 'std::isnan' (#1502)
+   * Replacing '#define isnan' with 'std::isnan' (#1502)
 
 v0.7.7
 ======
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,7 @@ Next Version
    * Update run conditions for workflows (#1494)
    * Fix gamma intensity calculation (#1496)
    * Add workflow to build virtual machines based on Dockerfile (#1490)
+   * Replacig '#define isnan' with 'std::isnan' (#1502)
 
 v0.7.7
 ======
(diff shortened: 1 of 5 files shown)


CHANGELOG.rst
```diff
@@ -25,6 +25,7 @@ Next Version
    * Update run conditions for workflows (#1494)
    * Fix gamma intensity calculation (#1496)
    * Add workflow to build virtual machines based on Dockerfile (#1490)
+   * Replacig '#define isnan' with 'std::isnan' (#1502)
```
gonuke:
```suggestion
   * Replacing '#define isnan' with 'std::isnan' (#1502)
```

gonuke (changes requested):
Found a typo in the changelog (I hate it when this is the only comment...). I'll also do a search for `isnan` separately just to put a 2nd pair of eyes on it

gonuke:
I've confirmed that there are no dangling instances of `isnan`, but did fine an interesting 10 year old entry in the style guide to always use `std::isnan()`!! :)

CHANGELOG.rst
pshriwise: Thanks!!

gonuke:
Thanks @pshriwise for the quick action

gonuke (approved):
This looks good to me

PR #1504, opened 2023-09-20
update the version string

gonuke:
Update the PyNE version string in preparation for a release.  (see [DAGMC release notes](https://github.com/svalinn/DAGMC/blob/develop/release.rst))


gonuke:
Merging my own PR because it's into a branch

PR #1508, opened 2023-11-22
Install packages with Conda in Dockerfile

bquan0:
## Description
This PR adds a `conda_deps` stage to the `ubuntu_22.04-dev.dockerfile`, which provides an alternative to using `apt` and `pip` to install packages. The new stages in the dockerfile are:
- `common_base`: does some basic ubuntu setup
- `apt_deps`: installs packages via `apt` and `pip`
- `conda_deps`: installs packages via `conda`
- `base_python`: installs hdf5 if specified. 

Note: `apt_deps` and `conda_deps` are mutually exclusive stages (only one is used in a given build)

I also made the necessary changes in the `docker_publish.yml` workflow to make the multistage action build these new stages. 

These conda stage changes were mostly copied from the [Dockerfile in the Cyclus repo](https://github.com/cyclus/cyclus/blob/main/docker/Dockerfile), with the packages installed in the `conda_deps` stage modified to fit PyNE. 

## Motivation and Context
This change provides another option to install packages and is also part of the process of updating the conda-forge feedstock for PyNE. 

## Behavior
The changes seem to be working according to [this workflow run](https://github.com/bquan0/pyne/actions/runs/6960830462). However, it is possible that I missed some packages in the `conda_deps` stage since I'm not sure of the names of the conda packages that correspond to their `apt` and `pip` counterparts.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -35,7 +35,11 @@ jobs:
         stage: [base_python, moab, dagmc, openmc]
         hdf5: ['']
         include:
-          - stage: dagmc
+          - pkg_mgr: 'apt'
+            stage: dagmc
+            hdf5: _hdf5
+          - pkg_mgr: 'conda'
+            stage: dagmc
             hdf5: _hdf5
       fail-fast: false
 
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -44,7 +44,7 @@ jobs:
       fail-fast: false
 
     container:
-      image: ghcr.io/pyne/pyne_ubuntu_22.04_py3_${{ matrix.pkg_mgr }}${{ matrix.hdf5 }}/${{ matrix.stage }}:stable
+      image: ghcr.io/pyne/pyne_ubuntu_22.04_py3${{ matrix.pkg_mgr }}${{ matrix.hdf5 }}/${{ matrix.stage }}:stable
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -44,7 +44,7 @@ jobs:
       fail-fast: false
 
     container:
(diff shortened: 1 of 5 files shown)


.github/workflows/build_test.yml
```diff
@@ -31,6 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        pkg_mgr: ['apt', 'conda']
         stage: [base_python, moab, dagmc, openmc]
         hdf5: ['']
         include:
```
gonuke, line 37: Do we need to update these special includes to match?

gonuke (changes requested):
Thanks @bquan0 - Just one question on the build/test workflow

.github/workflows/build_test.yml
bquan0: Yes, because the image names now contain the name of the package manager that was used: `ghcr.io/pyne/pyne_ubuntu_22.04_py3_${{ matrix.pkg_mgr }}${{ matrix.hdf5 }}/${{ matrix.stage }}:stable`. This image name change was also made in `docker_publish.yml` so that we aren't writing to the same image name when building with `apt` vs building with `conda`.

gonuke: Sorry - I wasn't clear - I think these special includes still need to be updated to include the package manager choice, don't they?

bquan0: Yep, good catch! I checked the `build_test.yml` [workflow run](https://github.com/bquan0/pyne/actions/runs/6960830453) and it only had a `dagmc, hdf5` run instead of two runs, one with `apt` and another with `conda`. I'll make those changes. 

That also reminds me, I have to update the `virtualbox_image.yml` workflow.

gonuke (approved):
This looks good, but do we need to separate the docker/CI part from the build_test part so that the images are available to demonstrate build_test?  You can maybe PR the docker/CI part from a branch on the upstream repo and we'll go from there

PR #1509, opened 2023-11-24
Upstream branch PR: Install packages with Conda in Dockerfile

bquan0:
## Description
See #1508 for more details on all the changes. 
This PR modifies `ubuntu_22.04-dev.dockerfile` to include separate stages `apt_deps` and `conda-deps` to allow for packages to be downloaded by either the package manager apt or conda. It also makes various changes in the workflows `docker_publish.yml`, `build_test.yml`, and `virtualbox_image.yml` to accomodate those changes. 

## Motivation and Context
This is the same PR as #1508, except that it's made from a branch on the `pyne` repo instead of my forked repo. This allows the workflows to build the images in the `pyne` ghcr.io, which will allow us to see if the `build_test.yml` will pass (since `build_test.yml` pulls images from the pyne ghcr.io to test). 

## Behavior
New behavior is that the `build_test.yml` workflow should run instead of erroring out because it couldn't find the proper image from ghcr.io.

## Other Information
I noticed that in `virtualbox_image.yml`, we never make a VM from the `hdf5` images. Is this something I should look into doing?
Also, I was wondering if I should add some logic in `build_test.yml` to make it not run when the `${{ github.repository_owner }}` isn't `pyne` because the workflow only uses images from the `pyne` ghcr.io.

## Changelog file

### DIFF
--- a/.github/workflows/virtualbox_image.yml
+++ b/.github/workflows/virtualbox_image.yml
@@ -36,7 +36,7 @@ jobs:
         uses: firehed/multistage-docker-build-action@v1
         id: build_pyne
         with:
-          repository: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.pkg_mgr }}
+          repository: ${{ env.DOCKER_IMAGE_BASENAME }}_${{ matrix.pkg_mgr }}
           stages: openmc
           server-stage: pyne
           quiet: false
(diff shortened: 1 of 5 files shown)


.github/workflows/virtualbox_image.yml
```diff
@@ -16,6 +16,10 @@ env:
 jobs:
   pyne_image_build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix: 
+        pkg_mgr: ['apt', 'conda']
```
gonuke: I don't think we need both options here.  For users relying on the VM, we can pick only one option - `apt` is probably fine.

gonuke (changes requested):
Thanks for transferring this over @bquan0.  I don't think this PR should include the `build_test` workflow since they CI image won't be available for them to to succeed.  That should be a separate PR - thoughts?

bquan0:
You're right, I'll revert the changes to `build_test.yml` and apply them on a separate PR after this one is merged.

gonuke (approved):
Thanks @bquan0 - I look forward to the follow on work here and in conda-forge

PR #1510, opened 2023-11-24
add apt and conda changes to build_test.yml

bquan0:
## Description
This is a continuation of #1509. #1509 made changes to `docker_publish.yml` and the Dockerfile to make it build new images using either the package manager `apt` or `conda`. This PR makes changes to `build_test.yml` to make sure we test the new images being built. 

## Behavior
Regrettably, one of the tests isn't passing in [this workflow run](https://github.com/pyne/pyne/actions/runs/6984643775/job/19007768351) because in the `pushing_test_stable_img` job of `docker_publish.yml`, both of the special includes were `apt` instead of one `apt` and one `conda`. Therefore, that test will not pass until after this PR is merged with the typo fix in `docker_publish.yml`.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -31,20 +31,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        pkg_mgr: ['apt', 'conda']
         stage: [base_python, moab, dagmc, openmc]
         hdf5: ['']
         include:
-          - pkg_mgr: 'apt'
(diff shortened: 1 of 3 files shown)


gonuke:
I think we need the docker build fix in a separate PR still, unfortunately ☹️

bquan0:
Just reverted the changes in `build_test.yml` again, so it should be ready to be merged once the actions finish running

gonuke (approved):
Thanks!

PR #1511, opened 2023-11-25
Add apt and conda changes to build_test.yml: Part 2

bquan0:
## Description
Does what #1510 tried to do, but couldn't because there was a typo in `docker_publish.yml` in the last job where it was tagging images with `stable`. Basically, I did the special include for `apt` twice instead of doing one special include for `apt` and another for `conda`, which meant the conda + hdf5 image was never getting tagged with `stable`. [This workflow](https://github.com/pyne/pyne/actions/runs/6985660935/job/19010284234) shows that the conda + hdf5 is now indeed getting tagged with `stable`, so the `build_test.yml` workflow should work properly.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -31,15 +31,20 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        pkg_mgr: ['apt', 'conda']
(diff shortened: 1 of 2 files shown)


gonuke (approved):
Thanks for the quick follow through @bquan0

PR #1516, opened 2023-12-25
Make builds actually use conda_deps stage, fix conda_deps stage

### LINKED ISSUE #1515: conda CI build was never actually relying on conda
**Describe the Bug**<br/>

**To Reproduce**<br/>

**Expected Behavior**<br/>

bquan0:
## Description
Closes #1515. Removes the spaces in the build args for `pkg_deps` in the multistage actions in `docker_publish.yml` so that the `conda_deps` stage will actually be used when the build arg is `conda`. Modifies the `conda_deps` stage in `ubuntu_22.04-dev.dockerfile` so that it actually works when images are built with the `conda_deps` stage.

## Motivation and Context
Previously, the `conda_deps` stage wasn't being used in builds of future stages because the build arg in the multistage action was `pkg_mgr = ${ matrix.pkg_mgr}` instead of `pkg_mgr=${ matrix.pkg_mgr}`. This meant that all images were being built off the default `apt_deps` stage. Thus, it seemed like the `conda_deps` stage was working and was being used when it wasn't. 

After removing the spaces, the builds that supposedly used the `conda_deps` stage stopped working because they were now actually using the `conda_deps` stage. This was because there were many dependencies missing in that stage, like `make`, `gxx_linux-64`, `gcc_linux-64`.

## Behavior
Currently, the `conda_deps` builds do not work, and I suspect it has to do with the fact that inside the `conda_deps` stage, we do an `ENV PATH /opt/conda/bin:$PATH` which messes up the `PATH` env variable for future uses. For example, line 104:
> `RUN echo "export PATH=$HOME/.local/bin:\$PATH" >> ~/.bashrc` 

I think this because the `apt_deps` stage doesn't modify the `PATH` env variable, and because this [most recent workflow run](https://github.com/pyne/pyne/actions/runs/7318807827/job/19935890580) (no hdf5) shows it having some undefined references while this [hdf5 run](https://github.com/pyne/pyne/actions/runs/7318582630/job/19935402386) shows it not being able to find the `libh5test.a` file.  

## Other Information
The description of #1515 was a bit cryptic but I was able to figure out the problem in the end.


ahnaf-tahmid-chowdhury:
For conda pkg build, I think we can only go with one build method (apt or conda)??? Is it necessary to support both?

gonuke:
I think it is good to test with multiple packaging systems because our users may want/need different systems and this helps us be sure that they work

bquan0:
Ben suggested a [possible solution in this link](https://docs.conda.io/projects/conda-build/en/stable/resources/compiler-tools.html#an-aside-on-cmake-and-sysroots) and I tried it. Unfortunately, it didn't work, with the problem being that [the workflow](https://github.com/pyne/pyne/actions/runs/7535877846/job/20512507259) can't find Eigen3.

ahnaf-tahmid-chowdhury:
It seems strange to me because, as far as I know, Eigen3 is not a dependency of MOAB.

ahnaf-tahmid-chowdhury:
It seems Eigen is installed through conda. Perhaps MOAB is unable to detect Eigen provided by conda? I think we can directly export EIGEN3_DIR and give it a try for a temporary fix.

ahnaf-tahmid-chowdhury:
Or, we can use the conda build of MOAB and skip the source build of MOAB, which is more convenient, I suppose.

bquan0:
I tried using the conda build of MOAB but that caused an error with cmake not being able to find the MOAB directory in the `dagmc` stage. Specifically, this line `&& cmake .. -DMOAB_DIR=$HOME/opt/moab \` which I changed to `&& cmake .. -DMOAB_DIR=$HOME/opt/conda/pkgs/moab-5.5.1-nompi_notempest_py310h89735b2_100 \` since that's the moab directory in the conda `pkgs` folder in the docker image that I built on my computer.

gonuke:
> It seems strange to me because, as far as I know, Eigen3 is not a dependency of MOAB.

Eigen3 is definitely a dependency of MOAB

ahnaf-tahmid-chowdhury:
> I tried using the conda build of MOAB but that caused an error with cmake not being able to find the MOAB directory in the `dagmc` stage. Specifically, this line `&& cmake .. -DMOAB_DIR=$HOME/opt/moab \` which I changed to `&& cmake .. -DMOAB_DIR=$HOME/opt/conda/pkgs/moab-5.5.1-nompi_notempest_py310h89735b2_100 \` since that's the moab directory in the conda `pkgs` folder in the docker image that I built on my computer.

It should be MOAB_DIR=CONDA_DIR, in this case /opt/conda.
However, for the conda build we can use the conda version of DAGMC and skip the source build. In that way we will be building only PyNE from source and other comes from conda pkg.

ahnaf-tahmid-chowdhury:
> Eigen3 is definitely a dependency of MOAB

Thanks for letting me know!

Ops! I have thought Eigen3 is Embree since I was working with that last time 🙂

bquan0:
Just had a workflow run that got to the `moab` stage and didn't have the `CC, CXX, CPP` ENV variables in the `conda_deps` stage. This workflow run has the same problem as previous workflow runs that did have the ENV variables, which makes me think that the ENV variables aren't actually doing anything. For reference, the error is that the linker looks for `libstdc++.so.6` in `/opt/conda/bin/../lib/gcc/x86_64-conda-linux-gnu/13.2.0/../../../../x86_64-conda-linux-gnu/bin/ld:` but it doesn't exist there; rather, it exists in `/opt/conda/lib`.

Now I'm wondering how I can change which folder the linker looks in.

bquan0:
The "no HDF5" build [is finally working](https://github.com/pyne/pyne/actions/runs/7565239738/job/20600678838) after adding this line in the `conda_deps` stage: `ENV LD_LIBRARY_PATH /opt/conda/lib:$LD_LIBRARY_PATH`

The "HDF5" build [isn't working](https://github.com/pyne/pyne/actions/runs/7565239738/job/20600679140) because make `cannot find ./.libs/libh5test.a: No such file or directory`.

ahnaf-tahmid-chowdhury:
According to the log, it seems you haven't configured the hdf5 root path to conda. It is set to /opt/hdf5

ahnaf-tahmid-chowdhury:
Move HDF5_INSTALL_PATH near to LD_LIBRARY_PATH and configure it for both apt and conda. We may move all the source build to the apt_deps stage and the conda_deps stage may contain only conda packages. If you are not familiar with multi stage docker, please let me know.

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -65,30 +71,55 @@ RUN conda uninstall -y conda-libmamba-solver
...
     mamba install -y --force-reinstall libsqlite && \
     conda clean -y --all
 RUN mkdir -p `python -m site --user-site`
 
+ENV CC /opt/conda/bin/x86_64-conda_cos6-linux-gnu-gcc
+ENV CXX /opt/conda/bin/x86_64-conda_cos6-linux-gnu-g++
+ENV CPP /opt/conda/bin/x86_64-conda_cos6-linux-gnu-cpp
+
+ENV LD_LIBRARY_PATH /opt/conda/lib:$LD_LIBRARY_PATH
+ENV HDF5_INSTALL_PATH=/opt/conda/lib
```
ahnaf-tahmid-chowdhury:
```suggestion
ENV HDF5_INSTALL_PATH=/opt/conda
```

bquan0:
Looks like the conda + hdf5 build is now working. I will fix the apt + hdf5 build with the next commit and also clean up the dockerfile so that it will be ready to be merged.

bquan0:
[All of the builds](https://github.com/pyne/pyne/actions/runs/7574049093) are now working!

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -65,30 +74,52 @@ RUN conda uninstall -y conda-libmamba-solver
...
 RUN mkdir -p $HOME/opt
 RUN echo "export PATH=$HOME/.local/bin:\$PATH" >> ~/.bashrc
 
 # build HDF5
 ARG build_hdf5="NO"
-ENV HDF5_INSTALL_PATH=$HOME/opt/hdf5/$build_hdf5
-RUN if [ "$build_hdf5" != "NO" ]; then \
+ARG pkg_mgr=apt
+# don't build hdf5 if we use conda (we already installed it)
+RUN if [ "$build_hdf5" != "NO" ] && [ "$pkg_mgr" = "apt" ]; then \
```
ahnaf-tahmid-chowdhury: I think we may move this to the apt_deps stage. Thus, there is no need to use extra conditions.

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -8,9 +8,13 @@ FROM ubuntu:${ubuntu_version} AS common_base
 # Ubuntu Setup
 ENV TZ=America/Chicago
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+ENV HOME /root
```
ahnaf-tahmid-chowdhury, line 11: Docker generally comes with root user privilege. I think it is not necessary to use extra env.

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -8,9 +8,13 @@ FROM ubuntu:${ubuntu_version} AS common_base
...
 
 FROM common_base AS apt_deps
-ENV HOME /root
+# set HDF5 path for base_python stage
+ARG build_hdf5="NO"
+ENV HDF5_INSTALL_PATH=$HOME/opt/hdf5/$build_hdf5
```
ahnaf-tahmid-chowdhury, line 16: As mentioned earlier, $HOME is not necessary.

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -65,30 +74,52 @@ RUN conda uninstall -y conda-libmamba-solver
...
+                gcc_linux-64 \
                 git \
+                cmake \
+                gfortran \ 
                 libblas \
                 liblapack \
+                eigen \
                 hdf5 \
-                setuptools \
+                numpy==1.23 \
```
ahnaf-tahmid-chowdhury: I think we can go with the latest version of numpy as #1485 fixed that issue

ahnaf-tahmid-chowdhury (changes requested):
Thank you very much for these improvements @bquan0! 🙌🏾

Just a few tidying comments.

docker/ubuntu_22.04-dev.dockerfile
bquan0: I think it's still necessary to specify the ENV variable HOME since it's used as a variable in a lot of other lines in the dockerfile.

docker/ubuntu_22.04-dev.dockerfile
bquan0: I think it's necessary because the `/opt` directory is different from the `/root/opt` directory, which is what `$HOME/opt` expands to.

ahnaf-tahmid-chowdhury: We can ignore the extra path /root (not necessary).
In linux we use /opt for 3rd party apps. I think installing hdf5 in /opt is fine. And the same method goes for others.

docker/ubuntu_22.04-dev.dockerfile
bquan0: We could do that, but then there would be almost nothing left in the `base_python` stage, which means that we should probably remove it. However, that would require some more edits in the `docker_publish.yml` file to accomodate the new stages. Overall, I think there needs to be more discussion on this topic, and maybe we can also consider splitting the dockerfile into two dockerfiles (one for conda, one for apt).

ahnaf-tahmid-chowdhury: @gonuke, we are expecting your valuable suggestions on this.

ahnaf-tahmid-chowdhury: My opinion, employing a single Dockerfile with a multi-stage build is a viable approach. It may necessitate a redesign of the `docker_publish.yml`. Additionally, I've observed that we are using the ubuntu base image for each build, resulting in larger image sizes. To optimize this, we can create `apt_deps` from the debian image, as we do not need snap support. For `conda_deps`, we can initiate the build from scratch or use [miniconda3](https://hub.docker.com/r/continuumio/miniconda3).

ahnaf-tahmid-chowdhury: @gonuke, what is your opinion on this?

gonuke: Let's stick with the Ubuntu as the base image since it also confirms that we can build on that image.

docker/ubuntu_22.04-dev.dockerfile
gonuke: Let's stick with this for now since we have historically installed in `/root/opt` rather than `/opt` to have some control over what's there.

docker/ubuntu_22.04-dev.dockerfile
gonuke: Let's keep `/root` for now since it's been part of our CI strategy for a while - try not to change too many things at once

gonuke:
This seems nearly ready, once we relax the numpy restriction

docker/ubuntu_22.04-dev.dockerfile
gonuke: Let me know when this had been updated @bquan0 and I think we're ready to merge

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -65,30 +74,52 @@ RUN conda uninstall -y conda-libmamba-solver
...
 # make starting directory
 RUN mkdir -p $HOME/opt
 RUN echo "export PATH=$HOME/.local/bin:\$PATH" >> ~/.bashrc
 
 # build HDF5
 ARG build_hdf5="NO"
-ENV HDF5_INSTALL_PATH=$HOME/opt/hdf5/$build_hdf5
-RUN if [ "$build_hdf5" != "NO" ]; then \
+ARG pkg_mgr=apt
+# don't build hdf5 if we use conda (we already installed it)
```
gonuke: We install HDF5 in the `apt` path already, too.  I think we do want to build in both cases.  However, when we choose not to build (`build_hdf5 = NO`) then it fails, which is a different problem.  It looks like it's using HDF 1.14 when we install via conda - not sure we've tested PyNE with that before.

gonuke (changes requested):
This is failing on conda builds with what I think is a new HDF5.

docker/ubuntu_22.04-dev.dockerfile
bquan0: Guess I'll just limit it to HDF5 version 1.13 for conda builds then? And also make it build HDF5 for both apt and conda.

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -65,29 +74,50 @@ RUN conda uninstall -y conda-libmamba-solver
...
+                gcc_linux-64 \
                 git \
+                cmake \
+                gfortran \ 
                 libblas \
                 liblapack \
-                hdf5 \
-                setuptools \
+                eigen \
+                hdf5<1.14 \
```
ahnaf-tahmid-chowdhury:
```suggestion
                "hdf5<1.14" \
```

ahnaf-tahmid-chowdhury (changes requested):

docker/ubuntu_22.04-dev.dockerfile
ahnaf-tahmid-chowdhury: HDF5 and MOAB are dependencies of PyNE to enable all functionalities. When building a conda package, it is advisable to rely on the conda versions of HDF5 and MOAB. This ensures that when a user installs PyNE through conda, HDF5 and MOAB will be automatically installed for them.

docker/ubuntu_22.04-dev.dockerfile
ahnaf-tahmid-chowdhury: May be this is not the robust. We should use `$build_hdf5` instead. But, it seems, it is not well configured for Conda build.

gonuke:
I wonder if we should undo/revert all the conda build stuff and approach it more carefully, rather than trying to patch this.

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -65,29 +74,50 @@ RUN conda uninstall -y conda-libmamba-solver
...
     mamba install -y --force-reinstall libsqlite && \
     conda clean -y --all
 RUN mkdir -p `python -m site --user-site`
 
+ENV CC /opt/conda/bin/x86_64-conda_cos6-linux-gnu-gcc
+ENV CXX /opt/conda/bin/x86_64-conda_cos6-linux-gnu-g++
+ENV CPP /opt/conda/bin/x86_64-conda_cos6-linux-gnu-cpp
+
+# put conda on the path
+ENV LD_LIBRARY_PATH /opt/conda/lib:$LD_LIBRARY_PATH
```
ahnaf-tahmid-chowdhury, line 108: The error arises because LD_LIBRARY_PATH has been defined to the conda directory, causing issues with the source build. To address this, try unsetting LD_LIBRARY_PATH. However, there's no need to install hdf5 twice for the conda build method.

gonuke:
I propose that we withdraw this in favor of two separate actions with two separate Dockerfiles that build under apt and conda, respectively

ahnaf-tahmid-chowdhury:
@bquan0, what is your plan for this? Would you like to modify/close it and go with @gonuke's proposal? Or, I could take care of it with a new PR.

bquan0:
@ahnaf-tahmid-chowdhury I will close this PR and open a new PR that applies the change that @gonuke proposed.

PR #1526, opened 2024-02-27
Update FindMOAB to use MOAB config file

ahnaf-tahmid-chowdhury:
## Description

This PR updates the `FindMOAB.cmake` file to utilize the MOAB config file for obtaining the `MOAB_INCLUDE_DIRS` variable. Previously, this variable was configured manually, causing issues with newer versions of MOAB. 

## Changes Made

- Updated `FindMOAB.cmake` to use MOAB config file for `MOAB_INCLUDE_DIRS`.
- Removed manual configuration of `MOAB_INCLUDE_DIRS` to prevent conflicts with newer MOAB versions.
- Added env dir in .gitignore file.


ahnaf-tahmid-chowdhury:
@gonuke, could you please take a look at this? This PR should be merged unless we encounter issues installing PyNE with the latest MOAB.

CMakeLists.txt
```diff
@@ -103,17 +103,7 @@ if(WITH_MOAB)
...
-  if(${MOAB_FOUND})
-    include_directories(${MOAB_INCLUDE_DIRS})
-    get_filename_component(MOAB_LIBRARY_DIRS ${MOAB_LIBRARY} PATH)
-    if(APPLE)
-      link_directories(${MOAB_LIBRARY_DIRS})
-    else(APPLE)
-      link_directories(${MOAB_LIBRARY_DIRS})
-    endif(APPLE)
-  endif(${MOAB_FOUND})
+  find_package(MOAB REQUIRED)
```
gonuke, line 105: MOAB is not required for PyNE

CMakeLists.txt
```diff
@@ -124,8 +114,7 @@ if(WITH_DAGMC)
       set(DAGMC_ROOT "${DEPS_ROOT_DIR}")
     endif()
     MESSAGE("-- DAGMC Root: ${DAGMC_ROOT}")
-    find_package(DAGMC)
-    message("-- DAGMC Found: ${DAGMC_FOUND}")
+    find_package(DAGMC REQUIRED)
```
gonuke, line 121: DAGMC is not required for PyNE

CMakeLists.txt
```diff
@@ -103,17 +103,7 @@ if(WITH_MOAB)
     set(MOAB_ROOT "${DEPS_ROOT_DIR}")
   endif()
   MESSAGE("-- MOAB Root: ${MOAB_ROOT}")
-  find_package(MOAB)
-  message("-- MOAB Found: ${MOAB_FOUND}")
```
gonuke, line 107: I think some kind of message about what is found in MOAB would be useful, unless the MOAB cmake file does that?  Like the version string, perhaps?

CMakeLists.txt
```diff
@@ -124,8 +114,7 @@ if(WITH_DAGMC)
       set(DAGMC_ROOT "${DEPS_ROOT_DIR}")
     endif()
     MESSAGE("-- DAGMC Root: ${DAGMC_ROOT}")
-    find_package(DAGMC)
-    message("-- DAGMC Found: ${DAGMC_FOUND}")
```
gonuke, line 128: Some sort of message about what is found for DAGMC would be useful

gonuke (changes requested):
I have a few changes here about what's required.  It's technically possible to build PyNE without MOAB and DAGMC and many folks may want to do that if they are only interested in the more fundamental capabilities.  This is also reflected in our CI testing process.

CMakeLists.txt
ahnaf-tahmid-chowdhury: Yes, I know. I have noticed there is a statement available `WITH_MOAB`. So, MOAB is only required when any user sets this value to `ON`. Please note, by default this is already set to `ON` in the configuration file. 

I think we should make this required here, thus user can gets error message if they have turned `ON` `WITH_MOAB` and they do not have MOAB installed. Please, let me know should I still remove this or not.

Another note: it seems that a pip package of MOAB will soon be available. I am currently adjusting different configurations to find MOAB in the [pyne/scikit-build-core](https://github.com/pyne/pyne/tree/scikit-build-core) branch. Once all the tests are passing, I will make a PR for the big update.

CMakeLists.txt
ahnaf-tahmid-chowdhury: Ack

CMakeLists.txt
ahnaf-tahmid-chowdhury: Same logic here too, a variable `WITH_DAGMC` is available. Let me know.

CMakeLists.txt
ahnaf-tahmid-chowdhury: Ah, I see. For the current latest release, the MOAB config file does not come with version strings, since my PR was recently merged to the develop branch. It will then be available in the master branch. So, we need to wait for a new release of MOAB, like 5.6. For now, I think I can add some message strings to the FindMOAB file.

CMakeLists.txt
ahnaf-tahmid-chowdhury: I noticed that, FindDAGMC comes with some logs. However, let me know if you still want to add the logs here as well.

CMakeLists.txt
gonuke: I understand why to make this required now

CMakeLists.txt
gonuke: I think we just want a message about what version of DAGMC is found

cmake/FindMOAB.cmake
gonuke, line 1: Can we just delete this now and rely on MOAB to install the right info?  It may require changing some variables later in the file to match the ones MOAB sets

gonuke (changes requested):
I think we can remove this custom `FindMOAB` - see #1537

CMakeLists.txt
ahnaf-tahmid-chowdhury: Updating FindDAGMC file.

cmake/FindMOAB.cmake
ahnaf-tahmid-chowdhury: I think it is good practice to have a custom find method. This will let the user set a custom path for MOAB, for example, in the /opt or /home directory or anywhere else. Additionally, in the updated FindMOAB, I am using the MOAB config file to set all the necessary variables.

ahnaf-tahmid-chowdhury: Also added some custom messages for more information. Let me know if you are still interested to remove this file.

gonuke: I disagree with the idea of a custom `FindMOAB` as it frequently can become out of date with the software it is trying to find.  Relying on the files installed by the package is the most robust and CMake's search rules (https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure) still allow a user to specify a custom location.

ahnaf-tahmid-chowdhury: I think we are also ready to remove FindDAGMC as well?

gonuke: I'm not sure about that one because I'm not sure if we are installing the appropriate file with DAGMC yet... that's a new issue for us to tackle in DAGMC

ahnaf-tahmid-chowdhury: I have dropped FindDAGMC file and it seems tests are passing.

gonuke: I think I'd like to test this with DAGMC installed in an unusual location and confirm that the build process still works/finds it.  Perhaps it does because if we set `DAGMC_ROOT` in the build process, the build in CMake `FindPackage` macro will recognize that variable and use it as a hint?

gonuke: I think I'd like to test this with DAGMC installed in an unusual location and confirm that the build process still works/finds it.  Perhaps it does because if we set `DAGMC_ROOT` in the build process, the build in CMake `FindPackage` macro will recognize that variable and use it as a hint?

ahnaf-tahmid-chowdhury:
```make
-- HDF5 Root: /root/opt/hdf5/hdf5-1_14_3
-- Found HDF5: /root/opt/hdf5/hdf5-1_14_3/lib/libhdf5.so;/root/opt/hdf5/hdf5-1_14_3/lib/libhdf5_hl.so;/root/opt/hdf5/hdf5-1_14_3/lib/libhdf5.so;/usr/lib/x86_64-linux-gnu/libz.so;/usr/lib/x86_64-linux-gnu/libdl.a;/usr/lib/x86_64-linux-gnu/libm.so (found version "1.14.3")  
--    HDF5 Include directory: /root/opt/hdf5/hdf5-1_14_3/include
--    HDF5 Library directories: /root/opt/hdf5/hdf5-1_14_3/lib
--    HDF5 Libraries: /root/opt/hdf5/hdf5-1_14_3/lib/libhdf5.so;/usr/lib/x86_64-linux-gnu/libz.so;/usr/lib/x86_64-linux-gnu/libdl.a;/usr/lib/x86_64-linux-gnu/libm.so
--    HDF5 library version: 1.14.3
-- MOAB library Version: 
-- MOAB Include Directory: /root/opt/moab/include;/usr/include/eigen3
-- MOAB Library Directories: /root/opt/moab/lib
-- MOAB Libraries: -L/root/opt/moab/lib -lMOAB /root/opt/hdf5/hdf5-1_14_3/lib/libhdf5_hl.so;/root/opt/hdf5/hdf5-1_14_3/lib/libhdf5.so;/usr/lib/x86_64-linux-gnu/libdl.a;/usr/lib/x86_64-linux-gnu/libz.so;/usr/lib/x86_64-linux-gnu/libm.so
-- Configuring DAGMC 3.2.3
-- DAGMC library version: 3.2.3
-- DAGMC Include directory: /root/opt/dagmc/include;/root/opt/moab/include;/usr/include/eigen3
-- DAGMC Library directories: /root/opt/dagmc/lib;/root/opt/moab/lib
-- DAGMC Libraries: MOAB;dagmc
```
We have already installed different libraries in a custom location: `/root/*`.

gonuke (approved):
This should fix a few problems and clean up our build maintenance burden - thanks @ahnaf-tahmid-chowdhury

PR #1527, opened 2024-03-12
Add separate workflow and dockerfile for conda build

### LINKED ISSUE #1515: conda CI build was never actually relying on conda

bquan0:
## Description
This is a continuation of the closed PR #1516. In the dockerfile of that PR, we had one stage for installing dependencies via apt and one stage for installing dependencies via conda. However, we decided that it would be best to split up the apt build and the conda build for a few reasons
- the builds weren't working.
- When we had apt and conda in one dockerfile, we weren't able to take advantage of the fact that we could install `moab`, `dagmc`, and `openmc` via conda. This structural difference means that it's better to have 2 separate workflows and dockerfiles. 

In this PR, I have reverted `ubuntu_22.04-dev.dockerfile` and `docker_publish.yml` to what they were 7 months ago, right before any of the conda changes were made. The only change I made was to remove the `cython` restriction in the dockerfile. 

I also added a new `conda_build.dockerfile` and `conda_docker.yml` to build and test PyNE by installing packages via conda. It's still a work in progress and I'm open to any feedback about how I can change and improve it. 

## Motivation and Context
This PR will be the new PR to close #1515. 

## Changes
This should be a fix to the current situation on develop where conda builds weren't working properly at all. This should also be a better way to implement what #1516 was trying to accomplish. 

## Behavior
Current behavior is that we try to build PyNE via conda dependencies or apt dependencies in the same workflow/dockerfile. New behavior is that there will be separate workflows/dockerfiles.

### DIFF
--- a/.github/workflows/conda_docker.yml
+++ b/.github/workflows/conda_docker.yml
@@ -1,0 +1,138 @@
+name: Build & Publish docker image for PyNE-CI
+
+on:
+  # allows us to run workflows manually
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - 'docker/*'
+      - '.github/workflows/conda_docker.yml'
+      - '.github/actions/build-test/action.yml'
+  push:
+    paths:
+      - 'docker/*'
+      - '.github/workflows/conda_docker.yml'
+      - '.github/actions/build-test/action.yml'
+
+env:
+  DOCKER_IMAGE_BASENAME: ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3_conda
+  USE_LATEST_TAG: false
+
+jobs:
+  # builds and pushes docker images of various stages to ghcr.
+  # These docker images are also stored in ghcr and can be pulled
+  # to be built upon by the subsequent stage.
+  multistage_image_build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix: 
+        hdf5: ['']
+        hdf5_build_arg: ['NO']
+        include:
+          - hdf5: _hdf5
+            hdf5_build_arg: hdf5-1_12_0
+      fail-fast: false
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      
+      - name: Tag images with latest if on the main repo's develop branch
+        if: github.repository_owner == 'pyne' && github.ref_name == 'develop'
+        run: echo "USE_LATEST_TAG=true" >> $GITHUB_ENV
+
+      # build base python, moab, dagmc, openmc using multistage docker build action
+      - uses: firehed/multistage-docker-build-action@v1
+        id: build_all_stages
+        with:
+          repository: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}
+          stages: base_python, moab, dagmc
+          server-stage: openmc
+          quiet: false
+          parallel: true
+          tag-latest-on-default: ${{ env.USE_LATEST_TAG }}
+          dockerfile: docker/conda_build.dockerfile
+          build-args: build_hdf5=${{ matrix.hdf5_build_arg }}, pyne_test_base=openmc
+
+  # Downloads the images uploaded to ghcr in previous stages and runs pyne
+  # tests to check that they work.
+  BuildTest:
+    needs: [multistage_image_build]
+    runs-on: ubuntu-latest
+    
+    strategy:
+      matrix: 
+        pyne_test_base: [base_python, moab, dagmc, openmc]
+        hdf5: ['']
+        hdf5_build_arg: ['NO']
+        include:
+          - pyne_test_base: dagmc
+            hdf5: _hdf5
+            hdf5_build_arg: hdf5-1_12_0
+      fail-fast: false
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      
+      - name: Tag images with latest if on the main repo's develop branch
+        if: github.repository_owner == 'pyne' && github.ref_name == 'develop'
+        run: echo "USE_LATEST_TAG=true" >> $GITHUB_ENV
+
+      # build test stage and pyne-dev stage using multistage docker build action
+      - uses: firehed/multistage-docker-build-action@v1
+        id: multistage_build_and_test
+        with:
+          repository: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}
+          stages: ${{ matrix.pyne_test_base }}
+          server-stage: pyne-dev
+          quiet: false
+          parallel: true
+          tag-latest-on-default: ${{ env.USE_LATEST_TAG }}
+          dockerfile: docker/conda_build.dockerfile
+          build-args: build_hdf5=${{ matrix.hdf5_build_arg }}, pyne_test_base=${{ matrix.pyne_test_base }}
+
+  # if the previous step that tests the docker images passes, then the images
+  # can be copied from the ghcr where they are saved using :ci_testing tags to
+  # :latest and :stable tags.
+  pushing_test_stable_img:
+    if: github.repository_owner == 'pyne' && github.ref_name == 'develop'
+    needs: [BuildTest]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        stage: [base_python, moab, dagmc, openmc]
+        hdf5: ['']
+        include:
+          - stage: dagmc
+            hdf5: _hdf5
+
+    name: "ghcr.io/${{ github.repository_owner }}/pyne_ubuntu_22.04_py3_conda${{ matrix.hdf5 }}/${{ matrix.stage }}: latest -> stable"
+
+    steps:
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Push Image to stable img
+        uses: akhilerm/tag-push-action@v2.1.0
+        with:
+          src: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:latest
+          dst: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}/${{ matrix.stage }}:stable
--- a/.github/workflows/conda_docker.yml
+++ b/.github/workflows/conda_docker.yml
@@ -53,7 +53,7 @@ jobs:
         id: build_all_stages
         with:
           repository: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}
-          stages: base_python, moab, dagmc
+          stages: base_conda, moab, dagmc
           server-stage: openmc
           quiet: false
           parallel: true
@@ -69,7 +69,7 @@ jobs:
     
     strategy:
       matrix: 
-        pyne_test_base: [base_python, moab, dagmc, openmc]
+        pyne_test_base: [base_conda, moab, dagmc, openmc]
         hdf5: ['']
         hdf5_build_arg: ['NO']
         include:
@@ -115,7 +115,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        stage: [base_python, moab, dagmc, openmc]
+        stage: [base_conda, moab, dagmc, openmc]
         hdf5: ['']
         include:
           - stage: dagmc
--- a/.github/workflows/conda_docker.yml
+++ b/.github/workflows/conda_docker.yml
@@ -93,13 +93,13 @@ jobs:
         if: github.repository_owner == 'pyne' && github.ref_name == 'develop'
         run: echo "USE_LATEST_TAG=true" >> $GITHUB_ENV
 
-      # build test stage and pyne-dev stage using multistage docker build action
+      # build test stage and pyne stage using multistage docker build action
       - uses: firehed/multistage-docker-build-action@v1
         id: multistage_build_and_test
         with:
           repository: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}
           stages: ${{ matrix.pyne_test_base }}
-          server-stage: pyne-dev
+          server-stage: pyne
           quiet: false
           parallel: true
           tag-latest-on-default: ${{ env.USE_LATEST_TAG }}
--- a/.github/workflows/conda_docker.yml
+++ b/.github/workflows/conda_docker.yml
@@ -1,4 +1,4 @@
-name: Build & Publish docker image for PyNE-CI
+name: Build & Publish Conda docker image for PyNE-CI
 
 on:
   # allows us to run workflows manually
--- a/docker/conda_build.dockerfile
+++ b/docker/conda_build.dockerfile
@@ -9,8 +9,7 @@ ENV TZ=America/Chicago
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 ENV HOME /root
 
-# set default python version to 3.10
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 10
+
 
 RUN apt-get update \
     && apt-get install -y --fix-missing \
@@ -25,6 +24,10 @@ RUN echo 'export PATH=/opt/conda/bin:$PATH' > /etc/profile.d/conda.sh && \
     rm ~/miniconda.sh
 
 ENV PATH /opt/conda/bin:$PATH
+
+# install python 3.10 and set it as default
+RUN conda install python=3.10
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 10
 
 RUN conda config --add channels conda-forge
 RUN conda update -n base -c defaults conda
--- a/docker/conda_build.dockerfile
+++ b/docker/conda_build.dockerfile
@@ -25,10 +25,8 @@ RUN echo 'export PATH=/opt/conda/bin:$PATH' > /etc/profile.d/conda.sh && \
 
 ENV PATH /opt/conda/bin:$PATH
 
-# install python 3.10 and set it as default
+# install python 3.10 so we have imp python module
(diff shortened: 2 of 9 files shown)


docker/conda_build.dockerfile
```diff
@@ -0,0 +1,91 @@
...
+                && \
+    mamba install -y --force-reinstall libsqlite && \
+    conda clean -y --all
+RUN mkdir -p `python -m site --user-site`
+
+# make starting directory
+RUN mkdir -p $HOME/opt
+RUN echo "export PATH=$HOME/.local/bin:\$PATH" >> ~/.bashrc
+
+# build HDF5
```
gonuke: Skip the HDF5 build in conda also - it should come with MOAB

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,91 @@
...
+RUN conda install -y mamba
+RUN conda uninstall -y conda-libmamba-solver
+RUN conda config --set solver classic
+RUN conda update -y --all && \
+    mamba install -y \
+                cmake \
+                git \
+                libblas \
+                liblapack \
+                hdf5 \
```
gonuke: Try without these?

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,91 @@
...
+
+FROM moab AS dagmc
+RUN conda install conda-forge::dagmc
+
+FROM dagmc AS openmc
+RUN conda install conda-forge::openmc
+
+# Build/Install PyNE from release branch
+FROM ${pyne_test_base} AS pyne
+RUN conda install conda-forge::pyne
```
gonuke: We do need to build PyNE  - the test is to see if we can build it against these dependencies

gonuke (changes requested):
Thanks for kicking this off

ahnaf-tahmid-chowdhury:
I have a suggestion to make. Recently, while working on the DAGMC CI, I found its implementation to be quite well. Would it be possible for us to apply a similar method here as well? It could make it easier for us to understand all the projects as they would be configured in the same way. In the case of PyNE, we may need to add a few steps to make it compatible with the CI. What do you think, @gonuke? Do you agree, or do you have a better alternative in mind?

gonuke:
I think we first need to get some version of this working, and then we can look into improvements.  The PyNE build process is different from the DAGMC one for a few reasons and we need to make sure that our CI is meeting our overall needs.

.github/workflows/conda_docker.yml
```diff
@@ -0,0 +1,138 @@
...
+
+  # Downloads the images uploaded to ghcr in previous stages and runs pyne
+  # tests to check that they work.
+  BuildTest:
+    needs: [multistage_image_build]
+    runs-on: ubuntu-latest
+    
+    strategy:
+      matrix: 
+        pyne_test_base: [base_python, moab, dagmc, openmc]
```
ahnaf-tahmid-chowdhury:
```suggestion
        pyne_test_base: [base_conda, moab, dagmc, openmc]
```

.github/workflows/conda_docker.yml
```diff
@@ -0,0 +1,138 @@
...
+      - name: Tag images with latest if on the main repo's develop branch
+        if: github.repository_owner == 'pyne' && github.ref_name == 'develop'
+        run: echo "USE_LATEST_TAG=true" >> $GITHUB_ENV
+
+      # build base python, moab, dagmc, openmc using multistage docker build action
+      - uses: firehed/multistage-docker-build-action@v1
+        id: build_all_stages
+        with:
+          repository: ${{ env.DOCKER_IMAGE_BASENAME }}${{ matrix.hdf5 }}
+          stages: base_python, moab, dagmc
```
ahnaf-tahmid-chowdhury:
```suggestion
          stages: base_conda, moab, dagmc
```

.github/workflows/conda_docker.yml
```diff
@@ -0,0 +1,138 @@
...
+  # if the previous step that tests the docker images passes, then the images
+  # can be copied from the ghcr where they are saved using :ci_testing tags to
+  # :latest and :stable tags.
+  pushing_test_stable_img:
+    if: github.repository_owner == 'pyne' && github.ref_name == 'develop'
+    needs: [BuildTest]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        stage: [base_python, moab, dagmc, openmc]
```
ahnaf-tahmid-chowdhury:
```suggestion
        stage: [base_conda, moab, dagmc, openmc]
```

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -31,7 +29,7 @@ RUN apt-get update \
     pip install --upgrade pip; \
     pip install numpy==1.23 \
```
ahnaf-tahmid-chowdhury:
```suggestion
    pip install numpy \
```

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,91 @@
...
+ENV LIBRARY_PATH $HDF5_INSTALL_PATH/lib:$LIBRARY_PATH
+
+FROM base_conda AS moab
+RUN conda install conda-forge::moab
+
+FROM moab AS dagmc
+RUN conda install conda-forge::dagmc
+
+FROM dagmc AS openmc
+RUN conda install conda-forge::openmc
```
ahnaf-tahmid-chowdhury: I think, we do not need to create extra stages here.

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,91 @@
...
+                jinja2 \
+                cython \
+                && \
+    mamba install -y --force-reinstall libsqlite && \
+    conda clean -y --all
+RUN mkdir -p `python -m site --user-site`
+
+# make starting directory
+RUN mkdir -p $HOME/opt
+RUN echo "export PATH=$HOME/.local/bin:\$PATH" >> ~/.bashrc
```
ahnaf-tahmid-chowdhury, line 72: Not required I suppose. We may set pyne install prefix to conda dir.

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,91 @@
...
+        && cd hdf5 \
+        && ./configure --prefix=$HDF5_INSTALL_PATH --enable-shared \
+        && make -j 3 \
+        && make install \
+        && cd .. \
+        && rm -rf hdf5; \
+    fi
+# put HDF5 on the path
+ENV LD_LIBRARY_PATH $HDF5_INSTALL_PATH/lib:$LD_LIBRARY_PATH
+ENV LIBRARY_PATH $HDF5_INSTALL_PATH/lib:$LIBRARY_PATH
```
ahnaf-tahmid-chowdhury: Please set this to conda dir after removing hdf5.

ahnaf-tahmid-chowdhury (changes requested):

docker/conda_build.dockerfile
bquan0: I think it's necessary because the original dockerfile had these stages for the purpose of testing the install of various combinations of dependencies before building PyNE, and I think that's something we want to preserve for the conda build dockerfile too.

ahnaf-tahmid-chowdhury: I see, you are right!

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,81 @@
...
+
+FROM dagmc AS openmc
+RUN conda install conda-forge::openmc
+
+# Build/Install PyNE from develop branch
+FROM ${pyne_test_base} AS pyne-dev
+
+RUN export PYNE_HDF5_ARGS="" ;\
+    && cd $HOME/opt \
+    && git clone -b develop --single-branch https://github.com/pyne/pyne.git \
```
ahnaf-tahmid-chowdhury: Please use COPY command here and copy the source to the dockerfile.

ahnaf-tahmid-chowdhury (changes requested):

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,81 @@
...
+RUN conda config --set solver classic
+RUN conda update -y --all && \
+    mamba install -y \
+                cmake \
+                git \
+                setuptools \
+                pytest \
+                pytables \
+                jinja2 \
+                cython \
```
ahnaf-tahmid-chowdhury:
```suggestion
                "cython<3" \
```

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -29,9 +27,9 @@ RUN apt-get update \
     update-alternatives --install /usr/bin/python python /usr/bin/python3 10; \
     update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 10; \
     pip install --upgrade pip; \
-    pip install numpy==1.23 \
+    pip install numpy \
             scipy \
-            'cython<3' \
+            cython \
```
ahnaf-tahmid-chowdhury:
```suggestion
            "cython<3" \
```

ahnaf-tahmid-chowdhury (changes requested):
MOAB v5.3 is old and doesn't support Cython v3. We have a separate PR #1528 for this. Please consider using Cython v0.29 for now.

gonuke:
The limitation isn't from MOAB.  Conda-forge has MOAB 5.5.1

ahnaf-tahmid-chowdhury:
Unfortunately, we still need to consider staying with Cython 0.29 for this current PR, as PyNE has issues with Cython 3. This may be resolved through a separate PR #1528.

docker/conda_build.dockerfile
bquan0: To apply this change, I have removed the `pyne-dev` stage. Now, we will build the existing `pyne` stage in `docker_publish.yml` since that uses the COPY command.

ahnaf-tahmid-chowdhury:
It seems, the action workflows, share similarities. I'm curious if we could potentially use a single action file to select different Docker files, if feasible. I'd like to pass this decision over to @gonuke.

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,85 @@
...
+# Build/Install PyNE from release branch
+FROM ${pyne_test_base} AS pyne
+ARG build_hdf5
+
+RUN export PYNE_HDF5_ARGS="" ;\
+    if [ "$build_hdf5" != "NO" ]; then \
+            export PYNE_HDF5_ARGS="--hdf5 $HDF5_INSTALL_PATH" ; \
+    fi;
+COPY . $HOME/opt/pyne
+RUN cd $HOME/opt/pyne \
```
gonuke, line 93: I'm not sure I understand why, but the `COPY` command seems to be resolving `$HOME/opt/pyne` as `/opt/pyne` and the `RUN` command is resolving it as `/root/opt/pyne` and therefor the `cd` is failing.

gonuke (changes requested):
This is failing tests because of the inconsistent resolution of `$HOME`

docker/conda_build.dockerfile
ahnaf-tahmid-chowdhury: It is because, we need to run `mkdir -p $HOME/opt/pyne` at first to make the dir available.

docker/conda_build.dockerfile
ahnaf-tahmid-chowdhury: Here is the issue. This should be `RUN mkdir -p $HOME/opt/pyne`

docker/conda_build.dockerfile
gonuke: I'm not sure that's the problem?  If you look at the build logs, the `COPY` resolves this as `/opt/pyne`.  I don't expect that making the dir first will result in this resolving differently.

ahnaf-tahmid-chowdhury: OPS!! That's strange! If RUN can recognize $HOME, then COPY should as well.

ahnaf-tahmid-chowdhury: I think $HOME env is not configured in this file.

We should set it by `ENV HOME=/root`?

bquan0: Yeah I think that's what's missing.

bquan0:
After running into a bunch of errors (which were fixed by applying the solutions I previously found in the `conda-build-fix` branch that I forgot about), I'm now running into a "file not found" [error](https://github.com/pyne/pyne/actions/runs/8445269120/job/23132361554) when running the tests, which is strange. I did a quick vimdiff between the conda setup that I currently have and the conda setup in `conda-build-fix` but there weren't any differences other than order of dependencies and not having `hdf5<1.14` (since having hdf5 breaks the build).

gonuke:
I think something changed with how pytest handles a `setup` function.  @bennibbelink seemed to notice this with Cyclus a few days ago and made a simple fix there that avoided the issue rather than corrected it....

The missing file is downloaded in a `setup` method that maybe is not being run by pytest?

ahnaf-tahmid-chowdhury:
I'm not sure if this is the problem. I've been working on [scikit-build-core](https://github.com/pyne/pyne/tree/scikit-build-core) and have fixed all the test issues; now all tests are passing. However, I've noticed some UnicodeDecodeError instances and resolved the use of the setup function in two files. @bquan0, could you please try running `pytest -ra` in the `tests` folder locally and see what happens?

bquan0:
I noticed that the tests were also not working for the original `docker_publish.yml` where we install dependencies with apt. I think this has to do with the `pytest` version since they deprecated the `setup()` and `teardown()` functions in [version 7.2.0](https://docs.pytest.org/en/stable/changelog.html#pytest-7-2-0-2022-10-23). From some initial testing on my laptop with the pytest 8.1.1, it seems like we need to rename these functions into `setup_function()` and `teardown_function()`.

bquan0:
The change to the `setup()` function seems to have fixed a few of the errors. Also, the most recent commit isn't necessary, I just added it because I thought the `setup()` function still wasn't running after the rename to `setup_function()`. In actuality, it was running, but the failed workflow made me think otherwise. 

The error that's causing the workflow to fail now has to do with the `pyne.ace.ascii_to_binary()` function. 

The file that is missing is the `C12-binary.ace` file created by `pyne.ace.ascii_to_binary()` from `C012-n.ace` (file imported in the `setup()` function which now actually runs).

tests/test_ace.py
```diff
@@ -6,8 +6,8 @@
 
 import pyne.ace
 
-
-def setup():
+@pytest.fixture(autouse=True)
+def setup_function():
```
gonuke: I think `setup_module()` will recover the intended behavior and don't use the fixture:
```suggestion
def setup_module():
```

tests/test_ace.py
```diff
@@ -102,6 +102,6 @@ def test_read_c12_binary():
     assert table.reactions[2].sigma[-1] == 1.00772
 
 
-def teardown():
+def teardown_function():
```
gonuke:
```suggestion
def teardown_module():
```

gonuke (changes requested):
I suggested the change to `setup_module()` in once place and it should probably propagate to the others.

Our previous usage of `setup()` assumed that it was called once per module, so we want to ensure that same behavior now, and not once per function.  I think (speculation) that using `setup_function()` and `teardown_function()` results in teardown happening after each test, thus removing a file generated in one test that is needed for another test.  (this practice may have other flaws)

bquan0:
After applying that change and removing the `dec` import in `test_chainsolve.py` (since it is [deprecated in newer numpy versions](https://numpy.org/doc/1.18/reference/generated/numpy.testing.dec.deprecated.html) and isn't used in the file), the first BuildTest (only `base_python`) is passing. 

The other BuildTests error out in `test_activation_responses.py` in the `test_response_hdf5_to_mesh()` test due to an [error](https://github.com/pyne/pyne/actions/runs/8482325900/job/23241455281) with creating a mesh from PyMOAB, which I don't know how to fix.

gonuke:
Somethings to try in `test_mesh.py:test_structured_mesh_from_coords()`

1. force ranges to be lists in `assert_array_equal`
2. loop over elements in list:
```
for obs, exp in zip(sm.structured_coords, [range(1, 5), range(1, 4), range(1, 3)]):
    assert_array_equal(obs, exp)
```

gonuke:
For conda builds, let's pin the moab version to 5.3.0

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,101 @@
...
+
+# put conda on the path
+ENV LD_LIBRARY_PATH /opt/conda/lib:$LD_LIBRARY_PATH
+
+# make starting directory
+RUN mkdir -p $HOME/opt
+RUN echo "export PATH=$HOME/.local/bin:\$PATH" >> ~/.bashrc
+
+FROM base_conda AS moab
+RUN conda install conda-forge::moab
```
gonuke: Try to pin this to v5.3.0 to see if it fixes the moab issue:
something like:
```suggestion
RUN conda install "conda-forge::moab==5.3.0"
```

bquan0:
[All of the tests](https://github.com/pyne/pyne/actions/runs/8485311737?pr=1527) in `docker_publish.yml` are now passing except for the dagmc + hdf5 test. It's failing in `test_activation_responses.py` in the `test_activation_responses_script()` function due to a FileNotFoundError for `alara_inp`. It seems like a mismatch in hdf5 version might be the cause.

ahnaf-tahmid-chowdhury:
Please consider pinning HDF5 to 1.10 not <1.14

bquan0:
Big news: the original `docker_publish.yml` is [now working](https://github.com/pyne/pyne/actions/runs/8486995288/job/23254255408?pr=1527)! 
Also, I propose that the github workflows for each dockerfile only runs when its respective dockerfile changes (and not when anything in the `docker/` folder changes). 

Also, I just realized that the 10 in `update-alternatives --install /usr/bin/python python /usr/bin/python3 10` refers to the priority of the install, not the version of python. So, I have pinned the python version in conda to `<3.10` because moab 5.3.0 doesn't work with python 3.10.

bquan0:
After pinning `"python<3.10"`, the moab install works.
After updating conda and using mamba to install dagmc and openmc, those installations also work.

The moab tests are now passing. However, the [dagmc and openmc tests are failing](https://github.com/pyne/pyne/actions/runs/8488059561/job/23256659946) due to the same `ValueError: Cannot create cython.array from NULL pointer` in `test_response_hdf5_to_mesh()` of `test_activation_responses.py`  that we had with the moab test of the conda build previously (which we solved by pinning the moab version to 5.3.0).

gonuke:
> The moab tests are now passing. However, the [dagmc and openmc tests are failing](https://github.com/pyne/pyne/actions/runs/8488059561/job/23256659946) due to the same `ValueError: Cannot create cython.array from NULL pointer` in `test_response_hdf5_to_mesh()` of `test_activation_responses.py` that we had with the moab test of the conda build previously (which we solved by pinning the moab version to 5.3.0).

This may be because the DAGMC conda package is out of date.  I'll have to see to that.

ahnaf-tahmid-chowdhury:
I think it's not a good idea to set Python < 3.10 since ubuntu 22.04 comes with Python 3.10.6 which is old now. I will check this on the scikit-build-core branch and try to resolve that there.

bquan0:
Closing this PR since other merged PRs have already accomplished its goal.

PR #1525, opened 2024-02-19
Support Cython3+

bennibbelink:
## Description
This PR includes changes to support the use of Cython3+ in building PyNE. 

## Motivation and Context
It removes the pin of `cython<3` in the Dockerfile, closes #1522 

## Changes
Dependency update

## Behavior
Current behavior: PyNE build fails on the Cython compilation of `enrichment.pyx`, `material_library.pyx`, and `cram.pyx` when using Cython3+
New behavior: PyNE build succeeds using Cython3+

## Other Information
This updates the MOAB version used in `docker/ubuntu_22.04-dev.dockerfile` to a 5.5.1, a version that supports Cython3+.  Also, I created this branch from `conda-build-fix`

PR #1528, opened 2024-03-15
Support Cython3+ (new)

### LINKED ISSUE #1522: Update to Cython 3+
With API changes in Cython, PyNE no longer functions with version 3.0 of Cython or higher.

This was identified in #1520, and has been seen in other projects.

This issue will be closed when PyNE can by built with newer versions of Cython.

bennibbelink:
## Description
This PR includes changes to support the use of Cython3+ in building PyNE.  Replaces #1525 

## Motivation and Context
It removes the pin of `cython<3` in the Dockerfile, closes #1522 

## Changes
Dependency update

## Behavior
Current behavior: PyNE build fails on the Cython compilation of `enrichment.pyx`, `material_library.pyx`, and `cram.pyx` when using Cython3+
New behavior: PyNE build succeeds using Cython3+

## Other Information
This updates the MOAB version used in `docker/ubuntu_22.04-dev.dockerfile` to a 5.5.1, a version that supports Cython3+.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Next Version
 ============
 **Change**
    * Test install script on multiple Ubuntu versions (#1484)
+   * Support Cython3+ (#1528)
 
 **Fix**
    * Fix Type Mismatch Error in PyNE's ENSDF Processing Module (#1519)
--- a/docker/ubuntu_22.04-dev.dockerfile
+++ b/docker/ubuntu_22.04-dev.dockerfile
@@ -31,7 +31,7 @@ RUN apt-get update \
     pip install --upgrade pip; \
     pip install numpy==1.23 \
             scipy \
-            'cython<3' \
+            cython \
             nose \
             pytest \
             tables \
@@ -74,7 +74,7 @@ RUN conda update -y --all && \
                 pytest \
                 pytables \
                 jinja2 \
-                "cython<3" \
+                cython \
                 && \
     mamba install -y --force-reinstall libsqlite && \
     conda clean -y --all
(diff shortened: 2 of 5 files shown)


ahnaf-tahmid-chowdhury:
I think you may have missed moving all of your commits here.

bennibbelink:
I think my commits are all here.  I think there is a problem in `docker/ubuntu_22.04-dev.dockerfile` at line 189, the `develop` branch of `pyne` is being checked out and built instead of our local version.  Since the dependencies no longer pin `cython<3` the `develop` branch will fail.  

@bquan0 is this checkout intentional or a remnant from copy/paste?  Unless I'm missing something it seems like we should be `COPY`ing in the local version of `pyne` to the docker image instead.

bquan0:
Seems like it was something that was done in the [original](https://github.com/pyne/pyne/blob/ee445910c15c38ea30eac39f1c4fae0e78401383/docker/ubuntu_20.04-dev.dockerfile) `ubuntu_20.04-dev.dockerfile` before I started working on it, which is why it was kept all this time. I'll have to ask @gonuke about this tomorrow at the meeting.

ahnaf-tahmid-chowdhury:
I believe previously, the setup involved running PyNE from a Docker image where we were installing PyNE using Git. Later, it was transitioned to CI for testing, with the `pyne-dev` stage still intact. Subsequently, a new stage named `pyne` was created to copy files from the source bellow the `pyne-dev` stage. From then on, we've been utilizing the `pyne` stage in our action workflow. However, a mistake occurred when someone accidentally changed the stage back to `pyne-dev` or added this stage for building and since this comes before `pyne` stage, resulting in an issue. My suggestion is that we no longer need the `pyne-dev` stage anymore.

PR #1529, opened 2024-04-11
Update current apt build

bquan0:
## Description
This PR updates the versions of OpenMC, MOAB, and hdf5 that are built in `docker_publish.yml` and `ubuntu_22.04-dev.dockerfile`. It also removes the build without hdf5 so that we always build hdf5. It also doesn't download the apt dependencies `libhdf5-dev` and `hdf5-tools` so that they come directly from the hdf5 build and are the most up-to-date version. 

## Motivation and Context
There are new versions of MOAB and OPENMC that we should use. Also, linux is slow to update hdf5, so we should always build it so we have the most recent version. 

## Behavior
Previously, we had one build with hdf5 and one build without. Now, we will always build with hdf5. This means removing a bunch of the matrices in `docker_publish.yml` used for having an hdf5 build and a no-hdf5 build. I also removed the build arg that allows `docker_publish.yml` to specify which version of hdf5 to build. Now, it's just hardcoded into the dockerfile in the hdf5 build section. 

The version changes are:
- hdf5 v1.12.0 --> v1.14.3
- MOAB v5.3.0 --> v5.5.1
- OpenMC v0.13.0 --> v0.14.0


docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -169,43 +118,17 @@ ENV PYNE_DAGMC_ARGS "--dagmc $HOME/opt/dagmc"
...
-
 # Build/Install PyNE from release branch
 FROM ${pyne_test_base} AS pyne
 ARG build_hdf5
 
-RUN export PYNE_HDF5_ARGS="" ;\
-    if [ "$build_hdf5" != "NO" ]; then \
-            export PYNE_HDF5_ARGS="--hdf5 $HDF5_INSTALL_PATH" ; \
-    fi;
+RUN export PYNE_HDF5_ARGS="--hdf5 $HDF5_INSTALL_PATH"
```
gonuke: This shell does not persist over different `RUN` commands

gonuke: Either combine it with the `RUN` command that launches setup.py or use `ENV`. The former is probably preferred.

gonuke (changes requested):
I think this comment will fix your failed runs.

bquan0:
@gonuke now that we are using OpenMC v0.14.0, we need to regenerate the statepoint file. (see [forum post](https://openmc.discourse.group/t/oserror-statepoint-file-has-a-version-of-17-0-which-is-not-consistent-with-the-version-expected-by-openmc-18/2304)) How should I go about this? This is currently causing the OpenMC file to fail. 

Also, we had to go back to installing MOAB v5.3.0 because v5.5.0 uses Cython3, and we don't have the Cython3 compatibility fix in PyNE yet. (This was causing the MOAB, DAGMC, and OpenMC tests to fail). We will have to update MOAB in the same (future) PR where we update to Cython3 in PyNE.

gonuke:
What's the prospect of updating everything to Cython3 as part of this process, rather than the other way around?

ahnaf-tahmid-chowdhury:
> What's the prospect of updating everything to Cython3 as part of this process, rather than the other way around?

Sure, we can. Only thing we need to do is merging [cython3-upgrade](https://github.com/pyne/pyne/tree/cython3-upgrade) branch into our current setup.

P.S. Apologies for my recent inactivity; I'm currently on leave.

bquan0:
I rebased my branch onto the [cython3-update](https://github.com/pyne/pyne/tree/cython3-upgrade) branch. 

There were tests that were failing in `test_material.py` related to how [Cython3 updates operator overloading](https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#arithmetic-methods) to be more like Python. I have fixed them by adding `radd` and `rmul` in `material.pyx`. Also, I would like to note that there is no `rdiv` test. 

As for the moab, dagmc, openmc, tests, they were failing due to the same `ValueError: Cannot create cython.array from NULL pointer` in `test_response_hdf5_to_mesh()` that we've seen quite a few times now.

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -41,84 +36,38 @@ RUN apt-get update \
...
-        && ./configure --prefix=$HDF5_INSTALL_PATH --enable-shared \
-        && make -j 3 \
-        && make install \
-        && cd .. \
-        && rm -rf hdf5; \
-    fi
+RUN cd $HOME/opt \
+    && mkdir hdf5 \
+    && cd hdf5 \
+    && git clone --single-branch --branch hdf5-1_14_3 https://github.com/HDFGroup/hdf5.git \
```
shimwell: shouild this line make use of the ARG build_hdf5 value
```suggestion
    && git clone --single-branch --branch $build_hdf5 https://github.com/HDFGroup/hdf5.git \
```

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -41,84 +36,38 @@ RUN apt-get update \
...
-    then \
-            export MOAB_HDF5_ARGS="-DHDF5_ROOT=$HDF5_INSTALL_PATH"; \
-    fi \
-    && cd $HOME/opt \
+RUN export MOAB_HDF5_ARGS="-DHDF5_ROOT=$HDF5_INSTALL_PATH"; \
+    cd $HOME/opt \
     && mkdir moab \
     && cd moab \
-    && git clone --depth 1 --single-branch -b 5.3.0 https://bitbucket.org/fathomteam/moab \
+    && git clone --depth 1 --single-branch -b 5.5.1 https://bitbucket.org/fathomteam/moab \
```
shimwell: we could make the moab version a ARG which defaults to 5.5.1

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -169,45 +118,19 @@ ENV PYNE_DAGMC_ARGS "--dagmc $HOME/opt/dagmc"
...
 ARG build_hdf5
 # build/install OpenMC Python API
-RUN if [ "$build_hdf5" != "NO" ]; then \
-            export HDF5_ROOT="$HDF5_INSTALL_PATH" ; \
-    fi ;\
+RUN export HDF5_ROOT="$HDF5_INSTALL_PATH" ; \
     git clone https://github.com/openmc-dev/openmc.git $HOME/opt/openmc \
     && cd  $HOME/opt/openmc \
-    && git checkout tags/v0.13.0 \
+    && git checkout tags/v0.14.0 \
```
shimwell: We could make the HDF5 version an ARG which defaults to 0.14.0

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -169,45 +118,19 @@ ENV PYNE_DAGMC_ARGS "--dagmc $HOME/opt/dagmc"
 FROM dagmc AS openmc
 ARG build_hdf5
 # build/install OpenMC Python API
-RUN if [ "$build_hdf5" != "NO" ]; then \
-            export HDF5_ROOT="$HDF5_INSTALL_PATH" ; \
-    fi ;\
+RUN export HDF5_ROOT="$HDF5_INSTALL_PATH" ; \
     git clone https://github.com/openmc-dev/openmc.git $HOME/opt/openmc \
```
shimwell: we can clone the specific tag release instead of cloning the repo and checking out the tag

gonuke:
I've issued a [PR in MOAB](https://bitbucket.org/fathomteam/moab/pull-requests/682) to hopefully fix this problem.

docker/ubuntu_22.04-dev.dockerfile
shimwell: ```git clone --depth 1 --branch ${openmc_version} https://github.com/openmc-dev/openmc.git $HOME/opt/openmc \```

gonuke:
Alternatively, we may need to `try` and `catch` an exception at line 434 in `pyne/mesh.py`.  It seems that this method will throw an exception if the default value is None/NULL.  I'm not sure why we aren't seeing that instead of this error.

gonuke:
> Alternatively, we may need to `try` and `catch` an exception at line 434 in `pyne/mesh.py`. It seems that this method will throw an exception if the default value is None/NULL. I'm not sure why we aren't seeing that instead of this error.

But first... let's update to the newest MOAB (5.5.1) to see if/how this error presents itself.

bquan0:
But we were already building and testing moab 5.5.1, ever since [this commit](https://github.com/pyne/pyne/pull/1529/commits/3557e3638fa94da46cff881644540ddd2ba78df9). I'll test the `try` `catch` solution you mentioned

bquan0:
Seems like the `try except` has fixed the `moab` and `dagmc` tests. The `openmc` tests are still failing because the statepoint file needs to be regenerated.

shimwell:
I see the 5 year old statepoint file on the repo but I don't see any scripts that generated it. I also looked for (but couldn't see) a summary.h5 file in case it was present as I think that contains the openmc model so could e helpful to regenerate the statepoint.

gonuke:
> But we were already building and testing moab 5.5.1, ever since [this commit](https://github.com/pyne/pyne/pull/1529/commits/3557e3638fa94da46cff881644540ddd2ba78df9). I'll test the `try` `catch` solution you mentioned

Sorry - my bad - I was looking at the comment by @shimwell and misinterpreted it as suggested change.

pyne/mesh.py
```diff
@@ -431,7 +431,11 @@ def __init__(
             self.size = self.tag.get_length()
             self.dtype = self.tag.get_dtype()
             self.pymbtype = types.pymoab_data_type(self.dtype)
-            self.default = self.tag.get_default_value()
+            try: 
+                self.default = self.tag.get_default_value()
+            except ValueError:
+                self.default = int
```
gonuke: I think this should be `None`

gonuke (changes requested):
Clarification on value of mesh tag `default` when `ValueErrror`

gonuke:
> I see the 5 year old statepoint file on the repo but I don't see any scripts that generated it. I also looked for (but couldn't see) a summary.h5 file in case it was present as I think that contains the openmc model so could e helpful to regenerate the statepoint.

We may need to work backwards from the things that are tested in `tests/test_openmc_utils.py`:
Tally ID=1
* [ ] has `ebounds = [0.0, 1.0, 2.0] * 1e6`
* [ ] is a structured mesh tally of size 3x2x1 on domain (-40, -12.5, -2.5) to (80, 25, 5) that scores `flux`
* [ ] the values of the results don't really matter since the test uses `openmc` directly to evaluate the tally and compares to PyNE's evaluation

I think those are the only things that are true.

We should make an openmc python script that will generate something like this and add it to the repo.  That sounds a like something that @shimwell would be great at :)

shimwell:
> > I see the 5 year old statepoint file on the repo but I don't see any scripts that generated it. I also looked for (but couldn't see) a summary.h5 file in case it was present as I think that contains the openmc model so could e helpful to regenerate the statepoint.
> 
> We may need to work backwards from the things that are tested in `tests/test_openmc_utils.py`: Tally ID=1
> 
> * [ ]  has `ebounds = [0.0, 1.0, 2.0] * 1e6`
> * [ ]  is a structured mesh tally of size 3x2x1 on domain (-40, -12.5, -2.5) to (80, 25, 5) that scores `flux`
> * [ ]  the values of the results don't really matter since the test uses `openmc` directly to evaluate the tally and compares to PyNE's evaluation
> 
> I think those are the only things that are true.
> 
> We should make an openmc python script that will generate something like this and add it to the repo. That sounds a like something that @shimwell would be great at :)

I have had a go at this in PR https://github.com/pyne/pyne/pull/1533

bquan0:
After making these 2 changes to `generate_statepoint.py`:
- change the upper bounds of the rectangular prism from (80, 25, 5), to (40, 12.5, 5) to match what the tests were expecting. This is because (80, 25, 5) are added to the lower bounds of (-40, -12.5, -2.5).
- change 2e6 in `energy_bins` to 2e7 because the test expects 20e6 and not 2e6.

and regenerating the statepoint file, the `openmc` test is passing. There are 4 `openmc` BuildTests which are not passing but that's to be expected since they are testing on images built with version 0.13.0, and we have updated to version 0.14.0

@gonuke I think this is ready to be merged now since all the relevant tests are passing.

gonuke:
Just to summarize my understanding of why the remaining `openmc` build tests are failing, and for posterity:
* this PR is largely designed to update the CI image which normally would not run build tests on old images
* because of the Cython updates, it also includes code changes which cause the build tests to run, and run on old images
* the new images work with the newer OpenMC statepoint file
* we expect the build tests to pass once the CI images have been updated

gonuke (approved):

docker/ubuntu_22.04-dev.dockerfile
gonuke: This is an openmc version, but yes, your point is well-taken

gonuke (changes requested):
Please incorporate @shimwell 's suggestions of using more build-args w/ default values for the Dockerfile, and cloning single branches.

bquan0:
@gonuke I just applied @shimwell's suggestions, though I'm not sure if they are necessary. The build-args for `moab` and `openmc` are each used in one spot for the dockerfile, and right now we aren't using the github workflow to build images with different versions of `moab` and `openmc`. If we want to in the future, this change could be added then, but as of right now it's doing nothing.

I also updated `build_test.yml` and its composite action `action.yml` to test the new images that we are building.

gonuke (approved):

gonuke:
Thanks @bquan0 - let's go forward with this and see how it all goes once the images are updated

PR #1532, opened 2024-04-26
add script to generate statepoint file, update statepoint filename test_openmc_utils.py

Edgar-21:
## Description
This PR adds a script called `generate_statepoint.py` which will run build a model and generate a statepoint file, in the correct directory for `test_openmc_utils.py`. It also updates `test_openmc_utils.py` to use the default statepoint filename as outputted by openmc

## Changes
Hopefully will make the tests run


bquan0:
lgtm, thanks @Edgar-21 !

PR #1540, opened 2024-05-31
Separate dockerfile and workflow for conda build [PART 2]

### LINKED ISSUE #1515: conda CI build was never actually relying on conda

bquan0:
## Description
This is a continuation of #1527, which was discontinued because we decided that we needed to update the versions of OpenMC, MOAB, and HDF5 for the apt build. This time, for the Conda build, we will not have separate stages for `moab, dagmc, openmc`, so there will only exist one image with all those dependencies. 

## Motivation and Context
This allows people to build PyNE using Conda, which is an alternative to using apt. 
This PR will also be the new PR to close #1515.

## Changes
This PR adds `conda_build.dockerfile`, which lists all the steps for building PyNE with Conda. It also adds `conda_docker.yml`, which runs the multistage build action on the dockerfile to build PyNE images with Conda. 

## Other Information
I'm not sure if the multistage action is relevant anymore because we only have one stage before the `pyne` stage (this is because we build everything but `pyne` in the first job, then build `pyne` in the BuildTest job). I guess it's still necessary for caching purposes in GHCR?


docker/conda_build.dockerfile
```diff
@@ -0,0 +1,95 @@
...
+# install OpenMC
+RUN mamba install "conda-forge::openmc=0.14.0"
+
+# Build/Install PyNE from release branch
+FROM openmc AS pyne
+
+COPY . $HOME/opt/pyne
+RUN cd $HOME/opt/pyne \
+    && python setup.py install --user \
+                                $PYNE_MOAB_ARGS $PYNE_DAGMC_ARGS \
```
gonuke, line 89: These need to be set to look for MOAB & DAGMC

gonuke (changes requested):
This won't fix your errors, but will be important

bquan0:
I fixed the `UnicodeDecodeError` in the conda build by changing the Python version from 3.11 --> 3.10 because the apt build uses 3.10. I then re-added the change to `test_mesh.py` from the previous conda PR to fix another error. Now, both the conda and apt build are failing in the `openmc` stage due to a `FileNotFoundError`.

ahnaf-tahmid-chowdhury:
It seems `alara_inp` is missing in the test dir.

gonuke:
We need to update the OpenMC statepoint file in `tests/files_test_r2s/r2s_examples/openmc_r2s`.

Short term: update the statepoint version string in the H5 file and see if it works

Long term: create an OpenMC input that generates that statepoint file - or some replacement

gonuke:
@zxkjack123 - any chance you have the OpenMC input that generated the statepoint file for testing R2S? (see more above). We need to regenerate it with a newer version of OpenMC.

bquan0:
I tried the short term solution of editing the statepoint file's version using `h5py` in a python command line. Looks like it worked! 

There was another error for the conda build in `test_chainsolve.py` where `dec` couldn't be imported from `numpy.testing` because it was deprecated. I was thinking about downgrading the version of `numpy` for the conda build, but I looked at `test_chainsolve.py` and realized that it didn't even use `dec`, so I removed it. 

Looks like all tests are passing now.

tests/files_test_r2s/r2s_examples/openmc_r2s/statepoint.nps5e5.ves8.ebins175.h5
ahnaf-tahmid-chowdhury, line 1: Is this file required, or is it just generated during each test run?

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,98 @@
...
+                                --clean -j 3;
+ENV PATH $HOME/.local/bin:$PATH
+RUN cd $HOME \
+    && nuc_data_make \
+    && cd $HOME/opt/pyne/tests \
+    && ./ci-run-tests.sh python3
```
ahnaf-tahmid-chowdhury, line 95: I think we can use `pytest -ra` command here which comes with better summery format.

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,98 @@
...
+                numpy \
+                scipy \
+                nose \
+                matplotlib \
+                git \
+                setuptools \
+                pytest \
+                pytables \
+                jinja2 \
+                "cython<3" \
```
ahnaf-tahmid-chowdhury:
```suggestion
                cython \
```

tests/test_mesh.py
```diff
@@ -130,7 +130,8 @@ def test_structured_mesh_from_coords():
         structured_coords=[range(1, 5), range(1, 4), range(1, 3)], structured=True
     )
     assert sm.dims == [0, 0, 0, 3, 2, 1]
-    assert_array_equal(sm.structured_coords, [range(1, 5), range(1, 4), range(1, 3)])
+    for obs, exp in zip(sm.structured_coords, [range(1, 5), range(1, 4), range(1, 3)]):
+        assert_array_equal(obs, exp)
```
ahnaf-tahmid-chowdhury: I guess if you value detailed error reporting, your approach is good to go. However, for simplicity, I am suggesting this, though it's not mandatory.
```suggestion
    assert sm.structured_coords == [range(1, 5), range(1, 4), range(1, 3)]
```

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,98 @@
...
+                gcc_linux-64 \
+                cmake \
+                make \
+                gfortran \
+                libblas \
+                liblapack \
+                eigen \
+                numpy \
+                scipy \
+                nose \
```
ahnaf-tahmid-chowdhury: I think we have shifted to pytest as nose is old.
```suggestion
```

ahnaf-tahmid-chowdhury (changes requested):
There are a few changes I would like to request.

tests/files_test_r2s/r2s_examples/openmc_r2s/statepoint.nps5e5.ves8.ebins175.h5
gonuke: It is required. We use it to test our ability to process OpenMC output for R2S calculations.

We edited the binary file slightly to accommodate changing OpenMC statepoint file versioning, but need to find a way to include an OpenMC input to be able to regenerate this in the future.

docker/conda_build.dockerfile
gonuke: Good catch - I think we're all updated for Cython 3 now, right?

.github/workflows/conda_docker.yml
```diff
@@ -0,0 +1,113 @@
...
+  # if the previous step that tests the docker images passes, then the images
+  # can be copied from the ghcr where they are saved using :ci_testing tags to
+  # :latest and :stable tags.
+  pushing_test_stable_img:
+    if: github.repository_owner == 'pyne' && github.ref_name == 'develop'
+    needs: [BuildTest]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        stage: [base_conda, moab, dagmc, openmc]
```
gonuke: We don't have all these stages anymore

docker/conda_build.dockerfile
ahnaf-tahmid-chowdhury: Yap

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,98 @@
+ARG ubuntu_version=22.04
+
+FROM ubuntu:${ubuntu_version} AS openmc
```
gonuke: Let's call this stage `pyne-deps`.  `openmc` made sense when we had other stages before it, built now we only build this stage as the dependencies

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,98 @@
...
+                git \
+                setuptools \
+                pytest \
+                pytables \
+                jinja2 \
+                "cython<3" \
+                future \
+                progress \
+                && \
+    mamba install -y --force-reinstall libsqlite && \
```
gonuke: Do we know that this is still necessary?

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,98 @@
...
+
+# install MOAB
+RUN conda install "conda-forge::moab=5.5.1"
+
+# install DAGMC
+RUN conda update -n base conda
+RUN mamba install conda-forge::dagmc
+
+# install OpenMC
+RUN mamba install "conda-forge::openmc=0.14.0"
```
gonuke: Do we need to specify a version?

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,98 @@
...
+
+ENV CC /opt/conda/bin/x86_64-conda_cos6-linux-gnu-gcc
+ENV CXX /opt/conda/bin/x86_64-conda_cos6-linux-gnu-g++
+ENV CPP /opt/conda/bin/x86_64-conda_cos6-linux-gnu-cpp
+
+# put conda on the path
+ENV LD_LIBRARY_PATH /opt/conda/lib:$LD_LIBRARY_PATH
+
+# make starting directory
+RUN mkdir -p $HOME/opt
```
gonuke: We don't use this until later, let's do it only when we need it. It helps make it clearer why we need it

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,98 @@
...
+RUN echo "export PATH=$HOME/.local/bin:\$PATH" >> ~/.bashrc
```
gonuke: This doesn't need to be set until we install PyNE.  Move it to where you do the same `ENV`

docker/conda_build.dockerfile
```diff
@@ -0,0 +1,98 @@
...
+RUN mamba install conda-forge::dagmc
+
+# install OpenMC
+RUN mamba install "conda-forge::openmc=0.14.0"
+
+# Build/Install PyNE from release branch
+FROM openmc AS pyne
+
+ENV PYNE_MOAB_ARGS "--moab $HOME/opt/moab"
+ENV PYNE_DAGMC_ARGS "--dagmc $HOME/opt/dagmc"
```
gonuke: I don't think conda installs these in this location.  These are locations we use for manual install in the `apt` version, but conda will put these elsewhere.  You might not need to specify a path?

docker/conda_build.dockerfile
gonuke: let's leave changes on how we do testing to another PR

gonuke (changes requested):
Thanks @bquan0 - I'm glad all the tests pass now.  After these changes, we'll need to look carefully through the test logs to make sure we're actually testing what we think and not getting false positives (which has been a problem in the past)

docker/conda_build.dockerfile
bquan0: Originally, I didn't have these args, but in a previous comment you told me to set these args so that the install looks for the MOAB and DAGMC. I can try removing them and also changing them to the respective conda directory.

From what I can tell,  the tests for DAGMC and MOAB still run regardless of whether I set them or not, but I'm not 100% sure.

docker/conda_build.dockerfile
bquan0: I will remove it and see if it still works.
Update: the tests passed, so we don't need it.

docker/conda_build.dockerfile
bquan0: I'll try it without specifying the version.
Update: the tests passed, so we don't need to specify a version

bquan0:
I've applied the suggestions except for the one about the MOAB and DAGMC build args for PyNE. @gonuke let me know what I should try for that. 
All tests are passing after the changes.

docker/conda_build.dockerfile
gonuke: I think you need these args, but either with no path or a different correct path.  I think no path should be ok

bquan0: I made that change and the tests are still passing (though I'm not sure if the DAGMC and MOAB tests are running)

gonuke:
I can see that this finds MOAB and DAGMC during build and also runs relevant tests, so I think it's all good

gonuke (approved):

ahnaf-tahmid-chowdhury (approved):
Thanks, @bquan0.

gonuke:
@bquan0 - this failed on merge :( with a segfault in Python????

gonuke:
https://github.com/pyne/pyne/actions/runs/9520877255/job/26247815999

ahnaf-tahmid-chowdhury:
Important note: [Numpy v2 will be released on Jun 16 2024](https://numpy.org/)

PR #1543, opened 2024-06-24
Remove imp module from tests to support Python 3.12

### LINKED ISSUE #1542: ModuleNotFoundError: No module named 'imp'
I am using Python 3.12, and I am getting the following error during installation.

  File "/Users/gamalakabani/Codes/PyNE/pyne/setup.py", line 38, in <module>
    import imp
ModuleNotFoundError: No module named 'imp'

It seems that Python version 3.12  no longer supports "imp". It is deprecated.

Also, please update the conda installation environment for easy of installation and use.

bquan0:
## Description
Removes the `imp` module from 2 tests where they are not used (`setup.py` and `setup_sub.py`) and replaces its usage in another 2 tests (`test_dagmc.py` and `test_source_sample.py`) with a simple `from pyne import ______`. It also changes the conda build python version from `3.10.14` to `3.10.12` to match the apt build to try to fix the conda build that is failing on the develop branch after merging #1540. 

## Motivation and Context
Closes #1542.

## Changes
Bug fix to allow compatibility with Python 3.12, which doesn't have the `imp` module.

### DIFF
--- a/docker/conda_build.dockerfile
+++ b/docker/conda_build.dockerfile
@@ -23,7 +23,7 @@ ENV PATH /opt/conda/bin:$PATH
 
 # install python 3.10 because that's what apt uses
 RUN conda update conda
-RUN conda install "python=3.10.12"
+RUN conda install "python=3.12"
 RUN conda config --add channels conda-forge
 RUN conda update -n base -c defaults conda
 RUN conda install -y conda-libmamba-solver
--- a/docker/conda_build.dockerfile
+++ b/docker/conda_build.dockerfile
@@ -52,6 +52,7 @@ RUN conda update -y --all && \
                 cython \
                 future \
                 progress \
+                meson \
                 && \
     conda clean -y --all
 RUN mkdir -p `python -m site --user-site`
--- a/docker/conda_build.dockerfile
+++ b/docker/conda_build.dockerfile
@@ -23,7 +23,7 @@ ENV PATH /opt/conda/bin:$PATH
 
 # install python 3.10 because that's what apt uses
 RUN conda update conda
-RUN conda install "python=3.10"
+RUN conda install "python=3.10.12"
 RUN conda config --add channels conda-forge
 RUN conda update -n base -c defaults conda
 RUN conda install -y conda-libmamba-solver
--- a/docker/conda_build.dockerfile
+++ b/docker/conda_build.dockerfile
@@ -53,6 +53,7 @@ RUN conda update -y --all && \
                 future \
                 progress \
                 meson \
+                expat \
                 && \
     conda clean -y --all
 RUN mkdir -p `python -m site --user-site`
--- a/docker/conda_build.dockerfile
+++ b/docker/conda_build.dockerfile
@@ -33,6 +33,7 @@ RUN conda uninstall -y conda-libmamba-solver
 RUN conda config --set solver classic
 RUN conda update -y --all && \
     mamba install -y \
+                expat \
                 gxx_linux-64 \
                 gcc_linux-64 \
                 cmake \
@@ -53,7 +54,6 @@ RUN conda update -y --all && \
                 future \
                 progress \
                 meson \
-                expat \
                 && \
     conda clean -y --all
 RUN mkdir -p `python -m site --user-site`
--- a/docker/conda_build.dockerfile
+++ b/docker/conda_build.dockerfile
@@ -15,23 +15,18 @@ RUN apt-get update \
     && apt-get clean -y
 
 RUN echo 'export PATH=/opt/conda/bin:$PATH' > /etc/profile.d/conda.sh && \
-    wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh && \
-    /bin/bash ~/miniconda.sh -b -p /opt/conda && \
-    rm ~/miniconda.sh
-
+    wget --quiet "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh" -O ~/miniforge.sh && \
+    /bin/bash ~/miniforge.sh -b -p /opt/conda && \
+    rm ~/miniforge.sh
+    
 ENV PATH /opt/conda/bin:$PATH
 
 # install python 3.10 because that's what apt uses
 RUN conda update conda
 RUN conda install "python=3.12"
-RUN conda config --add channels conda-forge
-RUN conda update -n base -c defaults conda
-RUN conda install -y conda-libmamba-solver
-RUN conda config --set solver libmamba
-RUN conda install -y mamba
-RUN conda uninstall -y conda-libmamba-solver
-RUN conda config --set solver classic
-RUN conda update -y --all && \
+RUN mamba update -n base conda mamba && \
+    mamba update -y python --no-pin && \
+    mamba update -y --all && \
     mamba install -y \
                 expat \
                 gxx_linux-64 \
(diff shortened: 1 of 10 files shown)


docker/conda_build.dockerfile
```diff
@@ -23,7 +23,7 @@ ENV PATH /opt/conda/bin:$PATH
 
 # install python 3.10 because that's what apt uses
 RUN conda update conda
-RUN conda install "python=3.10"
+RUN conda install "python=3.10.12"
```
gonuke: I'm anxious about this... Any time we pin a version, we fall behind.  Can you try 3.12?

gonuke (changes requested):
One request to move this forward rather than pin it backward.

docker/conda_build.dockerfile
bquan0: Sure, but the apt build currently uses Python 3.10.12.

gonuke: it failed... there must be something in newer python that we need to catch up to

ahnaf-tahmid-chowdhury: The scikit-build-core branch is working with the latest Python and is only waiting for the MOAB release. Alternatively, can we do something similar to what we did with pydagmc?

gonuke: It would be nice to know what the cause of the problem is, though.

bquan0:
I was able to fix one of the errors by installing `meson`, but now I'm running into this `ImportError: libexpat.so.1: cannot open shared object file: No such file or directory` error which I can't find a solution to.

gonuke:
[This PR in Cyclus](https://github.com/cyclus/cyclus/pull/1763) seemed to solve the issue of the missing libexpat.  Not sure exactly what accomplished the fix other than a change in conda base forcing a reinstall - @bennibbelink ?

bquan0:
After applying Ben's fixes, the conda image now builds properly. Unfortunately, I'm running into this `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x95 in position 2: invalid start byte` in line 188 of `nucname.pyx` which I can't resolve. I looked into this error before and the previous fix was going from Python 3.11 to 3.10, but obviously that's not an option here. I did look into the [changelog of Python 3.11](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep624) and found some info about how the Py_UNICODE encoders were deprecated but I don't think that's the issue since we were using `encode()`.

bennibbelink:
The resolution of that PR actually did not force a reinstall of `libexpat` because I ran into compatibility issues with `glibc`.  I ended up switching the conda builds to use Miniforge instead of Miniconda to build the environment.  The motivation being that since we rely heavily on conda-forge we should run into less compatibility issues between Python and the packages we need (due to conda-forge's robust CI).

Note that line 28 of `conda_build.dockerfile` is updating python to the latest available version.  Miniforge's base environment uses `python==3.10`, so this was my workaround to use `python==3.12` (whatever the latest version is) in the base environment in CI.

bennibbelink:
@bquan0 I was able to get the `conda_build.dockerfile` to build locally but struggling with CI with what appear to be similar issues as you're experiencing

EDIT: I lied :smile: its not building locally... looking for a fix

bennibbelink:
Is there any desire to keep supporting Python 2?

gonuke:
No, we abandoned Python 2 support some time ago!

bennibbelink:
I'm a little confused, hopefully someone can help my understanding here.  In `test_material.py` line 774, `mat["U234"]` raises a `KeyError`.  It makes sense to me that an invalid key is passed, however it is being thrown from `material.pyx` line 1484 which is within a block to process an `int`.  Why is this the case? Is `"U234"` not a string? (or is there some other translation occurring behind the scenes somewhere?)

gonuke:
> I'm a little confused, hopefully someone can help my understanding here. In `test_material.py` line 774, `mat["U234"]` raises a `KeyError`. It makes sense to me that an invalid key is passed, however it is being thrown from `material.pyx` line 1484 which is within a block to process an `int`. Why is this the case? Is `"U234"` not a string? (or is there some other translation occurring behind the scenes somewhere?)

This is a recursive call to `__getitem__`.  The original entry point passes through line 1487, converts the string to an int using `nucname.id()` and the calls itself with the int.  We always store material internally with a canonical integer representation of the nuclide.

gonuke:
I think this needs a rebase now @bquan0 following the merge of #1544

gonuke:
Maybe a red herring, but... this is the only test that is supposed to throw a `RuntimeError`. The other tests throw other kinds of errors and don't see to have this problem.  I wonder how the exception is formulated in C++ and passed back to python when it occurs?  Found where it is generated in `nucname.cpp` Line 585???

bquan0:
One idea I had was using `except +RuntimeError` for those functions that throw the custom exceptions in the .cpp files.[ Here's the Cython doc section that explains it.](https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html#exceptions) However, I got a `'__pyx_builtin_RuntimeError' was not declared in this scope; did you mean '__pyx_builtin_ValueError'` when I tried it.

Based on the [Cython doc](https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html#exceptions) it seems like the problem is that Cython tries to convert the C++ exception message to a Python string but uses the decode() function which throws the `UnicodeDecodeError`. The Cython doc mentions that we can set up our own exception handlers for the C++ exceptions but I don't know if that's worth developing.

gonuke:
Thanks for this update @bquan0 … a few things to try in order to learn more, focusing on the `NotANuclide` exception thrown by `nucname`:
1. Write a small C++ code that calls the `nucname` method directly and see what the exception message is in C++ space. Are there some characters in addition to the message generated by `what()`?
2. Delete the `what()` method from the exception and see whether it still causes a problem.  Is the problem caused by the way that message is generated, or something more fundamental?

bquan0:
After some research, I found that the problem is that in the `what()` function, we return a `const char*` pointer to a string, but the string gets deallocated by the end of the function so the pointer is no longer valid. Here's a few solutions:

- have `what()` choose between a few default `const char*` messages
- remove the `what()` function entirely
- allocate the string on the heap as a char array so that it doesn't deallocate (this works but causes memory leaks)
- allocate the string on the stack (doesn't work if the message is too long)

UPDATE: I found another solution.
I added a string `errormsg` as another member of the `NotANuclide` class and had all the non-default constructors set `errormsg` with the same code that was originally in `what()`. Then, in `what()`, I just do `return errormsg.c_str()`. This way, we can have the error message and don't need to deal with memory leaks.

bquan0:
After implementing the last solution, the tests that expect `RuntimeErrror`s are now passing. I also replaced `scipy.integrate.trapz` with `scipy.integrate.trapezoid` to fix another test fail. This change was caused by the changes of [v1.14.0 of scipy](https://docs.scipy.org/doc/scipy/release/1.14.0-notes.html). 

All the tests are passing now!

src/nucname.h
```diff
@@ -84,7 +84,10 @@ namespace nucname
   {
   public:
     /// default constructor
-    NotANuclide () {};
+    NotANuclide () 
+    {
+      NaNEstr = "";
+    };
```
gonuke: We can use an initializer here:
```suggestion
    NotANuclide () : NaNEstr("") {};
```

src/nucname.h
```diff
@@ -96,6 +99,7 @@ namespace nucname
     {
        nucwas = wasptr;
        nucnow = nowptr;
+       setNaNEstr();
```
gonuke, line 99: Do we need this function, or can we just assign the value of this new member variable in `what()`?

src/nucname.h
```diff
@@ -138,13 +143,19 @@ namespace nucname
...
       }
-      const char* NaNEstr_rtn = NaNEstr.c_str();
-      return NaNEstr_rtn;
+    };
+
+    /// Generates an informational message for the exception
+    /// \return The error string
+    virtual const char* what() const throw() 
+    {
+      return NaNEstr.c_str();
```
gonuke, line 149: Good catch!!! This was probably always wrong and just lucky that we haven't been impacted before now.

gonuke (changes requested):
I suggested a few changes that should also be applied to `rxnname.h`

src/nucname.h
bquan0: The `const` after `what()` prevents setting the value of member variables like `NaNEstr`, so the only way is to assign the error string in the constructors.

gonuke (approved):
Thanks for the deep dive on this @bquan0

PR #1550, opened 2024-09-15
Amalgamate Script Enhancement

ahnaf-tahmid-chowdhury:
## Summary

This PR adds a new functionality to `amalgamate.py` that automatically generates the `pyne_version.h` file based on the current Git tag. The purpose of this update is to ensure the versioning of PyNE is captured directly from the Git repository during development without requiring the entire PyNE library to be built.

### Key Changes:
1. **Version Fetching**:
   - The script now attempts to fetch the current Git tag using the command:
     ```python
     version = subprocess.check_output(['git', 'describe', '--tags']).strip().decode('utf-8')
     ```
   - If the tag cannot be retrieved (e.g., not in a Git repository), an exception is raised to alert the user:
     ```python
     raise Exception("Could not get the Git version. Make sure you are in a Git repository and there are tags.")
     ```

2. **Header File Generation**:
   - A new function `create_version_header(version, filepath)` is introduced, which:
     - Accepts the fetched `version` and the desired `filepath` for the header file.
     - Generates a `pyne_version.h` header file with the following format:
       ```c
       #ifndef PYNE_VERSION_HEADER
       #define PYNE_VERSION_HEADER

       #define PYNE_VERSION "{version}"

       #endif // PYNE_VERSION_HEADER
       ```
     - Ensures the target directory exists using `os.makedirs`.
     - Writes the generated content to the specified file.

3. **File Path**:
   - By default, the version header is created at: `src/pyne_version.h`.

### Purpose:

The primary motivation for this change is to streamline the version management during development. The `amalgamate.py` script does not require the full PyNE package to be installed, and as such, it uses Git to obtain the version and create the necessary version header.

In the future, when PyNE is built via `pip`, the `pyne_version.h` will be generated automatically during the build process. For now, developers can use this approach to ensure the version file is created correctly from the current Git state.

---

## Testing Instructions:

To test the updated `amalgamate.py` script, follow these steps:

1. Clone the PyNE repository:
   ```bash
   git clone https://github.com/pyne/pyne.git
   ```

2. (Optional) Checkout the desired tag (default is `develop`):
   ```bash
   git checkout <tag>
   ```

3. Run the `amalgamate.py` script:
   ```bash
   python amalgamate.py
   ```

4. Verify that the `src/pyne_version.h` file has been generated and contains the correct version number from the Git tag.

---

## Notes:

- This update assumes the user is working within a valid Git repository with tags available.
- Future enhancements may focus on automating the version file generation during package build processes with `pip`.


gonuke:
This change results in the possibility that the python interface (`pyne/pyne_version.py`) has a different version number than the C++ interface.

We have historically made it a feature that the version number is defined in once place and in a file in the repo, not just by git tags.  If we are going to move away from this, we need a strong justification and clear strategy to ensure that python and C++ are guaranteed to have the same version string.

ahnaf-tahmid-chowdhury:
Actually, pyne_version.py will be removed. I am using importlib.metadata in pyne.__init__.py to retrieve the version info, similar to the approach used by OpenMC. The `pyproject.toml` will pull the version from the git tag, ensuring that both Python and C++ will have the same version string.

gonuke:
How will the version string be set if someone downloads a zip file of the source code with no git repository?

ahnaf-tahmid-chowdhury:
actually `pyproject.toml` uses dynamic version where it uses `setuptool_scm` to get the version which requires git tag. The issue I found, it is very difficult for `pyproject.toml` to get the version information from the `pyne_version.py`. I didn't get success to extract the version info. So, to mitigate this issue, what I am thinking is to make a `version.txt file` that only contains the version number which can be easily read by `amalgamate.py`, we can create a workflow which will update the file whenever we release a new tag. Please let me know what do you think of this approach.

gonuke:
I think that makes sense.   I'll also see how OpenMC folks deal with versioning without git....

gonuke:
Maybe any changes to how we handle versioning should be part of the same PR

ahnaf-tahmid-chowdhury:
> I think that makes sense. I'll also see how OpenMC folks deal with versioning without git....

OpenMC uses `pyproject.toml` and `CMakeLists.txt` to set version manually. And I have found, the version string for both are not the same.

```console
# pyproject.toml
version = "0.15.1-dev"

# CMakeLists.txt
set(OPENMC_VERSION_MAJOR 0)
set(OPENMC_VERSION_MINOR 15)
set(OPENMC_VERSION_RELEASE 1)
set(OPENMC_VERSION ${OPENMC_VERSION_MAJOR}.${OPENMC_VERSION_MINOR}.${OPENMC_VERSION_RELEASE})
configure_file(include/openmc/version.h.in "${CMAKE_BINARY_DIR}/include/openmc/version.h" @ONLY)
```

ahnaf-tahmid-chowdhury:
I've moved the auxiliary scripts to tools folder and adjusted accordingly to make the project structure cleaner.

gonuke:
While it is important to make sure version strings are consistent, I do like the approach in OpenMC.

gonuke:
However, OpenMC has a very recent change.  If you look at the ZIP file for v0.15.0, they used a different approach.  The version string was set in `openmc/__init__.py` and then extracted from there in the `pyproject.toml` file.

It looks like they are moving to a different approach because the version string in `openmc/__init__.py` is now extracted from the C++ library, and that is set from CMake when they configure `version.h`

ahnaf-tahmid-chowdhury:
Previously, OpenMC relied on the setup.py file to extract version information from the __init__.py file. However, in recent updates, OpenMC has completely removed setup.py and now relies solely on pyproject.toml. I believe this transition is part of moving to scikit-build-core as well. I think @shimwell might have better insights on this.

tools/amalgamate.py
gonuke, line 1: Because of the way this file was moved, we've lost the ability to see the differences from the old file.  I wonder if we can make the moving of all these files to a `tools` directory a different PR from this one so that we can be more focused.

gonuke (changes requested):
Looking for a little help to see the difference in the amalgamate file.

CHANGELOG.rst
```diff
@@ -15,6 +15,8 @@ Next Version
    * Update MOAB, OpenMC, hdf5 versions in apt build (#1529)
    * Add dockerfile and workflow to build PyNE with Conda (#1540)
    * Remove deprecated references to ``basestring`` (#1544)
+   * Add CI for MacOS (#1557)
```
gonuke: I think this line is a carry over from a previous branch point - this line is above in the CHANGELOG

CHANGELOG.rst
```diff
@@ -15,6 +15,8 @@ Next Version
...
+   * Automatically Generate pyne_version.h using Git Versions (#1550)
```
gonuke:
```suggestion
   * Automatically generate version.h during amalgamation using Git Versions (#1550)
```

amalgamate.py
```diff
@@ -60,112 +68,152 @@
     "src/_decay.cpp",
 ]
 
+DEFAULT_FILES = [os.path.join("..", f) for f in DEFAULT_FILES]
```
gonuke: This requires the script to be run from a specific directory - which has always been the case. Can we make this more flexible? Or at least change this to match the restored location of this script?

amalgamate.py
```diff
@@ -60,112 +68,152 @@
...
+        if not version:
+            raise Exception("Empty version string from git.")
+        return version
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(
+            f"Git describe failed. Output: {e.output.decode('utf-8').strip()}\n"
+            "Hint: Ensure your repo has tags or fallback to .git_archival.txt."
+        )
+    except FileNotFoundError:
+        archival = os.path.join(os.path.dirname(__file__), "..", ".git_archival.txt")
```
gonuke: check path of where we call this script from now

amalgamate.py
```diff
@@ -60,112 +68,152 @@
...
+        )
+        if not version:
+            raise Exception("Empty version string from git.")
+        return version
+    except subprocess.CalledProcessError as e:
+        raise RuntimeError(
+            f"Git describe failed. Output: {e.output.decode('utf-8').strip()}\n"
+            "Hint: Ensure your repo has tags or fallback to .git_archival.txt."
+        )
+    except FileNotFoundError:
```
gonuke: It's not obvious how we get here?  Does the subprocess issue a `FileNotFoundError` exception?

gonuke (changes requested):
A few questions here - I wonder if there is a useful way to test this?

amalgamate.py
ahnaf-tahmid-chowdhury: When git is not found. However, noticed a bug when not a git repo. Let me add that feature as well.

amalgamate.py
ahnaf-tahmid-chowdhury: Updated

amalgamate.py
ahnaf-tahmid-chowdhury: Using abspath.

amalgamate.py
```diff
@@ -1,31 +1,34 @@
...
-import io
+import subprocess
 from argparse import ArgumentParser
 
-sys.path.append('pyne')
-import pyne_version as pv
-
-pv.write_cpp_header('src')
+# Configuration
+SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
```
gonuke: This is a good improvement, but if/when we change the location of `amalgamate.py`, we'll either have to update each location that this is used with some relative path changes, or this will be updated once with some relative path changes.

The latter seems preferred, in which case, this variable may be better called `BASE_DIR` since we want it to always point to the root of the source file system

gonuke (changes requested):
Just one small suggestion about variable naming to hopefully improve long term readability/maintenance

ahnaf-tahmid-chowdhury:
For testing, I have moved the `amalgamate.py` to the tools dir again. 😁

gonuke:
I appreciate all the work you are doing here, but the number & nature of changes is making this hard to review:
* non-semantic changes, including black formatting & directory restructuring, are best done in a standalone PR, ideally after enough testing is in place
* separation of concerns: this PR was originally focused on updating version handling, but the scope is creeping with some optional changes

It is always easier to review and merge a PR if it is focused and clear.

Maybe this could become 4 PRs:
1. Fix versioning
2. Add basic testing
3. Rearrange directories & black formatting
4. Improvements to amalgamation as necessary

ahnaf-tahmid-chowdhury:
I understand. Initiating multiple PRs!
Start with #1571.

ahnaf-tahmid-chowdhury:
I guess everything is okay?

tools/amalgamate.py
```diff
@@ -215,7 +215,17 @@ def main():
     )
 
     args = parser.parse_args()
-    output_dir = Path(args.output_dir).resolve()
+    # Path objects are handled for both specified and default arguments
+    if args.output_dir == ".":
```
gonuke: This logic does not seem robust. How does the setting of `output_dir` determine the complexity of the other paths?

gonuke:
I no longer see the changes such as `create_version_header`

PR #1551, opened 2024-09-15
Resolve DAGMC Target Conflict and Improve Test Skips

ahnaf-tahmid-chowdhury:
## Summary

This PR refactors the PyNE build system to handle the conflict caused by the existing `dagmc` target while using Scikit Build Core. Additionally, I have improved the test suite by updating the pytest skips with clear reasons. Key changes include:

1. **Resolving DAGMC Target Conflict:**
   - In the legacy build system, we were managing the `dagmc` target using the line:
     ```cmake
     set_target_properties(_dagmc PROPERTIES OUTPUT_NAME dagmc)
     ```
     However, in the new approach, this method was not feasible. To resolve the conflict:
     - Renamed `dagmc.pyx` to `_dagmc.pyx`, which automatically creates a target `_dagmc` when building the Cython module.
     - Introduced a new `dagmc.py` that loads the `_dagmc` module.
  
2. **Test Suite Updates:**
   - Updated the test files to reflect the changes in the module name.
   - Improved the pytest skip statements by adding clear reasons for each skip, enhancing the understandability and maintainability of the test suite.

## Notes

This refactor ensures compatibility with the Scikit Build Core while preserving the integrity of the module structure and resolving conflicts with the `dagmc` target. The added reasons for pytest skips improve clarity during test runs.

### DIFF
--- a/pyne/_dagmc.pyx
+++ b/pyne/_dagmc.pyx
@@ -9,7 +9,7 @@ from pyne.utils import QA_warn
 cimport numpy as np
 import numpy as np
 
-from pyne cimport cpp_dagmc_bridge
+cimport cpp_dagmc_bridge
 from pyne.mesh import Mesh
 from numpy.linalg import norm
 from pyne.material import Material
--- a/pyne/_dagmc.pyx
+++ b/pyne/_dagmc.pyx
@@ -19,9 +19,6 @@ np.import_array()
 
 QA_warn(__name__)
 
-if sys.version_info[0] >= 3:
-    unichr = chr
-
 # Mesh specific imports
 from pyne.mesh import HAVE_PYMOAB
 from pymoab import core as mb_core, types
@@ -720,7 +717,7 @@ def _tag_to_string(tag):
         # if the byte char code is non 0
         if (part != 0):
             # convert to ascii and join to string
-            a.append(str(unichr(part)))
+            a.append(str(chr(part)))
             string = ''.join(a)
     return string
 
--- a/pyne/_dagmc.pyx
+++ b/pyne/_dagmc.pyx
@@ -9,7 +9,7 @@ from pyne.utils import QA_warn
 cimport numpy as np
 import numpy as np
 
-cimport cpp_dagmc_bridge
+from pyne cimport cpp_dagmc_bridge
 from pyne.mesh import Mesh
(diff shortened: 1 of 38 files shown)


tests/test_openmc_utils.py
```diff
@@ -262,11 +262,11 @@ def test_get_result_error_from_openmc_sp():
 
 def test_create_meshtally():
     if not HAVE_PYMOAB or sys.version_info[0] == 2:
```
gonuke: I don't think we support python 2 anymore, so maybe we can get rid of that part of the test while we're here...

tests/test_dagmc.py
```diff
@@ -5,20 +5,16 @@
...
-
 from pyne.mesh import HAVE_PYMOAB
-
 from pyne.utils import QAWarning
 
 warnings.simplefilter("ignore", QAWarning)
 
-try:
-    from pyne.mesh import Mesh
+from pyne.mesh import Mesh
```
gonuke: Could this be combine with the import above?

pyne/_dagmc.pyx
```diff
@@ -9,7 +9,7 @@ from pyne.utils import QA_warn
 cimport numpy as np
 import numpy as np
 
-from pyne cimport cpp_dagmc_bridge
+cimport cpp_dagmc_bridge
```
gonuke, line 12: I'm not sure I understand why this change and/or how it works...

gonuke (changes requested):
A few small questions - thanks for streamlining this for review @ahnaf-tahmid-chowdhury

pyne/_dagmc.pyx
ahnaf-tahmid-chowdhury: scikit-build-core uses cmake to build cython modules, and I found that the build fails if we don't switch to this. It is unable to find cpp_dagmc_bridge.

gonuke:
Note - there were many copies of the Python 2 test in that file....

ahnaf-tahmid-chowdhury:
Sorry, just have removed all the python2 support with the recent commit 🙂

gonuke (approved):
Thanks for the py3 updates @ahnaf-tahmid-chowdhury

PR #1549, opened 2024-09-14
Update PyNE Build System with Scikit-Build Core and Additional Improvements

ahnaf-tahmid-chowdhury:
## Summary
This PR updates the PyNE build system by removing deprecated components, modernizing the build process with `scikit-build-core`, and ensuring compatibility with the latest tools and libraries. Key changes include replacing `setup.py` with `pyproject.toml`, adding rpath support, HDF5 compatibility updates, and multiple bug fixes. Additionally, Dockerfile updates, testing improvements, and build automation have been incorporated.

### Build System
- **Removed deprecated CMake macros** and cleaned up legacy code.
- **Removed `setup.py` and `setup_sub.py`**: Transitioned to `pyproject.toml` for build configuration.
- Added **scikit-build-core** for modern build management.
- Added **build dependencies** that will automatically install when setting up PyNE.
- Added **rpath support** to ensure correct linking during installation.

### Source Code Updates
- Updated `material.cpp` and the **material library** for compatibility with the latest HDF5 versions.
- Fixed bugs in **transport_specials_methods** to ensure stability.

### Documentation
- Updated the **installation instructions** for building PyNE from source using the new build system.
  
### Testing and CI/CD
- Updated **Dockerfiles** (Note: tests are still failing).
- Added **skip comments** in test scripts for non-essential checks.
- **Updated test scripts** (`test_material_xrays`, `test_material_photons`) for macOS compatibility.
- **Amalgamate.py** now extracts the PyNE version from the Git commit.
- Added a temporary `quick_wheel_test.yml` workflow to check if all changes pass CI tests (this will be removed later).
- Added `build_wheels.yml` to automatically build wheels upon merge.
- Added `publish_wheels.yml` to publish PyNE wheels to PyPI.

### Removed/Deprecated
- Removed **ubuntu.sh** and **virtualbox** setup files.
- Removed outdated `install_script.yml` as it's no longer needed with the new CI system.

### New Python Scripts
- Added **dagmc.py** to handle `_dagmc` functionalities.

## Notes:
- Please note that Docker-based tests are currently failing, and I am investigating the issue. Temporary workflows like `quick_wheel_test.yml` will be removed after final testing.

PR #1552, opened 2024-09-15
Fix material testing inaccuracies by using assert_allclose for value comparison

ahnaf-tahmid-chowdhury:
## Description: 

In the material tests on macOS, there were issues due to minor floating-point inaccuracies, as values were close but not exactly the same. To resolve this, I replaced strict equality checks with assert_allclose to allow for a small tolerance in value comparison, ensuring the tests pass with similar values. This approach improves the reliability of the tests across different environments where floating-point precision may vary slightly.


gonuke:
Have you experimented to see how low we can go with `rtol` and still match?

tests/test_material.py
```diff
@@ -2042,22 +2043,26 @@ def test_material_gammas():
...
+            (93.35, 7.135646069465136e-18),
+            (89.953, 4.4112564001433475e-18),
+            (105.0, 3.394789326064896e-18),
+            (13.0, 2.9652250902501514e-17),
+            (93.35, 5.2767097753586244e-21),
+            (89.953, 3.2620619831267015e-21),
+            (105.0, 2.510398896994686e-21),
+            (13.0, 3.4433858446293e-17),
+        ],
+        rtol=1e-10, atol=0
```
gonuke: Can we set `rtol` once and use that variable everywhere?

gonuke (changes requested):
Thanks @ahnaf-tahmid-chowdhury - just one question about setting a single value of `rtol`

tests/test_material.py
ahnaf-tahmid-chowdhury: Done

tests/test_material.py
```diff
@@ -46,6 +47,11 @@
 
 leu = {922380000: 0.96, 922350000: 0.04}
 
+# Relative Tolerance
+rtol = 1e-10
```
gonuke: Do we know if this works at 1e-12?

gonuke (changes requested):

tests/test_material.py
ahnaf-tahmid-chowdhury: We can test this while working on PR #1549. For now I am setting the value to 1e-20.

gonuke (approved):

PR #1554, opened 2024-09-16
Remove endftod.f90 and rely on C++ version

### LINKED ISSUE #1553: Fortran Dependency in PyNE for ENDF Data Handling
### Background:
Currently, PyNE uses Fortran as a core dependency due to its reliance on `endftod.90` for building the core library related to ENDF (Evaluated Nuclear Data File) data processing. The Fortran component is linked with the following source files:

- **C++ Source Files**: 
  - `utils.h`
  - `utils.cpp`

On the Python side, the following files are used to build the Cython module:

- **Cython Files**:
  - `_utils.pyx`
  - `cpp_utils.pyx`

Additionally, the module is loaded in Python via `utils.py`. There is also a file `endf.pyx` that extends the functionality of the ENDF data processing.

### Problem:
The use of Fortran for this core functionality introduces a dependency that might complicate the build process and limit flexibility. A refactor of the codebase could explore alternatives to Fortran to improve maintainability and portability of PyNE.

### Proposed Solutions:

**Option 1: Refactor Fortran Dependency**
   - Investigate the feasibility of replacing the Fortran component with a C++ or Python-based solution while maintaining or improving the performance of the ENDF data processing routines.
   - This would involve refactoring `endftod.90` and integrating its functionality directly into C++ or Python, as well as enhancing the Cython modules (`_utils.pyx`, `cpp_utils.pyx`, `endf.pyx`).

**Option 2: Make Fortran Optional**
   - Keep the Fortran files optional, allowing the build process to work without Fortran if it's not available. In this case:
     - Fortran would only be used when ENDF data processing is specifically required.
     - Provide a fallback solution for `utils` since it handles other modules. This may involve splitting `utils` into separate modules:
       - **Core `utils` (without Fortran dependency)**: For general utility functions and core functionality.
       - **ENDF-specific components**: Move Fortran-related functionality from `endftod.90` to `endftod.h` and `endftod.cpp`. These C++ files would handle the ENDF-related processing conditionally, based on the availability of Fortran.
   - This approach maintains flexibility for users who may not need the Fortran-based ENDF processing while preserving the option to use it when necessary.

### Impact:
- **Option 1**: Would streamline PyNE’s build process, reduce external dependencies, and enhance overall portability.
- **Option 2**: Offers a more flexible solution, maintaining Fortran for specific ENDF functionality while allowing users to avoid the dependency if it's not required. The use of `endftod.h` and `endftod.cpp` would handle ENDF-related functionality conditionally, improving adaptability.

ahnaf-tahmid-chowdhury:
See #1553


src/utils.h
```diff
@@ -63,10 +63,9 @@ namespace pyne {
   /// Converts a string from ENDF format to a float. Only handles E-less format
   /// but is roughly 5 times faster than endftod.
   double endftod_cpp(char * s);
-  double endftod_f(char * s); ///< Converts a string from ENDF format to a float.
-  extern  double (*endftod)(char * s); ///< endftod function pointer. defaults to fortran
 
-  void use_fast_endftod();/// switches endftod to fast cpp version
+  // Set the function pointer to always use the C++ version.
+  extern double (*endftod)(char * s);
```
gonuke: If we are getting rid of the different versions, can't we just rename `endftod_cpp` to `endftod` and get rid of the function pointers?

tests/test_endf.py
```diff
@@ -921,9 +921,9 @@ def test_evaluation_neutron():
     assert u235.info["sublibrary"] == 10
     assert u235.info["format"] == 6
     assert u235.material == 9228
-    assert u235.projectile["mass"] == 1.0
+    assert_allclose(u235.projectile["mass"], 1.0)
```
gonuke: I'm concerned about these changes.  I'm definitely curious to see how the `endftod` result is different from the exact test.  I know that there can be floating point issues with exact comparison, but I hoped that conversion from string would not be subject to hose kinds of issues.

tests/test_endf.py
```diff
@@ -1167,7 +1167,7 @@ def test_evaluation_electroatomic():
     assert_array_almost_equal(brem.xs(E), xs)
 
     # Check bremsstrahlung secondary distributions
-    assert brem.products[0]["za"] == 11  # electrons
```
gonuke, line 1170: This should be an integer comparison and should be exact.  Maybe the issue is that elsewhere in PyNE we need to make sure this is converted from float to integer, since it's probably formatted as a float in ENDF even though it's an integer.

tests/test_endf.py
```diff
@@ -1210,7 +1210,7 @@ def test_evaluation_relaxation():
 
     # Levels
     data = xe.atomic_relaxation
-    assert len(data) == 17
+    assert len(data) >= 17
```
gonuke: This change concerns me a little....

gonuke (changes requested):
I think we can simplify this further if we are going to make this change.

I'm also a little worried about the need to change the test.

tests/test_endf.py
ahnaf-tahmid-chowdhury: Hmm,interesting to me as well. It failed with Python 3.13. But with 3.10 its working. Maybe it was buffer issue.

tests/test_endf.py
ahnaf-tahmid-chowdhury: I have changed it to 11.0

gonuke (approved):

Issue #1559, opened 2025-01-31
Duplicated cmake code


php1ic:
In the top level [CMakeLists.txt](https://github.com/pyne/pyne/blob/develop/CMakeLists.txt) file the block of code from line 14 to line 21 is also defined as a function in [cmake/PyneMacros.cmake](https://github.com/pyne/pyne/blob/develop/cmake/PyneMacros.cmake) which is then called on line 22 of the top level file.

I think the block in the top level file can be deleted?

Happy to create a PR to make this simple change.

gonuke:
Thanks for catching this @php1ic - we'll definitely take a look and clean this up

gonuke:
Also happy for a PR if you're willing, @php1ic

php1ic:
Fixed in #1562

PR #1562, opened 2025-02-05
Remove duplicated code in top level cmake file

php1ic:
## Description
Remove duplicated code in top level cmake file

## Motivation and Context
PR to address #1559

## Changes
Code cleanup

## Behavior
Remove duplicate check, no change to behaviour

## Other Information
None


gonuke (approved):
This passes tests, so I think it is all working.  Thanks for the help @php1ic

PR #1564, opened 2025-04-11
Fix `in` operator for MaterialLibrary

### LINKED ISSUE #1563: Querying a MaterialLibrary for a material that is not in the library results in a new Material being created
**Describe the Bug**<br/>
A `MaterialLibrary` is expected to behave like a Python dictionary, but when checking to see if a key exists, it always returns true because non-existent keys end up adding a new empty `Material` to the `MaterialLibrary`.  There is no record of this being an intentional behavior and should probably be resolved.

Changing this behavior has the potential to create a backwards compatibility issue for many users.

**To Reproduce**<br/>
```
# generate a nucvec...
mat_lib = pyne.MaterialLibrary({ 'foo' : nucvec })
if "bar" in mat_lib.keys():
    print("'bar' should not be found)
```

**Expected Behavior**<br/>
If a `MaterialLibrary` does not contain a key, it should report that it does not exist.

**Possible Remedy**<br/>
* Change the behavior of `MaterialLibrary::get_material_ptr()` to return some indication of failure.
* Change `MaterialLibrary.__getitem__` to handle this failure properly

ahnaf-tahmid-chowdhury:
## Description

Currently, checking for membership in a `MaterialLibrary` using the `in` operator (e.g., `"nonexistent_key" in mat_lib`) incorrectly returns `True` even if the key does not exist in the library. This happens because the default `__contains__` implementation relies on `__getitem__`, and `MaterialLibrary.__getitem__` returns a default-constructed Material instead of raising a `KeyError` for missing keys. This behavior is misleading and can lead to bugs.

## Solution

This PR fixes the issue by:

1.  Explicitly implementing the `__contains__` method in the `_MaterialLibrary` Cython class (`material_library.pyx`).
2.  The new `__contains__` method calls the underlying C++ `MaterialLibrary::count()` method, which efficiently and correctly checks for key existence without side effects.
3.  Adding the necessary declaration for `count()` to `cpp_material_library.pxd`.
4.  Adding a new test case (`test_matlib_contains`) to verify the correct behavior of the `in` operator and ensure that checking for non-existent keys does not modify the library.

## Backward Compatibility

This change corrects buggy behavior. Code relying on `"key" in mat_lib` will now receive the correct `False` value for missing keys. Code that implicitly relied on the incorrect `True` return value will break, but this is necessary to fix the bug. The behavior of `mat_lib[key]` (returning a default Material for missing keys) remains unchanged.

## Testing

Added `test_matlib_contains` to `test_material_library.py`.

## Additional Information

Closes #1563

### DIFF
--- a/.github/workflows/macos_build_test.yml
+++ b/.github/workflows/macos_build_test.yml
@@ -59,7 +59,7 @@ jobs:
         run: |
           mamba install -y \
               compilers \
-              cmake \
+              "cmake<4.0.0" \
               make \
               libblas \
               liblapack \
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,7 @@ Next Version
    * Fix material testing inaccuracies by using assert_allclose for value comparison (#1552)
    * Fix mcnp version check for serial mcnp6 in PtracReader (#1558)
    * Remove duplicate cmake code that checks for compiler compatibility (#1562)
+   * Fix in operator for MaterialLibrary (#1564) 
 
 v0.7.8
 ======
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Defines the CMake commands/policies
-cmake_minimum_required(VERSION 3.0.0)
+cmake_minimum_required(VERSION 3.5.0)
 
 if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12.0")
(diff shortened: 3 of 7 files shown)


gonuke (approved):
This looks good to me - thanks @ahnaf-tahmid-chowdhury

gonuke:
@fusionsandwich - this change may be relevant for you, assuming you can get this updated version tied into your code base

PR #1566, opened 2025-05-28
Update docker

gonuke:
## Description
Updates the Dockerfile to build with more recent versions of dependencies.  Specifically,
* all references to `long` are removed.  They became identical to `int` in a recent version and now create an error, specifically in Cython compilations
* add a `make_cores` Docker build_arg to support variable sized parallel builds at all stages
* updated syntax of Docker `ENV` command
* newer standard version of HDF5 (with newer well-formed version string)
* add a `dagmc_version` Docker build_arg and set a new version as default
* force OpenMC to install in `--user` mode

## Motivation and Context
Build under CI had become unreliable.

## Changes
This addresses bugs in the Docker build

### DIFF
--- a/docker/conda_build.dockerfile
+++ b/docker/conda_build.dockerfile
@@ -5,7 +5,7 @@ FROM ubuntu:${ubuntu_version} AS pyne-deps
 ENV TZ=America/Chicago \
     HOME=/root \
     PATH=/opt/conda/bin:$PATH \
-    LD_LIBRARY_PATH=/opt/conda/lib:$LD_LIBRARY_PATH
+    LD_LIBRARY_PATH=/opt/conda/lib
 
 # Use bash for RUN shell
 SHELL ["/bin/bash", "-c"]
--- a/docker/conda_build.dockerfile
+++ b/docker/conda_build.dockerfile
@@ -6,7 +6,7 @@ FROM ubuntu:${ubuntu_version} AS pyne-deps
 ENV TZ=America/Chicago
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
-ENV HOME /root
+ENV HOME=/root
 RUN apt-get update \
     && apt-get install -y --fix-missing \
         wget \
@@ -19,7 +19,7 @@ RUN echo 'export PATH=/opt/conda/bin:$PATH' > /etc/profile.d/conda.sh && \
     /bin/bash ~/miniforge.sh -b -p /opt/conda && \
     rm ~/miniforge.sh
     
-ENV PATH /opt/conda/bin:$PATH
+ENV PATH=/opt/conda/bin:$PATH
 
 # install python 3.12 because that's what apt uses
 RUN conda install "python=3.12"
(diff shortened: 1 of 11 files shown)


docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -108,31 +114,33 @@ RUN cd /root \
...
 FROM dagmc AS openmc
-ARG openmc_version="v0.14.0"
+# pinned at version 0.14 because Ubuntu 22.04 comes with Python 3.10 and 0.15.x requires Python 3.11+
+ARG openmc_version="v0.14.0"  
 # build/install OpenMC Python API
 RUN export HDF5_ROOT="$HDF5_INSTALL_PATH" ; \
     git clone --depth 1 --branch $openmc_version https://github.com/openmc-dev/openmc.git $HOME/opt/openmc \
     && cd  $HOME/opt/openmc \
-    && pip install .
+    && pip install --user .
```
ahnaf-tahmid-chowdhury, line 130: Perhaps we can use a virtual environment.

docker/ubuntu_22.04-dev.dockerfile
```diff
@@ -108,31 +114,33 @@ RUN cd /root \
...
 
 # Build/Install PyNE from release branch
 FROM ${pyne_test_base} AS pyne
+ARG make_cores=4
 
 COPY . $HOME/opt/pyne
 RUN export PYNE_HDF5_ARGS="--hdf5 $HDF5_INSTALL_PATH"; \
     cd $HOME/opt/pyne \
     && python setup.py install --user \
```
ahnaf-tahmid-chowdhury, line 139: This user prefix is not the same as the standard one set by pip. PyNE uses custom logic for it, which is causing a conflict.

ahnaf-tahmid-chowdhury (changes requested):
Thanks for the PR. I think using a virtual env may fix the issue, at least for now.

ahnaf-tahmid-chowdhury:
The current actions are failing, saying that python is not found.

gonuke:
> The current actions are failing, saying that python is not found.

I think that's for conda - I'm investigating now

ahnaf-tahmid-chowdhury:
Yes, conda is not activated. It's trying to use the system python. We may need to activate the base environment or use `conda run python setup.py`.

gonuke:
> Yes, conda is not activated. It's trying to use the system python. We may need to activate the base environment or use `conda run python setup.py`.

That's not the problem.  You don't need to activate a conda environment if you have never established a new environment.  For some reason, it seems that installing openmc is causing python to be uninstalled from conda!!

ahnaf-tahmid-chowdhury:
We're encountering a `Fatal Python error: Illegal instruction` in the `test_dagmc.py`. This usually points to a hardware-specific instruction being used that CPU doesn't support or an incompatible binary being run.

gonuke:
> We're encountering a `Fatal Python error: Illegal instruction` in the `test_dagmc.py`. This usually points to a hardware-specific instruction being used that CPU doesn't support or an incompatible binary being run.

I'm pretty sure this is because OpenMC is being installed with DAGMC compiled with DoubleDown and Embree, which includes processor-specific instructions that may not be available on all platforms.  One solution is to build/tests without dagmc
```
"conda-forge::openmc=*=*nodagmc*"
```

I tested this and it works for me locally

docker/conda_build.dockerfile
```diff
@@ -1,92 +1,74 @@
...
 FROM ubuntu:${ubuntu_version} AS pyne-deps
 
-# Ubuntu Setup
-ENV TZ=America/Chicago
-RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+# Set environment variables
+ENV TZ=America/Chicago \
+    HOME=/root \
+    PATH=/opt/conda/bin:$PATH \
+    LD_LIBRARY_PATH=/opt/conda/lib:$LD_LIBRARY_PATH
```
gonuke: I think docker now complains (warning) that `LD_LIBRARY_PATH` is not yet defined, so it's probably better to do this:
```suggestion
    LD_LIBRARY_PATH=/opt/conda/lib
```

gonuke (commented):
Thanks @ahnaf-tahmid-chowdhury - this looks like it might be ready now

ahnaf-tahmid-chowdhury (approved):
Let me know if there are further changes needed.

PR #1568, opened 2025-06-20
Improve support for newer ubuntu and python

gonuke:
This PR into an active branch add a new Dockerfile that aims to be a recipe for installing in updated Ubunutu with newer Python versions.

It doesn't currently work, primarily due to continued reliance on long-deprecated `setuptools` and `easy_install`.

It does install, but installs in a strange location - `$HOME/.local/local/` rather than `$HOME/.local` - and the python install seems incompatible with current packaging and discovery (see `setuptools` reference above)

It also has spatial solvers turned off to avoid FORTRAN build challenges.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -396,7 +396,7 @@ def parse_args():
     )
     other.add_argument(
         "--spatial_solvers",
-        default=True,
+        default=False,
         action="store_true",
         help="Build spatial solvers (Default)",
     )


ahnaf-tahmid-chowdhury:
We could consider creating a virtual environment and configuring the prefix to point to its directory.

PR #1570, opened 2025-06-21
Add Git versioning support

ahnaf-tahmid-chowdhury:
Introduce a mechanism to retrieve and display version information from Git, including commit hash and count. Update relevant files to reflect the new versioning approach.


setup_sub.py
```diff
@@ -75,7 +73,7 @@ def main():
     }
     setup_kwargs = {
         "name": "pyne",
-        "version": PYNE_VERSION,
+        "version": "0.7.8", # Will be updated by PR 1549
```
gonuke: For the time being, do we want some kind of 'dev' modifier here?

cmake/GetVersionFromGit.cmake
```diff
@@ -0,0 +1,120 @@
...
+
+# Ensure version string format
+if(VERSION_STRING MATCHES "${VERSION_REGEX_WITH_PREFIX}")
+    set(VERSION_NO_SUFFIX "${CMAKE_MATCH_1}")
+else()
+    message(FATAL_ERROR "Invalid version format: Missing base version in ${VERSION_STRING}")
+endif()
+
+# Check for development state
+if(VERSION_STRING MATCHES "-([0-9]+)-g([0-9a-f]+)")
```
gonuke: Isn't this where you would use `VERSION_SUFFIX`? I notice it isn't used anywhere...

gonuke (changes requested):
Thanks @ahnaf-tahmid-chowdhury  Two little question

cmake/GetVersionFromGit.cmake
gonuke: Hmmm... that was more of a question and something to think about because:
• the regex you replaced is different from the one in SUFFIX
• the you probably need to replace the regex on line 96 to match (one of the advantages of using a variable)

So which regex is correct?

ahnaf-tahmid-chowdhury: Hmm. It seems the original code was modified, and unnecessary values were removed from OpenMC at that time. This was generally used for checking the suffix in the version tag. Now I can remember.

cmake/GetVersionFromGit.cmake
```diff
@@ -0,0 +1,120 @@
...
+set(VERSION_PREFIX "v" CACHE STRING "Prefix used in version tags")
+set(VERSION_SUFFIX "[~+-]([a-zA-Z0-9]+)" CACHE STRING "Suffix used in version tags")
+set(VERSION_REGEX "?([0-9]+\\.[0-9]+\\.[0-9]+)" CACHE STRING "Regex for extracting version")
+set(ARCHIVAL_FILE "${CMAKE_SOURCE_DIR}/.git_archival.txt" CACHE STRING "Path to .git_archival.txt")
+set(DESCRIBE_NAME_KEY "describe-name: " CACHE STRING "Key for describe name in .git_archival.txt")
+set(COMMIT_HASH_KEY "commit: " CACHE STRING "Key for commit hash in .git_archival.txt")
+
+
+# Combine prefix and regex
+set(VERSION_REGEX_WITH_PREFIX "^${VERSION_PREFIX}${VERSION_REGEX}")
```
ahnaf-tahmid-chowdhury, line 25: Here we have used VERSION_REGEX.

gonuke: Yes - I knew that - my comment below is something different - perhaps I'm not being clear.

cmake/GetVersionFromGit.cmake
```diff
@@ -0,0 +1,120 @@
...
+    set(VERSION_NO_SUFFIX "${CMAKE_MATCH_1}")
+else()
+    message(FATAL_ERROR "Invalid version format: Missing base version in ${VERSION_STRING}")
+endif()
+
+# Check for development state
+if(VERSION_STRING MATCHES "-([0-9]+)-g([0-9a-f]+)")
+    set(DEV_STATE "true")
+    set(COMMIT_COUNT "${CMAKE_MATCH_1}")
+    string(REGEX REPLACE "-([0-9]+)-g([0-9a-f]+)" "" VERSION_WITHOUT_META "${VERSION_STRING}")
```
gonuke: I think you misunderstood my point originally, let's try this....

The snippet of code below used a regular expression that was NOT defined as a variable above.  It appears to be similar to, but not quite the same as the `VERSION_SUFFIX` defined above. It certainly appears to semantically be used for the purpose of finding/recording/replacing a suffix.

If the original one here is the correct regular expression, then perhaps `VERSION_SUFFIX` should be redefined to match this one?  In any case, I recommend that the "magic" regular expression here be replaced by some variable, because it is used in two places and you want to make sure it has the same value in both places.

```suggestion
if(VERSION_STRING MATCHES ${VERSION_SUFFIX})
    set(DEV_STATE "true")
    set(COMMIT_COUNT "${CMAKE_MATCH_1}")
    string(REGEX REPLACE ${VERSION_SUFFIX} "" VERSION_WITHOUT_META "${VERSION_STRING}")
```

gonuke (changes requested):
I've tried to clarify my suggestion about the use of the VERSION_SUFFIX below

cmake/GetVersionFromGit.cmake
ahnaf-tahmid-chowdhury: Sorry for the earlier miscommunication.

To clarify, `VERSION_STRING` contains the output of `git describe`, which includes the version tag, commit count, and short hash. That’s why I used the regular expression `-([0-9]+)-g([0-9a-f]+)`, where `[0-9]+` captures the commit count and `g([0-9a-f]+)` captures the hash.

`VERSION_SUFFIX`, on the other hand, is used for a different purpose—it represents semantic suffixes like `-dev` in version tags. For example, in a tag like `v0.7.8-dev-56-gabcd1234`, `-dev` would be matched by `VERSION_SUFFIX`, while `-56-gabcd1234` comes from `git describe` and is handled by the other regex.

While the current pattern hasn’t needed modification over time, it would still be a good improvement to define it explicitly as a variable to avoid "magic" regex strings.

Thanks for pointing it out!

gonuke (approved):
This looks good to me @ahnaf-tahmid-chowdhury - thanks for this contribution

PR #1571, opened 2025-06-25
Add Git support for version handling to amalgamate.py

ahnaf-tahmid-chowdhury:
Implement functionality to retrieve the version from a Git repository and create a version header file. This enhances the amalgamation script by ensuring it can dynamically include version information based on the current Git state.

### DIFF
--- a/amalgamate.py
+++ b/amalgamate.py
@@ -24,10 +24,9 @@ import io
 # Configuration
 BASE_DIR = Path(__file__).resolve().parent
 
-CODE_EXTS = {".c", ".cpp", ".cxx", ".h", ".hpp", ".hxx"}
-CODE_EXTS |= {ext.upper() for ext in CODE_EXTS}
 SOURCE_EXTS = {".c", ".cpp", ".cxx"} | {ext.upper() for ext in [".c", ".cpp", ".cxx"]}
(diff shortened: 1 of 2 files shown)


amalgamate.py
```diff
@@ -1,31 +1,37 @@
...
-from argparse import ArgumentParser
-
-sys.path.append('pyne')
-import pyne_version as pv
 
-pv.write_cpp_header('src')
+# Configuration
+BASE_DIR = Path(__file__).resolve().parent
 
 CODE_EXTS = {".c", ".cpp", ".cxx", ".h", ".hpp", ".hxx"}
```
gonuke: (I know this is not something you added/changed, and it may defy my suggestion about scope creep, but you are modifying these lines anyway....)

Could we just use the following definition after we define the others?
```
CODE_EXTS = SOURCE_EXTS + HEADER_EXTS
```

gonuke (changes requested):
This looks good.  I made a small suggestion here since it jumped out at me because of some lines you clarified.

I like the introduction of `Pathlib` and will maybe suggest a follow-up PR where we remove all the `os.path` and replace with `Pathlib` counterparts

gonuke (approved):

PR #1572, opened 2025-06-25
Update Ubuntu Docker configuration and fix compatibility issues

ahnaf-tahmid-chowdhury:
Adjust the Docker configuration for Ubuntu and resolve compatibility issues with newer Python and Cython versions. Enhance memory safety in the codebase.


docker/ubuntu.dockerfile
```diff
@@ -0,0 +1,150 @@
...
+        tables \
+        matplotlib \
+        jinja2 \
+        setuptools \
+        future \
+        progress
+
+ENV PATH=${VENV_PATH}/bin:$PATH \
+    LD_LIBRARY_PATH=${VENV_PATH}/lib:$LD_LIBRARY_PATH \
+    PYTHONPATH=${VENV_PATH}/lib/python3.12/site-packages:$PYTHONPATH
```
gonuke: This is fragile to the python version changing.  There is a way to generate this automatically, but not sure how it translates to `venv`:
```
`python3 -m site --user-site`
```

docker/ubuntu.dockerfile
```diff
@@ -0,0 +1,150 @@
...
+ARG INSTALL_PATH=/opt/openmc
+
+RUN git clone --depth 1 --branch ${OPENMC_VERSION} https://github.com/openmc-dev/openmc.git && \
+    cd openmc && \
+    mkdir build && cd build && \
+    cmake .. \
+        -DCMAKE_INSTALL_PREFIX=${INSTALL_PATH} && \
+    make -j ${MAKE_CORES} && \
+    make install && \
+    cd .. && \
```
gonuke: I don't think we need to do a full install of OpenMC, just the python API....

Doesn't mean we can't do a full install, but for testing purposes we don't need it.

docker/ubuntu.dockerfile
```diff
@@ -0,0 +1,150 @@
...
+        pytest \
+        tables \
+        matplotlib \
+        jinja2 \
+        setuptools \
+        future \
+        progress
+
+ENV PATH=${VENV_PATH}/bin:$PATH \
+    LD_LIBRARY_PATH=${VENV_PATH}/lib:$LD_LIBRARY_PATH \
```
gonuke: Docker complains if referencing empty ENV like `LD_LIBRARY_PATH` may be here?

docker/ubuntu.dockerfile
```diff
@@ -0,0 +1,150 @@
...
+        libeigen3-dev \
+        libblas-dev \
+        liblapack-dev \
+        libhdf5-dev && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Create and activate virtual environment
+RUN python3 -m venv ${VENV_PATH} && \
+    ${VENV_PATH}/bin/pip install --upgrade pip && \
+    ${VENV_PATH}/bin/pip install \
```
gonuke: I never know which of these to include using `apt` and which using `pip`.  Is there a clear "rule" you have used to decide in this case?  I suppose some are not available via `apt`?

docker/ubuntu.dockerfile
gonuke: Also if PYTHONPATH is originally empty, docker may complain

docker/ubuntu.dockerfile
```diff
@@ -0,0 +1,150 @@
...
+    PYNE_MOAB_ARGS="--moab ${INSTALL_PATH}"
+
+# ------------------------------
+# Stage 3: DAGMC Setup
+# ------------------------------
+FROM moab AS dagmc
+
+ARG DAGMC_VERSION=v3.2.4
+ARG MAKE_CORES
+ARG INSTALL_PATH=/opt/dagmc
```
gonuke: I think having this build arg have different defaults at each stage will not give the effect you want, especially is someone overwrites it at build time?  Maybe this should just be 
```
ARG INSTALL_PATH=/opt
```
And then embellish it with `moab` and `dagmc` as appropriate?

docker/ubuntu.dockerfile
```diff
@@ -0,0 +1,150 @@
...
+FROM ${PYNE_TEST_BASE} AS pyne
+ARG MAKE_CORES
+
+COPY . /opt/pyne
+WORKDIR /opt/pyne
+
+RUN python setup.py install $PYNE_MOAB_ARGS $PYNE_DAGMC_ARGS \
+        --clean -j ${MAKE_CORES} && \
+    cd tests && \
+    nuc_data_make
```
gonuke: This doesn't appear to run tests....

gonuke (changes requested):
This looks like some good progress @ahnaf-tahmid-chowdhury.  I have a few questions to follow up

docker/ubuntu.dockerfile
ahnaf-tahmid-chowdhury: I hadn't noticed. But we can remove this.

docker/ubuntu.dockerfile
```diff
@@ -0,0 +1,136 @@
...
+        python3-venv \
+        wget \
+        build-essential \
+        git \
+        cmake \
+        gfortran \
+        libeigen3-dev \
+        libblas-dev \
+        liblapack-dev \
+        libhdf5-dev \
```
gonuke: I think this is only version 1.10 of HDF5 in the ubuntu package.  Most of our tools prefer a newer version of HDF5, currently 1.14.  This is why we manually install HDF5.

gonuke (changes requested):
This relies on a very old HDF5 that is packaged for ubuntu

docker/ubuntu.dockerfile
ahnaf-tahmid-chowdhury: Here, I used source to activate the python environment for the current session, and later I added this virtual environment to the beginning of the PATH so that it always uses the virtual environment's python.

ahnaf-tahmid-chowdhury:
I have installed all source builds into the virtual env path, so we don't need to set env vars for each app. Similar to how conda ships its packages.

gonuke (approved):
Thanks @ahnaf-tahmid-chowdhury - this tidies things up nicely.

PR #1573, opened 2025-06-25
Refactor amalgamation logic for clarity and efficiency

ahnaf-tahmid-chowdhury:
Enhance the amalgamation process by improving the handling of header files and local includes, ensuring better organization and readability of the generated output.


amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
 
     def append_line(self, line):
-        """Adds some text to the end of the file."""
         if not line.endswith("\n"):
             line += "\n"
         self._blocks.append(line)
 
+    # Modify append_file to handle local includes
     def append_file(self, filename, comment_out=None):
-        """Adds a whole file to the end of this one."""
```
gonuke, line 154: Not sure why you are getting rid of docstrings - if anything, they should be longer/clearer

amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
+        _, ext = os.path.splitext(filename)
```
gonuke: While you are making changes, can we use Pathlib functions to do this and gradually remove `os`

amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
+                    # Extract the header name
+                    header_name = stripped_line.split('"')[1]
+                    if header_name in self.amalgamated_headers:
+                        # This is a local header that is being amalgamated, so we skip this line.
+                        # We can add a comment to show what we did.
+                        processed_lines.append(
+                            f'// Removed local #include "{header_name}"\n'
+                        )
+                        continue
+                processed_lines.append(line)
```
gonuke: Why do we need this?  We already have the
```
#ifndef PYNE_IS_AMALGAMATED
```
guards that make this not a syntax issue.

I worry about this solution being fragile to some edge case.

amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
-    def prepend_files(self):
-        """Adds a file listing to the begining of the almagamted file."""
-        s = "// This file is composed of the following original files:\n\n"
+    def prepend_file_listing(self):
+        listing = "// Amalgamated from the following files:\n"
         for f in self._filenames:
-            s += "//   {0}\n".format(f)
-        s += "\n"
-        self._blocks.insert(0, s)
+            listing += f"//   {f}\n"
```
gonuke: Can we just use join instead:
```
listing += "\n//.  ".join(self._file names)
```

amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
-        else:
-            txt = "".join([block.decode("utf-8") for block in self._blocks])
-        d = os.path.dirname(self.path)
-        if len(d) > 0 and not os.path.isdir(d):
-            os.makedirs(d)
-        with io.open(self.path, "wb") as f:
-            f.write(txt.encode("utf-8"))
+        self.prepend_file_listing()
+        final = "".join(self._blocks)
+        output_dir = os.path.dirname(self.path)
```
gonuke: Use pathlib instead of os if we're changing this anyway?

amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
-    parser = ArgumentParser()
+    parser = ArgumentParser(description="Amalgamate PyNE C++ code.")
     parser.add_argument(
         "-s",
-        dest="source_path",
-        action="store",
+        dest="source_name",
         default="pyne.cpp",
-        help="Output *.cpp source path.",
+        help="Output C++ source file name.",
```
gonuke:
```suggestion
        help="Output C++ (.cpp) source file name.",
```

amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
     )
     parser.add_argument(
-        "-i",
-        dest="header_path",
-        action="store",
-        default="pyne.h",
-        help="Output header path.",
+        "-i", dest="header_name", default="pyne.h", help="Output header file name."
```
gonuke:
```suggestion
        "-i", dest="header_name", default="pyne.h", help="Output header (.h) file name."
```

amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
+    headers_to_amalgamate = {
+        f for f in args.files if os.path.splitext(f)[1] in HEADER_EXTS
+    }
+
+    # Header and Source Generation
+    header = AmalgamatedFile(header_path)  # Header doesn't need to filter itself
+    header.append_line("// Amalgamated PyNE header - http://pyne.io/")
+    header.append_line("#ifndef PYNE_AMALGAMATED_HEADER")
+    header.append_line("#define PYNE_AMALGAMATED_HEADER\n")
+    header.append_line("#define PYNE_IS_AMALGAMATED\n")
```
gonuke: Consistency
```suggestion
    header.append_line("#define PYNE_AMALGAMATED_HEADER")
    header.append_line("#define PYNE_IS_AMALGAMATED")
```

amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
+
+    # Pass the list of amalgamated headers to the source file constructor
+    source = AmalgamatedFile(source_path, amalgamated_headers=headers_to_amalgamate)
+    source.append_line("// Amalgamated PyNE source - http://pyne.io/")
+    rel_header_path = os.path.relpath(header_path, start=os.path.dirname(source_path))
+
+    # Ensure forward slashes for cross-platform compatibility in #include
+    rel_header_path = rel_header_path.replace("\\", "/")
```
gonuke: I think using Pathlib will take care of this

amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
+    header = AmalgamatedFile(header_path)  # Header doesn't need to filter itself
+    header.append_line("// Amalgamated PyNE header - http://pyne.io/")
+    header.append_line("#ifndef PYNE_AMALGAMATED_HEADER")
+    header.append_line("#define PYNE_AMALGAMATED_HEADER\n")
+    header.append_line("#define PYNE_IS_AMALGAMATED\n")
+
+    # Pass the list of amalgamated headers to the source file constructor
+    source = AmalgamatedFile(source_path, amalgamated_headers=headers_to_amalgamate)
+    source.append_line("// Amalgamated PyNE source - http://pyne.io/")
+    rel_header_path = os.path.relpath(header_path, start=os.path.dirname(source_path))
```
gonuke: I know we did this before, but I wonder if we should use the relative path that arises from this amalgamation process, or we should just include `pyne.h` and make the user ensure that the header is in the include path for their build system?

amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
+
+    # Ensure forward slashes for cross-platform compatibility in #include
+    rel_header_path = rel_header_path.replace("\\", "/")
+    source.append_line(f'#include "{rel_header_path}"\n')
```
gonuke: I feel like it would be good to have one more blank line here
```suggestion
    source.append_line(f'#include "{rel_header_path}"\n\n')
```

amalgamate.py
```diff
@@ -138,112 +136,158 @@ def create_version_header(version, file_name="version.h"):
...
+            header.append_file(file, comment_out=True)
+            source.append_file(file, comment_out=True)
 
-    # write both
-    hdr.write()
-    src.write()
+    header.append_line("#endif  // PYNE_AMALGAMATED_HEADER")
+    header.write()
+    source.write()
+    os.remove(os.path.join(BASE_DIR, "version.h"))
```
gonuke: If you are removing this file anyway, can't it just be a string that you keep in memory rather than written as a file?

I see the value of writing it as a file because then it gets included in the order the user specifies, etc.  However, we could also decide that we always include the license and the version at the top and don't allow the user to specify these?

gonuke (changes requested):
Thanks for giving this some attention @ahnaf-tahmid-chowdhury 

I'm on board with some of these changes, but less convinced about others.

amalgamate.py
ahnaf-tahmid-chowdhury: Yes, we can keep this in memory. For license file, I think is good to have this fetcher here, that let the script makes sense we can use other text as comments.

amalgamate.py
ahnaf-tahmid-chowdhury: good catch!

amalgamate.py
ahnaf-tahmid-chowdhury: I noticed the issue while testing. Maybe not all the files have this. Need more inspection here.

amalgamate.py
gonuke: Agree to read the license from the file, but maybe we always add the license file whether or not it's specified by the user.

amalgamate.py
gonuke: I'd still prefer to rely on the `PYNE_IS_AMALGAMATED` include guards than processing the files in this way. Can you provide an example of a failure when you don't process files?

ahnaf-tahmid-chowdhury: Okay… In my test demo source, I didn’t add `PYNE_IS_AMALGAMATED`, so the build was crashing. But in the original code, since `PYNE_IS_AMALGAMATED` was not defined, the header gets skipped by the compiler, so it works. Anyway, I’ve added a strip parser that allows the loop to run. By default, it’s turned off. Let me know if that works; otherwise, I may remove this logic completely.

gonuke: I guess I'm not sure what test you tried, but `PYNE_IS_AMALGAMATED` should always be defined in the amalgamated PyNE header, so if that header is included, then it should always be defined.

gonuke: I'd prefer to remove this completely since I'm worried about edge cases were this fails and is hard to diagnose

amalgamate.py
```diff
@@ -134,116 +133,149 @@ def create_version_header(version, file_name="version.h"):
...
+# Amalgamation Logic
+class AmalgamatedFile:
+    """Class to handle the amalgamation of files into a single output file."""
+
+    def __init__(self, output_path, amalgamated_headers=None):
+        self.path = output_path
         self._blocks = []
         self._filenames = []
+        # Store the set of header for quick lookup
+        self.amalgamated_headers = amalgamated_headers or set()
```
gonuke: We don't need this anymore

amalgamate.py
```diff
@@ -134,116 +133,149 @@ def create_version_header(version, file_name="version.h"):
...
+        license_content = f"{LICENSE_FILE} not found."
+        print(f"[!] Warning: {license_content}")
+
+    header_path = output_dir / args.header_name
+    source_path = output_dir / args.source_name
+
+    # Determine which headers will be amalgamated to filter #includes later
+    amalgamated_header_names = {f.name for f in input_files if f.suffix in HEADER_EXTS}
+    # Add the generated version header to this set
+    amalgamated_header_names.add("version.h")
```
gonuke: We don't need this anymore

amalgamate.py
```diff
@@ -134,116 +133,149 @@ def create_version_header(version, file_name="version.h"):
...
+
+    # Add version header content
+    header.append_line("//\n// Begin: version.h\n//")
+    header.append_line(version_header_content)
+    header.append_line("//\n// End: version.h\n//\n")
+
+    # Source File Generation
+    source = AmalgamatedFile(
+        source_path,
+        amalgamated_headers=amalgamated_header_names,
```
gonuke: Don't need this anymore

amalgamate.py
```diff
@@ -134,116 +133,149 @@ def create_version_header(version, file_name="version.h"):
...
+    def append_commented_block(self, content, title):
+        """Appends a block of text, formatting it as a C++ comment."""
+        header = f"//\n// Begin: {title}\n//\n"
+        footer = f"//\n// End: {title}\n//\n\n"
+        commented_content = "// " + content.replace("\n", "\n// ")
+        self._blocks.append(header + commented_content + "\n" + footer)
+
+    def append_file(self, filename):
+        """Appends the content of a file to the amalgamated output."""
+        is_source_file = filename.suffix in SOURCE_EXTS
```
gonuke: Don't need this anymore

gonuke (changes requested):
Cleaning out a little more left over from the filter

gonuke (approved):

PR #1579, opened 2025-09-05
Add plot on nuclide chart as PyNE example

gonuke:
## Description
This adds a method that uses PyNE to plot results that are tabulated by nuclide on a chart of the nuclides.

## Motivation and Context
This can be a useful/interesting way to represent data that is tabulated by nucldie.

## Changes
This adds an example that includes a self-contained method to generate this type of plot, as well as a python notebook that shows how one might use it.


ahnaf-tahmid-chowdhury (commented):
It would be great to have minimal documentation in the notebook.

PR #1581, opened 2025-09-25
Allow metastable states higher than n

### LINKED ISSUE #1580: Incompatibility of nucname.cpp with pyne.alara.num_density_to_mesh()
**Describe the Bug**<br/>
The `pyne.alara.num_density_to_mesh()` method relies on the `id()` method defined in `nucname.h` to process nuclides from an ALARA output file. However, depending on the original problem, ALARA may output nuclides in the n-metastable state, as opposed to the m state. The `id()` method currently expects nuclides in metastable states to end only in the letter 'm,' which causes `pyne.alara.num_density_to_mesh()` to return a `Not a Nuclide!` error.

**To Reproduce**<br/>
Observed with a MOAB unstructured mesh overlaid onto a section of pure W. Nuclides hf-178n, hf-179n, and ta-182n cause the issue.

**Expected Behavior**<br/>
`pyne.alara.num_density_to_mesh()` should be able to accommodate at least the n state.

**Please complete the following information regarding your system:**<br/>
 - OS: debian Linux
 - PyNE Version: 0.7.8.dev0

**Additional Context**<br/>
Looking at `nucname.cpp`, lines with `newnuc += "m";` or similar may be modified to account for cases where ssss > 1, but I am unaware of whether this will break any other methods that rely on nucname.

anu1217:
## Description
`nucname.cpp` has been modified to accommodate metastable states higher than the first. These states are now allowed to be as high as the z-level. This was done by changing part of the code where the last character was hard-coded as "m" or "M." The openmc functions remain in their original format as openmc attaches numbers to the end of the nuclide to indicate different levels of metastability.  

## Motivation and Context
Fixes #1580

## Changes
This fixes the bug described in the issue referenced above. A documentation update to `nucname.h` is also likely in order.

### DIFF
--- a/src/nucname.cpp
+++ b/src/nucname.cpp
@@ -13,9 +13,13 @@
 /***String used to check for metastable states***/
 std::string pyne::nucname::metastable_states = "mnopqrstuvxyz";
 //Capitalize all letters in the metastable states string
-std::string upper_metastable = pyne::nucname::metastable_states;
-std::transform(upper_metastable.begin(), upper_metastable.end(), upper_metastable.begin(),
-               [](char c){ return std::toupper(c); });
+std::string get_upper_metastable() {
+    std::string upper_metastable = pyne::nucname::metastable_states;
+    std::transform(upper_metastable.begin(), upper_metastable.end(), upper_metastable.begin(),
+                   [](char c){ return std::toupper(c); });
+    return upper_metastable;
+}
+std::string upper_metastable = get_upper_metastable();
 
 /*** Constructs the LL to zz Dictionary ***/
 pyne::nucname::name_zz_t pyne::nucname::get_name_zz() {
@@ -1124,8 +1128,7 @@ std::string pyne::nucname::serpent(int nuc) {
 
   // Add meta-stable flag
   if (0 < ssss)
-    char meta_char = pyne::nucname::metastable_states[ssss - 1];
-    newnuc += meta_char;
+    newnuc += pyne::nucname::metastable_states[ssss - 1];
   return newnuc;
 }
 
--- a/src/nucname.cpp
+++ b/src/nucname.cpp
@@ -12,14 +12,21 @@
 
 /***String used to check for metastable states***/
 std::string pyne::nucname::metastable_states = "mnopqrstuvxyz";
-//Capitalize all letters in the metastable states string
-std::string get_upper_metastable() {
-    std::string upper_metastable = pyne::nucname::metastable_states;
-    std::transform(upper_metastable.begin(), upper_metastable.end(), upper_metastable.begin(),
-                   [](char c){ return std::toupper(c); });
-    return upper_metastable;
+
+int pyne::nucname::isomer_id(std::string end_char) {
+    
+    int id = -1;
+    size_t pos = pyne::nucname::metastable_states.find(std::tolower(end_char));
+
+    if (pos != std::string::npos) {
+        id = static_cast<int>(pos) + 1;
+    }
+    else if (pyne::contains_substring(pyne::digits, end_char)) {
+        id = 0;
+    }
+
+    return id;
 }
-std::string upper_metastable = get_upper_metastable();
 
 /*** Constructs the LL to zz Dictionary ***/
 pyne::nucname::name_zz_t pyne::nucname::get_name_zz() {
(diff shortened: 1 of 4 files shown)


src/nucname.cpp
```diff
@@ -586,12 +586,18 @@ int pyne::nucname::id(std::string nuc) {
 
     // Figure out if we are meta-stable or not
     std::string end_char = pyne::last_char(nucstr);
-    if (end_char == "M")
-      newnuc = (10000 * anum) + 1;
-    else if (pyne::contains_substring(pyne::digits, end_char))
-      newnuc = (10000 * anum);
-    else
+    std::string metastable_states = "MNOPQRSTUVWXYZ";
```
gonuke: Is there a way to declare this as a static member of the class, or something similar?

src/nucname.cpp
```diff
@@ -669,8 +675,10 @@ std::string pyne::nucname::name(int nuc) {
     newnuc += pyne::to_str(aaa);
 
   // Add meta-stable flag
+  std::string metastable_states = "MNOPQRSTUVWXYZ";
   if (0 < ssss)
-    newnuc += "M";
+    char meta_char = metastable_states[ssss - 1];
+    newnuc += meta_char;
```
gonuke:
```suggestion
    newnuc += metastable_states[ssss - 1];
```

src/nucname.cpp
```diff
@@ -833,8 +841,10 @@ std::string pyne::nucname::zzllaaam(int nuc) {
   if (0 < aaassss)
     newnuc += pyne::to_str(aaa);
   // Add meta-stable flag
+  std::string metastable_states = "mnopqrstuvwxyz";
   if (0 < ssss)
-    newnuc += "m";
+    char meta_char = metastable_states[ssss - 1];
+    newnuc += meta_char;
```
gonuke:
```suggestion
    newnuc += metastable_states[ssss - 1];
```

src/nucname.cpp
gonuke: Or maybe better would be a short function that converted from a single character to a state integer, with -1 being some error state.

gonuke (changes requested):
Here are a few suggestions... it might be good to add some tests to make sure these all work.  Certainly some new tests for correct behavior, but possibly some new tests that show failure.

src/nucname.cpp
anu1217: The string has now been added to the `nucname` namespace instead of being explicitly defined every time

src/nucname.cpp
```diff
@@ -670,7 +687,7 @@ std::string pyne::nucname::name(int nuc) {
 
   // Add meta-stable flag
   if (0 < ssss)
-    newnuc += "M";
+    newnuc += upper_metastable[ssss - 1];
```
gonuke:
```suggestion
    newnuc += std::toupper(pyne::nucname::metastable_states[ssss - 1]);
```

src/nucname.cpp
```diff
@@ -586,12 +598,17 @@ int pyne::nucname::id(std::string nuc) {
...
+
+    if (pos != std::string::npos) {
+        newnuc = (10000 * anum) + static_cast<int>(pos) + 1;
+    }
+    else if (pyne::contains_substring(pyne::digits, end_char)) {
+        newnuc = (10000 * anum);
+    }
+  else {
       throw NotANuclide(nucstr, newnuc);
+  }
```
gonuke: Maybe this can be a new function:
```
int nucname::isomer_id(std::string end_char) {
    
    int id = -1;
    size_t pos = pyne::nucname::metastable_states.find(std::tolower(end_char));

    if (pos != std::string::npos) {
        id = static_cast<int>(pos) + 1;
    }
    else if (pyne::contains_substring(pyne::digits, end_char)) {
        id = 0;
    }

    return id;
}
```

and then:

```
isomer_id = nucname::isomer_id(end_char);
if isomer_id > -1 {
     newnuc = (10000 * anum) + isomer_id(end_char, newnuc);
}
else {
    throw NotANuclide(nucstr, newnuc);
}
```

src/nucname.cpp
```diff
@@ -6,8 +6,20 @@
...
+/***String used to check for metastable states***/
+std::string pyne::nucname::metastable_states = "mnopqrstuvxyz";
+//Capitalize all letters in the metastable states string
+std::string get_upper_metastable() {
+    std::string upper_metastable = pyne::nucname::metastable_states;
+    std::transform(upper_metastable.begin(), upper_metastable.end(), upper_metastable.begin(),
+                   [](char c){ return std::toupper(c); });
+    return upper_metastable;
+}
+std::string upper_metastable = get_upper_metastable();
```
gonuke: I don't think we need these if we transform the individual character to upper/lower as appropriate.

gonuke (changes requested):
These tests look good - thanks @anu1217.

I think we can avoid the need for an uppercase version by being careful how we handle the individual character, and also having a function to go from char to number.

src/nucname.cpp
```diff
@@ -6,8 +6,27 @@
...
+#include <algorithm>
 #endif
 
+/***String used to check for metastable states***/
+std::string pyne::nucname::metastable_states = "mnopqrstuvxyz";
+
+int pyne::nucname::isomer_id(std::string end_char) {
+    
+    int id = -1;
+    size_t pos = pyne::nucname::metastable_states.find(std::tolower(end_char));
```
gonuke:
```suggestion
    size_t pos = pyne::nucname::metastable_states.find(std::tolower(end_char[0]));
```

src/nucname.cpp
```diff
@@ -6,8 +6,27 @@
...
 #include "nucname.h"
 #include "state_map.cpp"
+#include <cctype>
+#include <algorithm>
 #endif
 
+/***String used to check for metastable states***/
+std::string pyne::nucname::metastable_states = "mnopqrstuvxyz";
+
+int pyne::nucname::isomer_id(std::string end_char) {
```
gonuke, line 16: be sure to declare this function in the `nucname.h` header

gonuke (changes requested):
These should resolve the errors in the build during testing. 🤞

gonuke (approved):
Thanks @anu1217 - this looks good to me

